PR #5, opened 2020-01-22
Typography updates

soniaturcotte:
- Remove semi-bold and extra-bold font weight
- Add H4 and small body text 
- Update type sizes and remove blue for heading 
- Add link interactive states

Details in this issue: https://github.com/citizensadvice/design-system-testing/issues/1

### DIFF
--- a/scss/base/4-elements/_links.scss
+++ b/scss/base/4-elements/_links.scss
@@ -1,11 +1,9 @@
 a,
 .hyperlink {
     color: $cads-palette__heritage-blue;
-    line-height: 1.25em;
+    line-height: 1.25em; 
 
-    &:visited {
-        color: #53388D;
-    }
+    // Visited links should be browser default
 
     &:hover {
         color: #002F55;
--- a/scss/base/4-elements/_links.scss
+++ b/scss/base/4-elements/_links.scss
@@ -1,17 +1,21 @@
 a,
 .hyperlink {
-    color: $cads-palette__black;
+    color: $cads-palette__heritage-blue;
     line-height: 1.25em;
 
     &:visited {
-        color: green;
+        color: #53388D;
     }
 
     &:hover {
-        color: blue;
+        color: #002F55;
+        background-color: $cads-palette__blue-light;
     }
 
     &:active {
-        color: white;
+        color: $cads-palette__black;
+        background-color: $cads-palette__heritage-yellow;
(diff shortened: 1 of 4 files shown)


scss/base/4-elements/_links.scss
```diff
@@ -1,17 +1,21 @@
 a,
 .hyperlink {
-    color: $cads-palette__black;
+    color: $cads-palette__heritage-blue;
     line-height: 1.25em;
 
     &:visited {
-        color: green;
+        color: #53388D;
```
panosvoudouris: No hex values, use one in the language/palette and if doesn't exist then we have to add it.

scss/base/4-elements/_links.scss
```diff
@@ -1,17 +1,21 @@
...
+        color: #002F55;
+        background-color: $cads-palette__blue-light;
     }
 
     &:active {
-        color: white;
+        color: $cads-palette__black;
+        background-color: $cads-palette__heritage-yellow;
+        text-decoration: none;
+        border-bottom: 4px solid $cads-palette__black;
```
panosvoudouris: Maybe we should add something border width so we don't use px values?

scss/base/4-elements/_typography.scss
```diff
@@ -1,51 +1,82 @@
...
-    font-size: 3em;
-    line-height: 1.25em;
-    font-weight: $cads-font-weight__extra-bold;
-    color: $cads-palette__heritage-blue;
+    font-size: 2.5em;
+    line-height: 1.2em;
+    font-weight: $cads-font-weight__bold;
+    color: $cads-palette__black;
     margin: 0;
     padding: 1em 0 1em 0;
```
panosvoudouris: Are we leaving the spacing above/below?

soniaturcotte: Spacing still needs to be worked out / designed so we will add that as a separate pull request.

scss/base/4-elements/_links.scss
soniaturcotte: We will leave this for now and see if we want to have it as a variable once we have designed other interactive components.

scss/base/4-elements/_links.scss
soniaturcotte: Resolved in most recent commit

PR #6, opened 2020-01-23
Update colour language and palette

soniaturcotte:
Remove colours that are not used and add updated colours
Based on design work by Shakira and Sonia

These changes will need to be reviewed by the design system working group and are subject to change. But at least we are now using colour language throughout.

### DIFF
--- a/scss/base/6-components/_navigation.scss
+++ b/scss/base/6-components/_navigation.scss
@@ -22,8 +22,9 @@ nav,
         line-height: 1em;
 
         &:hover {
-            background-color: $cads-palette__blue-dark;
-            border-color: $cads-palette__blue-dark;
+            background-color: cads-language__link-hover-colour;
+            border-color: $cads-palette__heritage-blue;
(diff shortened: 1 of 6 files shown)


scss/base/6-components/_navigation.scss
```diff
@@ -22,8 +22,9 @@ nav,
         line-height: 1em;
 
         &:hover {
-            background-color: $cads-palette__blue-dark;
-            border-color: $cads-palette__blue-dark;
+            background-color: cads-language__link-hover-colour;
```
panosvoudouris: That's missing a $ so the colour is not picked up.

scss/base/6-components/_navigation.scss
```diff
@@ -22,8 +22,9 @@ nav,
...
+            border-color: $cads-palette__heritage-blue;
+            color: $cads-palette__heritage-blue;
```
panosvoudouris: That makes it blue text on dark blue?

PR #18, opened 2020-02-03
Callout

panosvoudouris:
Callout box styles and documentation. 

## Related cards
* https://trello.com/c/zLql5Phe/72-example-or-quote
* https://trello.com/c/osZMtAjB/71-callouts


scss/6-components/_callout.scss
```diff
@@ -11,24 +11,25 @@
...
     }
 
     &.cads-callout- {
         &example {
-            background-color: cads-colour-to-bg($cads-language__border-colour);
+            @include cads-border(
+                a,
+                $cads-border-width-medium,
+                0,
+                $cads-palette__mid-grey
```
panosvoudouris: Check colours please, not sure where we are with languages vs palette.

scss/6-components/_callout.scss
```diff
@@ -11,24 +11,25 @@
...
-        padding: 0.75em;
-        top: -1em;
-        left: -$cads-border-width-large;
-        font-weight: $cads-font-weight__bold;
+        position: relative;
         font-size: 0.75em;
-        line-height: 0.75em;
-        color: $cads-palette__white;
+        color: $cads-font-colour__base;
+        margin-bottom: 1em;
```
panosvoudouris: Switch to spacing var?

soniaturcotte: Yes that would be great, also seeing online now I think the spacing might be too large.

shakiradamji:
Callout label should be 1em on desktop and 0.875em on mobile. The colour should be #4A4E4F (secondary text colour). This style is consistent for all 'meta data' e.g. the publish date and the location switcher.

panosvoudouris:
@shakiradamji 1em is relative, do you mean 1em relative to the content? What should it be in pixels and I'll work it out from there. Checkout the callout branch and see what it looks like. Also consider the additional label for the example/important types.

shakiradamji:
@panosvoudouris Oh! 14px for small screens and 16px for large

shakiradamji:
@panosvoudouris Yes the additional label on example and important is what I meant should be the grey colour and a smaller size (although looks like the size is correct, it should be 14px small screens and 16px large screens as per comment above). Can you please put the main callout text into paragraph tags. Also reduce the padding all around the component to 1em and also reduce the margin on the bottom of the small label span to 8px. 

Screenshot there if helpful
Screenshot 2020-02-05 at 09 44 43

panosvoudouris:
> @panosvoudouris Yes the additional label on example and important is what I meant should be the grey colour and a smaller size (although looks like the size is correct, it should be 14px small screens and 16px large screens). Can you please put the main callout text into paragraph tags. Also reduce the padding all around the component to 1em and also reduce the margin on the bottom of the small label span to 8px.
> 
> Screenshot there if helpful
> Screenshot 2020-02-05 at 09 44 43

Ok so that's been updated.

shakiradamji:
Styles look good but text still isn't showing in paragraph tags, see screenshot: 
 
Screenshot 2020-02-05 at 09 44 43

shakiradamji:
@panosvoudouris sorry Ignore my last comment if you got that it's because I hadn't pulled the changes hah! All looks good except margin bottom on the span. It looks tighter than the one in the screenshot. Could you increase by 4 more px please to 12px?

PR #20, opened 2020-02-04
Notice banner

panosvoudouris:
Notice banner implementation, see the Notice Banner component story.


styleguide/components/notice-banner.js
```diff
@@ -0,0 +1,17 @@
...
+import wrapper from './component-wrapper';
+
+const noticeBanner = () => {
+    const content = text(
+        'Banner content',
+        'If you’re a Thomas Cook customer and you’re stuck abroad or want to get your money back, get help from the Civil Aviation Authority.'
+    );
+
+    const component = `<div class="cads-notice-banner">
+    <h4>Notice</h4>
```
soniaturcotte: I think this needs to be a span and not a heading. Does it make sense to have `<span class="h4">`

panosvoudouris: Depends how it would be read out, if span is more appropriate we can do the span with h4 style.

panosvoudouris: Also, the large version is just for the large breakpoint or md/lg?

scss/6-components/_notice-banner.scss
```diff
@@ -0,0 +1,40 @@
+.cads-notice-banner {
+    @include cads-border(
+        l,
+        $cads-border-width-large,
+        0,
+        $cads-language__focus-colour
```
soniaturcotte: should be `$cads-language__brand-secondary` otherwise, perfect 👍

panosvoudouris: Yes it should!

PR #21, opened 2020-02-04
Breadcrumb styles

panosvoudouris:
Update breadcrumbs to truncate long labels (note: consumer need to use the title tag to set the label accessibly).

### DIFF
--- a/scss/6-components/_breadcrumbs.scss
+++ b/scss/6-components/_breadcrumbs.scss
@@ -5,12 +5,16 @@
 
     li {
         display: inline-block;
+        max-width: 250px;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


soniaturcotte:
Nice! 

We discussed changing the mobile version on Public Advice to 
`< Previous page` rather than displaying the whold breadcrumb. It might not be the case for other products (Adviser etc). In this case, should it be handled in the project repo instead ?

panosvoudouris:
> Nice!
> 
> We discussed changing the mobile version on Public Advice to
> `< Previous page` rather than displaying the whold breadcrumb. It might not be the case for other products (Adviser etc). In this case, should it be handled in the project repo instead ?

We can probably handle that just in the CSS. This way there is no responsibility for the consumer to consider what happens on small/large.

PR #38, opened 2020-03-03
NP-91: Amend callout

clairgriffiths:
The way that the necessary information is being passed into the callout partial has changed (in the POC it was an object with `.title`, `.type` and `.body` attributes, however it is now a hash) so we need to reflect these changes in the template itself.


haml/_callout.html.haml
```diff
@@ -1,4 +1,4 @@
-.{class: "cads-callout cads-callout-#{notice.notice_type}"}
+%div{class: "cads-callout cads-callout-#{notice[:type]}"}
   %h3
-    %span.cads-callout-label= notice.title
-  = sanitize notice.body, scrubber: scrubber
+    %span.cads-callout-label= notice[:title]
+  = sanitize notice[:body], scrubber: scrubber
```
panosvoudouris: Will this enclose the body into a P tag or do we need to add one here?

clairgriffiths: This leads to the output as so:

```"<div class='cads-callout cads-callout-urgent'>\n<h3>\n<span class='cads-callout-label'>Test callout</span>\n</h3>\n<h2>Heading</h2>\n<p>This is the body of a test callout</p>```

PR #39, opened 2020-03-03
NP-11 greedy navigation

baseonmars:
Stand alone JS/CSS version of greedy nav. 

This version does not use React - thinking behind this is that navigation should be available as soon as possible and load as few resources as possible.

Not yet tested with rails so may require some revision to integrate smoothly.

Design: https://www.figma.com/file/IZY9uqY9RPjhIUv8ol1qjf/DS_Work-In-Progress?node-id=159%3A474

Ticket: https://citizensadvice.atlassian.net/browse/NP-11

Preview:
menu

### DIFF
--- a/styleguide/component-wrapper.js
+++ b/styleguide/component-wrapper.js
@@ -2,12 +2,16 @@ import hljs from 'highlight.js'; // eslint-disable-line
 import beautify from 'js-beautify'; // eslint-disable-line
 
 const a11yid = 'a11yComponentToTest';
-const wrapper = (title, component, usage) => {
+const wrapper = (title, component, usage, js) => {
     const source = hljs.highlightAuto(
         beautify.html_beautify(component, {
             indent_size: 2
         })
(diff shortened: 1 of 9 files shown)


styleguide/component-wrapper.js
```diff
@@ -2,13 +2,17 @@ import hljs from 'highlight.js'; // eslint-disable-line
...
 const a11yid = 'a11yComponentToTest';
-const wrapper = (title, component, usage) => {
+const wrapper = (title, component, usage, js) => {
     const source = hljs.highlightAuto(
         beautify.html_beautify(component, {
             indent_size: 2
         })
     ).value;
 
+    if (js && js) {
```
shahzainb: Not sure I understand what's happening here...

baseonmars: left over from a revision where I'd used an object - cheers - I'll clean it up.

PR #51, opened 2020-03-12
NP-16 - design system automation framework

celonicolosi:
The latest version of image comparison test + CONTRIBUTING.md updated with BackstopJS information


.gitignore
```diff
@@ -13,4 +13,9 @@ yarn-error.log
 .npmrc
 *.gem
 /dist
-vendor
\ No newline at end of file
+<<<<<<< HEAD
```
panosvoudouris: This is a merge conflict, please fix.

backstop_data/engine_scripts/chromy/clickAndHoverHelper.js
```diff
@@ -0,0 +1,43 @@
+module.exports = function (chromy, scenario) {
```
panosvoudouris: All this stuff looks like it was brought in as is? If that is the case is there an npm package we can reference them from vs copy/paste?

.gitignore
celonicolosi: Merge fixed!

backstop_data/engine_scripts/chromy/clickAndHoverHelper.js
celonicolosi: Actually, the folder 'engine_scripts' is related to custom scripts to run the tests. I added backstop as a dev dependency on package.json because this was missing.

I removed the folder 'html_report' because every time we run a new test, there's going to be newly generated files there and we don't need to commit those.

And finally, I added the folder 'bitmaps_reference' since this is the source of truth when someone runs the test.

I updated gitignore files accordingly, so everything should be fine now! 😄

PR #60, opened 2020-03-18
NP-52 header

panosvoudouris:
Responsive header with revealable search bar on small screens.


styleguide/tokens/icon_font.md
```diff
@@ -0,0 +1,23 @@
+# Icon Fonts
+
+Icons are made available via the `cads-icons` font. To use them apply the appropriate class to the element that is displayin them. For example, to display the left chevron  <span class="cads-icon-arrow-left"></span> you would use <span class="html inline">&lt;span class="cads-icon-arrow-left"&gt;&lt;/span&gt;</span>.
```
baseonmars: "that is displayin**g** them"

soniaturcotte:
Looks great, just a few comments

**Visual:** 
On desktop, the button label should be bold. 

I just tested this in Voiceover with Safari. 

**Search**
On mobile, the icon reads just as "button" 
I think we need an `aria-label`. [NHS](https://www.nhs.uk/) uses a similar interaction so might be a useful reference

On desktop, when the focus is on the input it reads "Search through site content, search text field, search" which is quite long and repetitive.

**Logo**
It currently reads out "back to homepage" which won't make sense if the user is already on the homepage. 

We should either have "Citizens Advice homepage" on the logo when they are on the homepage. If that's too complicated, then just change the text for that everywhere.

shakiradamji:
Looks good although on the designs the search icon that turns into a close is styled as a secondary button. The first screenshot shows the designs.
Screenshot 2020-03-19 at 09 16 20
---
Screenshot 2020-03-19 at 09 16 09

styleguide/tokens/icon_font.md
panosvoudouris: No it ain't

panosvoudouris:
> ## Looks good although on the designs the search icon that turns into a close is styled as a secondary button. The first screenshot shows the designs.
> Screenshot 2020-03-19 at 09 16 20
> 
> Screenshot 2020-03-19 at 09 16 09

So it is primary to open and secondary to close?

shakiradamji:
Screenshot 2020-03-19 at 10 49 39
Screenshot 2020-03-19 at 10 49 42

@panosvoudouris give me a call if you need me to explain further but those screenshots show the different states ^^

panosvoudouris:
> Looks great, just a few comments
> 
> **Visual:**
> On desktop, the button label should be bold.
> 
> I just tested this in Voiceover with Safari.
> 
> **Search**
> On mobile, the icon reads just as "button"
> I think we need an `aria-label`. [NHS](https://www.nhs.uk/) uses a similar interaction so might be a useful reference
> 
> On desktop, when the focus is on the input it reads "Search through site content, search text field, search" which is quite long and repetitive.
> 
> **Logo**
> It currently reads out "back to homepage" which won't make sense if the user is already on the homepage.
> 
> We should either have "Citizens Advice homepage" on the logo when they are on the homepage. If that's too complicated, then just change the text for that everywhere.

I added the missing labels

shakiradamji:
@panosvoudouris looks good, 2 things i'm still noticing. The 'search' text on the button for bigger screens should be bold and the search / close icon for smaller screens should be the primary button colour.

soniaturcotte:
Two little things. 

- There is a problem with the logo on mobile. It's not the same height as the search option. (The box is but not the actual graphic, see screenshot)
Screenshot 2020-03-23 at 15 34 10


- On voiceover on mobile, when the search is expanded, it should read out that it's expanded so users know they should tab down for search. (See NHS for example)

PR #61, opened 2020-03-19
NP-172 nav buttons

panosvoudouris:
Add up/down arrow next to "More" button in navigation bar. Uses the icon fonts.


shahzainb:
LGTM. Perhaps we can add documentation about using RVM (ruby version manager) to manage ruby versions (http://rvm.io/). I say this because as devs working on multiple projects, you may have a ruby version that is greater than the one that comes pre-installed on the Mac, which then may break compatability in some gems between projects.

shakiradamji:
Looks good, a couple of things I noticed: 

1) We spoke about this on the call - more button should be the same styles as the rest of the nav 
2) When I used it on Voiceover I got to the more menu and it read out 'button, to open button press caps lock + space) however actually the menu was already open and by doing that it actually closed the menu. Is this to do with the ARIA type button, where as it doesn't actually act like a button it's more of a dropdown?

soniaturcotte:
Two things: 

The styling for "active" isn't working. When you select the menu link with a mouse, you get this flicker. It's ok for focus but we may want to have active as something else. @shakiradamji  

On voiceover, the menu doesn't make sense. When you tab through it, it immediately reads out "close menu item" 

The behaviour should be  
> user tabs to link, read option to open menu
> user opens menu, read that menu is open, option to close menu
> user can tab down through options or close menu

We need to find a way to test this stuff with JAWs / other assistive technologies.

PR #62, opened 2020-03-25
NP-38 Create contact details

panosvoudouris:
Add contact details component. This is a very simple component with just title + body coming from Contentful.


shakiradamji:
@panosvoudouris looks good, only thing I've noticed: 

- [ ] There should be 24px paragraph spacing, each different contact option would be a new paragraph

panosvoudouris:
Closed as the content model is not yet defined.

PR #64, opened 2020-03-25
NP-241 copy fonts to dist folder

baseonmars:
When trying to consume the build `dist/lib.css` file in `public-website` the build fails as references to fonts are incorrect.

This is one way to fix this problem (at least I think it will) by copying the fonts into `dist/fonts`.

Another option would be to re-write the urls - this doesn't appear to be a feature of sass. Maybe we can explore using webpack to do this?

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -10,6 +10,18 @@ module.exports = async ({ config, mode }) => {
             use: ['style-loader', 'css-loader', 'sass-loader']
         },
         {
+            test: /\.(woff|ttf|eot|svg).*$/,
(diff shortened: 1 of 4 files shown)


panosvoudouris:
Is the issue when the fonts are consumed via the compiled css or the scss?

baseonmars:
@panosvoudouris take a look at the Jenkins build failure here: https://jenkins.devops.citizensadvice.org.uk/blue/organizations/jenkins/public-website/detail/PR-52/2/pipeline/

baseonmars:
Solves the wrong problem.

PR #66, opened 2020-03-26
NP-243 Design updates to components

soniaturcotte:
@shakiradamji and I went through the existing components and compared to figma design files. These are updates to bring them the same. 

### Changes: 

**Icon**
- Remove "redo" arrow from storybook display. We are removing this arror from the 're do' buttons and don't want to encourage its use.  

**Callouts:**
- Add margin-bottom to callouts so they are adequately spaced when next to other components
- Add content guidance as placeholder text instead of lorum ipsum

**Notice banner**
- Margin below as callout (above) and removing unnecessary padding from large viewports
- Label (notice/new) was missing on smaller viewport

**Footer:**
- Update header size 

**Inputs and radios** 
- Update styles on radios (active and selected border)
- Update height on inputs
- Update hint text size
- Show version without error message (future updates should include guidance and show different options (with or without hint text, etc)


haml/_callout.story.html.haml
```diff
@@ -1,20 +1,21 @@
+
```
panosvoudouris: This file should have been deleted actually.

PR #73, opened 2020-04-01
Updating button styles

soniaturcotte:
The mid grey box-shadow looked a bit fuzzy once we saw it in code. 
Screenshot 2020-04-01 at 12 24 45

Sonia and Shakira worked on some small design updates which keeps the box-shadow for affordance. 
Screenshot 2020-04-01 at 12 26 11

There are changes to hover and focus styles as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## <sub>v0.6.1</sub>
+
+#### _Apr. 1, 2020_
+
+* Update design of buttons
+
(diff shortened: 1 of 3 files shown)


shakiradamji:
@soniaturcotte focus state is missing the underline on the word. I'm happy to lose this though with the new changes I think it's really clear it's focussed without that. If we are losing it though we should update that on Figma. Realised this is also missing on the nav focus state.

PR #74, opened 2020-04-01
NP-39 revealables

panosvoudouris:
Adds targeted content expandable box.


shakiradamji:
@panosvoudouris If I click the close link at the bottom I then can't open and close it, is this a bug?

shakiradamji:
@panosvoudouris The targeted content should have a focus state, a box around the border that uses our focus colour. Please imitate current styles

soniaturcotte:
Looks great. On voiceover though, the "Close" button-link at the bottom can still be focused and reads out as "Group" (Safari + Voiceover) and "Button" (Chrome + Voiceover)

Looking at some docs for `aria-hidden`, is says that you aren't supposed to use aria-hidden on a focusable element. Even though it is a bit repetitive, I think we should still make that final "Close" button work normally.

panosvoudouris:
> Looks great. On voiceover though, the "Close" button-link at the bottom can still be focused and reads out as "Group" (Safari + Voiceover) and "Button" (Chrome + Voiceover)
> 
> Looking at some docs [[mozilla](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute)] for `aria-hidden`, is says that you aren't supposed to use aria-hidden on a focusable element. [W3C](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden) also advises extreme caution and says "is known to work inconsistently in browsers".
> 
> Given that, I think we should still make that final "Close" button work normally even if it may be a little repetitive.

Ok I added an aria label which says `Collapse <section title>`

PR #75, opened 2020-04-03
NP-237 Website feedback

panosvoudouris:
Website feedback component added. Note that this is used in the footer so the footer tests will have to be rerun @celonicolosi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* NP-237 Website feedback component
 * NP-39 Revealable panels
 
 ## <sub>v0.6.1</sub>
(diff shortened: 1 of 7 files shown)


panosvoudouris:
@soniaturcotte Not sure if we should make the entire sentence a link? Currently voice over reads "is there anything wrong with this page", then you have to move to the next element to say "let us know button".

soniaturcotte:
> @soniaturcotte Not sure if we should make the entire sentence a link? Currently voice over reads "is there anything wrong with this page", then you have to move to the next element to say "let us know button".

Yes, I think that's a good shout as the link doesn't make sense on its own (ie, if the user were reading out all links on the page). 

Could we also get the little icon the same size as the x-height of the font please 🙏

PR #76, opened 2020-04-06
testing: Move BackstopJS testing to separate package, some usability updates.

baseonmars:
Initial motivation was to reduce the install time of the main package to facilitate Jenkins builds.

Before:
```
$ time npm i --silent
...
Downloading Chromium r722234 - 116.4 Mb [====================] 100% 0.0s
...
real	2m0.689s
user	0m40.454s
sys	0m22.252s
```

After:
```
$ time npm i --silent
...
real	0m35.950s
user	0m32.280s
sys	0m18.557s
```

After taking a closer look at BackstopJS I realised there would be an advantage to running it in Docker for repeatability. I added some tooling around this. It would be possible to run the visual regression suite within Jenkins on a timed basis.

The included tooling prints a more user friendly log. If errors are present` CONTRIBUTING.md` has instructions on how to re-run without the tooling wrapper.

Backstop output


shahzainb:
This is great! Is there a way to run backstop JS on Jenkins, but it exports a report/screenshots that can be seen in the jenkins job as an artifact (I can handle that in the Jenkinsfile).

baseonmars:
@shahzainb Thank you :) It's set up to do this in this PR, although the tests run in docker the output is reported locally

The reports are output to `testing/backstop_data/ci_report` in XUnit format, bitmaps in `testing/backstop_data/bitmaps_test`.

PR #84, opened 2020-04-09
NP-290 success message

panosvoudouris:
Adds the success message component (see this https://www.figma.com/file/J9qfbS9n1pUM2dpRQnTXUD/Design-System-2.0?node-id=10%3A0&viewport=-1209%2C520%2C0.8197165727615356)


shakiradamji:
Wrong green used in Storybook, looks like it is using the primary button colour? 

- [ ] Change border to #0A8600 and the background to 10% of that same colour 

- [ ] Check that the padding around the edge is the same as callouts

PR #91, opened 2020-04-16
Fix failing publishing backstopjs report when tests fail

shahzainb:


shahzainb:
Thanks for the approval Dan. The build is now failing for known reasons i.e. the tests are broken, but the report is being generated properly. So I'll leave this open until the tests are fix and then I'll rebase on that.

PR #92, opened 2020-04-16
NP-239 advice feedback

panosvoudouris:
This is the advice feedback form as shown here https://www.figma.com/file/J9qfbS9n1pUM2dpRQnTXUD/Design-System-2.0?node-id=874%3A520

Currently missing the API call in the end to send the user replies (to be done later).


js/advice-feedback.js
```diff
@@ -0,0 +1,49 @@
...
+            // form.classList.remove('step1');
+            form.classList.add('step3');
+
+            // TODO Send respong to API
+            console.log('Send "page was useful" reply');
+        });
+
+        rg[1].addEventListener('change', () => {
+            // No
+            // form.classList.remove('step1');
```
shahzainb, line 19: Presumably this is commented out for a reason?

shahzainb:
Code looks fine to me mostly. We'll add approvals once the Accessibility issues and other people have reviewed as well.

Also a thought: should this be  in the design-system repo? Is it a reusable component or only public-advice specific?

Should the API call and the sending of the data be the responsibility of the design-system? I think not. Different apps will want to send data differently, to different sources, tagged differently. Maybe we let the apps handle the data part then? And we provide a callback function for the submit?

js/advice-feedback.js
panosvoudouris: To be decided if we show/hide the other steps.

panosvoudouris:
> Code looks fine to me mostly. We'll add approvals once the Accessibility issues and other people have reviewed as well.
> 
> Also a thought: should this be in the design-system repo? Is it a reusable component or only public-advice specific?
> 
> Should the API call and the sending of the data be the responsibility of the design-system? I think not. Different apps will want to send data differently, to different sources, tagged differently. Maybe we let the apps handle the data part then? And we provide a callback function for the submit?

Good questions. I'm not sure if this should be in the design system repo or not, it is effectively reusuable but will there be reuse? The API should ideally be the responsibility of the consuming app, however if everything goes to the same place then keeping it here makes it simpler to consume.

panosvoudouris:
>     * [ ]  the headings should be a H4 style so the same size as the paragraph text (16px mobile and 18px desktop) but bold at the moment they are showing as 20.25px

The headings are in line with the small/large radio group styles. Do these need to change now?

>     * [ ]  The font size on the submit button should be 16px on mobile and 18px on desktop. Currently it's showing 16px on desktop

The button is a standard button as used elsewhere. The button font size is 16px. Should the button font change 18/16 as the paragraph text?

>     * [ ]  An error doesn't show when I try to submit without selecting anything

What is the error message? The Figma file does not show an error requirement.

shakiradamji:
> @shakiradamji
> 
> > I've added a close link to the designs so the user can close the form and move back to the previous step.
> 
> Is that not a bit hard to see for someone visually impaired? Given that the user never wanted to "open" anything is "close" appropriate? Or should it say back? Should it be at the bottom of the form?
> 
> > ```
> > * [ ]  the headings should be a H4 style so the same size as the paragraph text (16px mobile and 18px desktop) but bold at the moment they are showing as 20.25px
> > ```
> 
> The headings are in line with the small/large radio group styles. Do these need to change now?
> 
> > ```
> > * [ ]  The font size on the submit button should be 16px on mobile and 18px on desktop. Currently it's showing 16px on desktop
> > ```
> 
> The button is a standard button as used elsewhere. The button font size is 16px. Should the button font change 18/16 as the paragraph text?
> 
> > ```
> > * [ ]  An error doesn't show when I try to submit without selecting anything
> > ```
> 
> What is the error message? The Figma file does not show an error requirement. What if none of the options are what the user thinks is wrong?

shakiradamji:
1) I don't think it should be at the bottom because it's easily missed, especially on mobile you would have to scroll all the way down to see it. Plus on a screenreader would you have to listen to the whole form before getting the option to go back? It's what gov have on their feedback so as they will have tested I think we should just copy them for now as we need a decision. I'll put it in the sheet so we know to user test it when we run our sessions. 

2) I don't know where 20.25px came from for the form headings as this isn't in Figma. Maybe it was changed down the line. Leave it as is for now and we can have a overall review of all the forms stuff when Sonia is back. 

3) Yes the button text size should reflect the paragraph size 16px for mobile and 18px for desktop. 

4) there should always be an error if the user hasn't entered in the correct information otherwise they won't know why the form isn't submitting. I'll mock up what the error screen should look like.

panosvoudouris:
@shakiradamji they're all there, did you pull the latest? The Close button is next to the 4 radio buttons and the error message appears if you try to submit without selecting one.

shakiradamji:
> @shakiradamji they're all there, did you pull the latest? The Close button is next to the 4 radio buttons and the error message appears if you try to submit without selecting one.

@panosvoudouris yes I can see them now, don't know what happened there. It looks great and sounds good on a screenreader, only 2 outlying things: 

- [ ] Missing the error summary at the top of the page, we should show this even if there's only one error 

- [ ] (accessibility) When I tab to the yes/no radios and then hit the left key to move to the no radio it skips me to the next step of the form so I'm unable to select the no option

panosvoudouris:
@shakiradamji ok I will add the summary. WIth regards to accessibility we'll need to discuss how this is expected to work/look as the browser automatically selects the tabs if you use the arrow keys, which triggers the change. I think the way keyboard navigation works is using the tab key + space to select.

panosvoudouris:
Closing and will merge with #111

PR #95, opened 2020-04-17
NP-312 Convert EM to REM

panosvoudouris:
All `em` sizes have been converted to `rem`, this ensures the sizes are as per the designs. There were a number of issues with `em` styles stacking up to incorrect font sizes and margins/paddings.

There are a number of backstop tests that fail, I have reviewed those and they look correct. @shakiradamji can you specifically review:

- Callouts
- Notice banners
- Radio
- Iinput
- Textarea
- Buttons

I will update the reference tests if these are ok.


scss/4-elements/_forms.scss
```diff
@@ -64,23 +68,25 @@ textarea,
 .cads-textarea {
     @include cads-interactive-element();
 
+    width: 100%;
+    max-width: 50%;
```
baseonmars, line 72: What do both together do when they're percentages?

panosvoudouris: Ah...that should be width 50%, there's the width: 100% for the small breakpoint further down. Will change it.

PR #112, opened 2020-04-30
NP-248 bullet list default style

panosvoudouris:
Makes bullets the default list style, this way the website renderer will not have to do any processing for lists in the body. Non-bullet lists are used mainly in components so this can be controlled separately via the design system.


shahzainb:
LGTM

PR #113, opened 2020-04-30
NP-325 Remove H4 from contact details

panosvoudouris:
Replaced H4 with P tag and appropriate styling from Contact Details as H4 tags should not be used.


shahzainb:
Very large PR. LGTM

panosvoudouris:
> Very large PR. LGTM

I know it took forever.

PR #114, opened 2020-05-05
fix: change Setup stage to Prepare [NP-336]

baseonmars:
Avoid naming clash with Setup stage


baseonmars:
Sometimes the Setup stage doesn't run which is where the build name is set but it's not clear why as there are no reserve names mentioned in the documentation.

PR #119, opened 2020-05-07
NP-352 Targeted content integration

panosvoudouris:
Use a normal for loop vs forEach to avoid need for polyfill. The consumer should in theory transpile it but this may not happen given that the package is in node_modules.

### DIFF
--- a/js/targeted-content.js
+++ b/js/targeted-content.js
@@ -3,7 +3,8 @@ const initTargetedContent = () => {
         const content = document.getElementsByClassName(
             'cads-targeted-content'
         );
-        content.forEach(item => {
+        for (let i = 0; i < content.length; i++) {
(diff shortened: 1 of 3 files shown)


js/targeted-content.js
```diff
@@ -3,7 +3,8 @@ const initTargetedContent = () => {
         const content = document.getElementsByClassName(
             'cads-targeted-content'
         );
-        content.forEach(item => {
+        for (let i = 0; i < content.length; i++) {
```
shahzainb, line 6: I think we should check if content is not null/undefined here and has a length before running the for loop to avoid weird stuff.

panosvoudouris: That's why I have it in a try block.

PR #123, opened 2020-05-12
NP-274 print styles

panosvoudouris:
CSS Print styles added to format components appropriately for printing.

Note that Safari does not respect @page so print margins won't work there.


shahzainb:
Nice, LGTM from Dev perspective.

PR #135, opened 2020-05-27
NP-411 search pagination

panosvoudouris:
Adds:
- Pagination results
- Pagination control

These are currently only used in the search page however they are generic enough to be in the design system. Only SCSS is available and the consumer will have to build the HTML/whatever markup based on the supplied example. HAML not supplied as currently these live as React components in `public-website`.


package.json
```diff
@@ -65,11 +65,11 @@
         "husky": "^4.2.5",
         "inquirer": "^7.1.0",
         "js-beautify": "^1.11.0",
-        "lint-staged": "^10.2.2",
+        "lint-staged": "^10.2.6",
         "markdown-loader": "^5.1.0",
-        "moment": "^2.25.3",
+        "moment": "^2.26.0",
```
baseonmars, line 70: Not sure if we're using this directly but moment is a HUGE dependency - we're likely better off using `date-fn` or something similar that just does what we want.

PR #138, opened 2020-06-03
NP-444 Website Feedback component

panosvoudouris:
Ensure last worf of link and icon wrap to new line together. Currently the icon can wrap on its own which does not look nice.


testing/package.json
```diff
@@ -9,7 +9,9 @@
     },
     "devDependencies": {},
     "scripts": {
-        "test": "npx backstop test"
+        "test": "npx backstop test",
```
baseonmars, line 12: This stuff already in the parent package.json, no big harm putting it here but two things to keep in sync.

panosvoudouris: I will remove to avoid confusion.

PR #145, opened 2020-06-08
NP-441 Add related conent styles and guidance

panosvoudouris:
Moved related content from `public-website` to the design system. This is supplied as CSS only, the consumer needs to do the implementation.


shahzainb:
LGTM

panosvoudouris:
@celonicolosi can you please check I've done the VRT correctly.

PR #155, opened 2020-06-22
Babel build process for js

panosvoudouris:
As we're going to have a TS build I thought we should have a proper build process for the JS we have in the repo as well. So I've restructured the repo a bit, javascript goes into `src`, either `js` or `ts` and babel is used for the `js` part. Typescript to be setup later.

Also, the output now goes into `lib`, which will include both js and ts. Should scss be there too (currently in `dist`)? Or should it all go into `lib`?


panosvoudouris:
Moved all output (including CSS) into `lib`.

PR #156, opened 2020-06-26
feature: Adviser callout and Expand All [NP-470][NP-477]

panosvoudouris:
This covers two tickets:
NP-470: Adviser callouts. New styles for adviser callout, including nested callouts. The Reference button (shown in designs) is not included as that requires underlying back end functioanlity to store adviser References to articles.
NP-477: Expand all targeted content: Adds tertiary button styles and Expand all button at the top of all adviser pages.


scss/6-components/_callout.scss
```diff
@@ -85,13 +89,52 @@
...
+                display: block;
+                /* stylelint-disable-next-line */
+                &::before {
+                    margin-bottom: $cads-spacing-3;
+                    background: $cads-language__brand-adviser;
+                    color: $cads-palette__white;
+                    padding: $cads-spacing-3;
+                    font-size: 0.875rem;
+                    font-weight: $cads-font-weight__bold;
+                    line-height: 1em;
```
baseonmars, line 115: mixed rem and em for font-size and line-height makes my head explode. Can we drop the unit on line-height for similar effect?

src/js/targeted-content.js
```diff
@@ -1,9 +1,15 @@
 const initTargetedContent = () => {
     try {
-        const content = document.getElementsByClassName(
+        const pageContents = document.getElementsByClassName(
+            'cads-main-content'
+        );
+        const pageContent =
+            pageContents && pageContents.length ? pageContents[0] : document;
```
shahzainb, line 7: Did we include a build system in DS? If so, then this line could be replaced by `const pageContent = pageContents?.[0] : document`. But need to make sure this will be polyfilled for earlier browsers.

panosvoudouris: Done

scss/6-components/_callout.scss
panosvoudouris: Done

shahzainb:
LGTM from dev!

PR #157, opened 2020-07-01
feature: Design tweaks [NP-534]

panosvoudouris:
- Related Content gets a top margin to make it start at the same height as the H1 title.
- H2 gets a top margin after targeted content and callouts.
- Notice banner template adjusted to account for p tags coming in with the content.

See [Advice Collection](http://localhost:6006/iframe.html?id=4-sample-pages--advice-collection&viewMode=story) sample page to check changes


shahzainb:
LGTM!

PR #159, opened 2020-07-02
chore: Typescript component and tools pipeline [NP-463]

baseonmars:
This commit adds a typescript build pipeline to the design-system.

For now we use the default typescript compiler to create systemjs compatible modules - the advantage here is we can build components and modules to a single file, or choose to build to multiple files for larger tools if needed with minimal configuration (read: no webpack).

Due to the way Storybook is built I was unable to get a working example of systemjs in the design system - this is an ongoing issue present with other libraries that we can choose to either solve (in another PR) or ignore for now. Within storybook webpack and babel are used to dynamically load typescript.

Placeholder code is left in as this has minimal impact on the design system while allowing us to have a working demonstration.


README.md
```diff
@@ -42,3 +42,16 @@ You will need to add a lookup path for `node_modules`, for example add
 to your `ApplicationController`. You can then reference the partials in your haml like this
 
 <pre class="html"><code>= render partial: "@citizensadvice/design-system/haml/breadcrumb"</code></pre>
+
+### Javascript components
+
+Most simple Design System Javascript components can be used by simply including them as part of your main bundle.
+
+For the more complex components we are providing them as a systemjs bundle - this allows us to package each component as a single file. If your projecy has requirements for a different module format please get in contact via slack so that we can discuss your needs and work on a solution.
```
panosvoudouris, line 50: `project` vs `projecy`

src/ts/greedy-nav/index.ts
```diff
@@ -1,7 +1,12 @@
+// this is just a placeholder file
+
 import { Config, defaultConfig } from './Config';
 
 class GreedyNav {
     constructor(config: Config) {
         Object.assign(defaultConfig, config);
+        console.log('greedy-nav loaded');
```
shahzainb, line 8: Can we remove this?

src/ts/greedy-nav/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
```
panosvoudouris, line 1: Why do we need a separate tsconfig file in here? Can the top level one not deal with it?

baseonmars: The top level one sets up the sub projects, and there's also a base config that this one extends.

we'll have one tsconfig.json per component - generally they'll just do what this one does - says "build everything to one file" but we also have the option to output to multiple files or override specific configuration if we need to.

The short answer is that this is the mechanism that allows us to build each component to a separate file. It uses a newish feature called Project References - make sense?

src/ts/greedy-nav/index.ts
baseonmars: Sure - it was just a non-invasive way of showing this works - but np. I'll remove it.

src/ts/greedy-nav/tsconfig.json
panosvoudouris: 👍

PR #160, opened 2020-07-02
Feature/responsive tables [NP-521]

panosvoudouris:
Create responsive tables.

The table will stack on small devices. It also stacks when printing. Long tables on the medium/large viewports will scroll.

VR tests have been updated.

Designs are here: https://www.figma.com/file/IZY9uqY9RPjhIUv8ol1qjf/DS_Experimentation-and-early-drafts?node-id=2065%3A1288&viewport=354%2C678%2C0.28301599621772766

The table is available as a template. Once this is released the relevant updates are to be made in `public-website`.

To see the tables in context run the branch and see this page http://localhost:6006/?path=/story/4-sample-pages--tables


src/js/tables.js
```diff
@@ -0,0 +1,36 @@
+const initTables = () => {
+    try {
+        const tables = document.getElementsByClassName('cads-table'); // eslint-disable-line
+
+        if (tables) {
+            for (let i = 0; i < tables.length; i++) {
+                const table = tables[i];
+                try {
```
panosvoudouris, line 8: Try around each table found to avoid breaking all tables if something goes 💩

src/js/advice-feedback.js
```diff
@@ -2,6 +2,11 @@ const initAdviceFeedback = () => {
     try {
         const d = document;
         const form = d.getElementById('cads-advice-feedback');
+
```
panosvoudouris, line 5: Unrelated bug spotted while debugging tables

haml/_table.html.haml
```diff
@@ -0,0 +1,16 @@
+.cads-table-container
+  %table.cads-table{ class: [table_data['caption'] && !table_data['caption'].empty? && 'cads-table__has-caption'] }
+    -if table_data['caption'] and not table_data['caption'].empty?
+      %caption= table_data['caption']
+
+    %tbody
+      %tr
+        - table_data['header'].each do |head|
```
baseonmars, line 8: Is there a reason this isn't in a `thead`?

scss/4-elements/_tables.scss
```diff
@@ -0,0 +1,129 @@
...
+        text-align: left;
+    }
+
+    th,
+    td {
+        padding: $cads-spacing-2;
+        font-weight: $cads-font-weight__regular;
+
+        @include cads-border(
+            b,
```
baseonmars, line 36: what's b and a here? Can you use descriptive name

src/js/tables.js
```diff
@@ -0,0 +1,36 @@
+const initTables = () => {
+    try {
+        const tables = document.getElementsByClassName('cads-table'); // eslint-disable-line
+
+        if (tables) {
+            for (let i = 0; i < tables.length; i++) {
```
baseonmars, line 6: Could use for..of loops here to make the code a little bit more terse. 💅

styleguide/haml_locals.rb
```diff
@@ -3,6 +3,120 @@
 # frozen_string_literal => true
 
 @locals = {
+  'table_data' => {
```
baseonmars, line 6: Not really needed here but we could introduce [faker.js](https://github.com/marak/Faker.js/) at some point to generate mock data - keep it consistent across components + generate data at extremes within parameters.

haml/_table.html.haml
panosvoudouris: Thead didn't seem to work that well for voiceover. Reading around it looks like it is more for styling/printning, with a normal TR voiceover read the columns better. I may be wrong though, need someone to test this stuff better.

scss/4-elements/_tables.scss
panosvoudouris: a/t/p/h/l/r guess what it does

We could but then we'll need to go change everything. TBH I'm not sure this cads-border mixin is necessary overall.

haml/_table.html.haml
baseonmars: That's fair, makes sense. we can always add it in if future features require it.

scss/4-elements/_tables.scss
panosvoudouris: Mixin removed as discussed on Slack.

celonicolosi:
Approved in Internet Explorer 11 on Windows 10 and iOS and Android devices. Attached some evidence in the Jira

PR #164, opened 2020-07-09
NP-557 dynamic nav links

panosvoudouris:
Gets the nav links dynamically as an array. No changes to styles or output so VRTs are the same.

### DIFF
--- a/styleguide/haml_locals.rb
+++ b/styleguide/haml_locals.rb
@@ -117,6 +117,46 @@
       ]
     ]
   },
+
+  'navigation_links' => [
+    {
+      'url' => 'https =>//www.citizensadvice.org.uk/benefits',
+      'title' => 'Benefits'
+    },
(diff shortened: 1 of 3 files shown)


styleguide/haml_locals.rb
```diff
@@ -117,6 +117,46 @@
       ]
     ]
   },
+
+  'navigation_links' => [
+    {
+      'url' => 'https =>//www.citizensadvice.org.uk/benefits',
```
shahzainb, line 123: `/benefits/` - trailing slash at the end of the URLs are important.

panosvoudouris: Will change, though this is just sample links for storybook.

panosvoudouris:
> While you're making changes to this, can you make the links relative? So remove the live domain host.

Done

PR #167, opened 2020-07-17
bug: font loading [NP-580]

panosvoudouris:
The Open Sans font was not imported so it fell back to helvetica. Added the missing CSS import.

All the VR tests had to be rebaselined!


scss/1-settings/_typography.scss
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&display=swap');
```
shahzainb, line 1: Hmm we shouldn't be getting fonts from googleapi. The problem is that, it has been down before and fonts have broken in prod. Also, fonts need to load asap, so having a request limits that. Plus, a lot of adblockers block googleapis.com which means our fonts would break. Can we serve it locally?

panosvoudouris: Downloaded and added locally like the icon font.

PR #177, opened 2020-07-31
feat: GreedyNav Modern Refactor [NP-464]

baseonmars:
Ok - so - some quite major changes here - GreedyNav is now written it typescript, there are also tests.

I think I have some comments to cleanup and I'd quite like to do some heavy re-naming ahead of any more involved structural refactoring.

- re-written in typescript
- tests added where functionality was refactored
- static methods and globals refactored to class and instance variables
- multiple menu support removed

**Update**

- PR comments addressed and actioned.
- removed `helpers.ts`.
- add additional compilation target of CommonJS for web pack users (turned out to be very easy to do).
-- added `index.ts` to make it easier to work with CommonJS.

There is still much more that can be done but as discussed the point of this refactor approach is to always have working code so we can stop and say "this is good enough". As we are looking to push for release soon I'm putting this forward as good enough (some clean up is inevitable - esp comments that I've left as messages to myself during the refactor).


src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+    navWrapperElement: HTMLElement,
+    navDropdownToggleSelector: string,
+    breaks: number[]
+): void => {
+    if (breaks.length < 1) {
+        const navDropdownToggle = navWrapperElement.querySelector<HTMLElement>(
+            navDropdownToggleSelector
+        );
+
+        if (navDropdownToggle == null) {
```
shahzainb, line 145: Is this `==` on purpose? Would a `if (!navnavDropdownToggle)` not work the same?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+
+/**
+ * Get innerwidth without padding
+ * @param element
+ * @returns {number}
+ */
+const getElementContentWidth = (element: HTMLElement) => {
+    const styles = window.getComputedStyle(element);
+    const padding =
+        parseFloat(styles.paddingLeft) + parseFloat(styles.paddingRight);
```
shahzainb, line 249: Would this work as you intend if the padding values are negative?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+    const totalWidth = getElementContentWidth(_this);
+    // Check if parent is the navwrapper before calculating its width
+
+    const restWidth = getChildrenWidth(_this) + offsetPixels;
+    const viewportWidth = viewportSize().width;
+
+    return { totalWidth, restWidth, viewportWidth };
+};
+
+export class GreedyNavMenu {
```
shahzainb, line 310: Shouldn't we include an interface?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
+import { blurEventName } from './helpers';
+import { Config, defaultConfig } from './Config';
+
+const root = window;
+
+const supports = !!document.querySelector && !!root.addEventListener; // Feature test
```
shahzainb, line 6: Is there a reason `supports` is up here and not in the Class?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+
+        this.totalWidth = 0;
+        this.restWidth = 0;
+        this.viewportWidth = 0;
+    }
+
+    init(): void {
+        // Feature test.
+        if (!supports && typeof Node === 'undefined') {
+            console.warn("This browser doesn't support priorityNav");
```
shahzainb, line 365: Mixed branding here. `priorityNav` or `GreedyNav`?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+export const insertAfter = (newNode: Node, referenceNode: Node): void => {
+    if (referenceNode.parentNode !== null) {
+        referenceNode.parentNode.insertBefore(
+            newNode,
+            referenceNode.nextSibling
+        );
+    }
+};
+
+const checkForSymbols = (str: string) => {
```
shahzainb, line 233: Again, why is this not a method on the Class?

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+                            this.navDropdownToggleSelector,
+                            this.settings.navDropdownLabelActive
+                        );
+                    }
+                }
+            });
+        }
+
+        const lastItemCloseHandler = (event: FocusEvent) => {
+            if (this.toggleWrapper === null) {
```
shahzainb, line 629: So here it's `===` and not `==` whereas above in all the cases, the wrapper toggle check was `==`. Should double check what is the intention.

src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -0,0 +1,1062 @@
...
+                            blurEventName,
+                            lastItemCloseHandler
+                        );
+                    }
+                }
+            );
+        }
+
+        /*
+         * Close menu when last item is selected
```
shahzainb, line 758: Seems to be a misplaced comment block...

src/ts/greedy-nav/GreedyNav.ts
baseonmars: I don't think so - it's implicit to the class - what purpose would it serve?

src/ts/greedy-nav/GreedyNav.ts
baseonmars: The reason some methods are class methods and others not... I refactored functions that relied on Global variables to be instance methods so that the class could encapsulate the global state.

The remaining functions are 'outside' of the class. I don't think it benefits anyone to move them to `helpers.ts` - if anything I could move the contents of helpers inside the main file.

As mentioned this is part of a refactoring process which leaves the code working. It would be possible to improve this further but I'm making a judgement call of effort vs value.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: Agreed - this and the classes should be fixed.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: See above.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: In this case it will either be null or a value, in other cases where `==` is used `undefined` is a posibility - I'll switch it to `!` in those cases as you seem to be more comfortable with type coercion to a boolean vs a null.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: Thanks - I'll get it cleaned up.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: the `==` is on purpose and `!` would work the same. I'll switch it over.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: I don't think you can have negative padding so it shouldn't be?

src/ts/greedy-nav/GreedyNav.ts
baseonmars: Actually I'm gonna leave null in explicitly - the only two values it can be is HTMLElement or null so It doesn't matter if it's `==` or `===`. I'll put `===` as it'll raise less conversation.

src/ts/greedy-nav/GreedyNav.ts
baseonmars: I removed the helpers.ts file - you're right - it wasn't providing any value, just added confusion. I think whomever (likely me, but who knows) continues to work on this can make a choice about a better structure for the code.

scss/6-components/_navigation.scss
```diff
@@ -25,15 +25,15 @@ nav,
         font-weight: $cads-font-weight__bold;
     }
 
-    .nav__dropdown-toggle.priority-nav__dropdown-toggle,
-    .nav__dropdown-toggle.priority-nav__dropdown-toggle:hover {
+    .nav__dropdown-toggle.cads-greedy-nav__dropdown-toggle,
```
panosvoudouris, line 28: Should all the `.nav__` classes become `cads-...`. That would keep the naming consistent within the DS.

src/ts/greedy-nav/greed-nav.scss
```diff
@@ -0,0 +1,66 @@
+/*
+ * Core styles for Cads GreedyNav
```
panosvoudouris, line 2: Shouldn't this live within the scss dir?

baseonmars: SCSS I co-located to group by function, but I can move it. no biggie. you think I should right?

scss/6-components/_navigation.scss
baseonmars: on it.

PR #189, opened 2020-08-12
feat: fix keyboard navigation in IE 11 [NP-465]

baseonmars:
Fixes keyboard navigation in ie11 by correctly checking document.activeElement.

I've also made some small refactors to name variables more clearly and use TS null checking features.

Finally a thicker default border is added to site nav items to prevent it 'jumping' when navigating.

JSDOM tests added, mostly as an example of how to write them - I had to make one fairly major change to allow this to work correctly which involves passing in a reference to document to GreedyNavMenu.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,12 @@
 {
-    "presets": ["@babel/preset-env", "@babel/preset-typescript"]
+    "presets": ["@babel/preset-env", "@babel/preset-typescript"],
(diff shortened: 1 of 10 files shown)


celonicolosi (approved):
Tested and approved locally! Sending to "Ready to Deploy"

panosvoudouris (approved):

PR #194, opened 2020-08-20
feature: Migration of Tests from Public-Website-Cucumber [NP-726]

luke-hill:
**Previously:**

All tests were apportioned into `features/design_system` in public-website-cucumber

**Now:**

All tests are in this repo under `testing/features` - note some files are under `testing/` like `cucumber.yml` and `.env` and the Core Gemfile/Rakefile has been modified to take on additional items.

People can run tests locally against the deployed `github.io` website. The next step is to get the .env configurable and working across all parameters, including browserstack.

**Assumptions**:

- Currently all developers were using ruby 2.6.5 - If this is wrong I can change it, but we should set something here.
- We install gems into `vendor/bundle` - So the gems are specific to this repo and not an rvm environment
-


luke-hill:
I've had to extensively modify the dockerfile to get C-extension gems installing. There may be a better way to do this (For example when we use a fatter template in public-website), but for now I've just used `apk` to add the packages in.

@michelesr can you let me know if I've done something wrong or if things can be optimised. I've spent quite a while googling replacements for apt packages so whilst this works, it might not be 100% elegant.

PR #202, opened 2020-08-28
chore: update prettier config and normalise formatting

baseonmars:
delete `.prettierrc` in favour of using `.editorconfig` as a source of truth (Prettier reads it)

Update `.prettierignore` (generated files excluded) and run `npx prettier --write *` over the repo to normalise formatting.

We now use `.editorconfig` as a source of truth, but **NOTE**, you **WILL** need to install prettier as editorconfig alone does not guarantee correct formatting.


panosvoudouris (dismissed):

luke-hill (approved):

shahzainb (dismissed):

shahzainb (approved):

luke-hill (approved):

.editorconfig
```diff
@@ -4,11 +4,12 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_size = 4
+indent_size = 2
 indent_style = space
 insert_final_newline = true
 max_line_length = 80
 trim_trailing_whitespace = true
+quote_type = "single"
```
luke-hill, line 12: This conflicts with the ca-style-guide gem. Not sure if that matters too much though.

davidsauntson (approved):
🎈 works for me in vscode

PR #203, opened 2020-08-28
feature: Dockerize and Run Cucumber tests in CI [NP-744]

luke-hill:
As title suggests, this should now enable the CI pipeline to run tests for the system.

Currently this will use the default `grid_ci.env` Config, which runs tests against the Dockerised chrome node in the jenkins grid spun up. This will be extended once this is running and proving useful.

This will run against a version of the codebase that is spun up at branch time, so it will always be testing the latest code!


docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     command: npm run styleguide:ci -- --quiet
     hostname: ca-styleguide.test
     ports:
-      - 6006
+      - '6006:6006'
```
baseonmars, line 11: I'm not convinced we should bind this port to localhost as users (developers) may already have the styleguide running and this would cause a clash. In this instance the docker container is only really here to support running tests, not to support development as we do not deploy design-system using docker.

testing/grid_ci.env
```diff
@@ -1,5 +1,6 @@
 BROWSER=chrome
 BROWSERSTACK=false
-HEADLESS=true
-HUB_URL=http://design-system-hub.test:4444
+HEADLESS=false
```
baseonmars, line 3: Should this be true? I have a feeling this might be a value I changed whilst trying to debug the branch when we paired.

luke-hill: No idea, Will set it back to be true. No issues really. Makes the tests slightly quicker so I suppose it makes sense.

baseonmars (approved):
looking good, we can do some more work on the docker/gem setup in future... make this repo a bit more mono-repo. @davidsauntson's got some nice changes on the way that help with that.

PR #205, opened 2020-08-28
feature: move vr testing into /testing/visual-regression

davidsauntson:
Reason for change: to allow `/testing` to be used as the base folder for all testing files by removing backstop specific files from here and into a new `/testing/visual-regression` folder.  This paves the way for adding WCAG tests into `/testing/wcag`

- create `/testing/visual-regression` and update `.gitignore` with new paths for vr testing output
- update vr testing paths in:
  - package.json
  - jenkins scripts
  - npm scripts
  - jenkins and docker scripts
- update documentation with correct paths

Note that the `package.json` files have been left in `testing` as these will be updated when pa11y etc is added in future tickets (once this is done I will also rename the `vr-test:install` task to be more generic).


baseonmars (approved):
Tested locally with `npm vr-test:test`. good stuff!

PR #206, opened 2020-09-01
feature: Specification - Contact Details [NP-717]

luke-hill:
This is a really small spec as it's static and almost completely configurable.

Only assertions from story / Panos' input was that the top item should be bold and there should be at least 1 paragraph at all times.

### DIFF
--- a/testing/features/contact_details.feature
+++ b/testing/features/contact_details.feature
@@ -1,0 +1,15 @@
+Feature: Contact Details component
+
+  The Contact Details component provides users with full contact information
+  so that they can contact the relevant authority via different methods
+
+  Background:
+    Given a Contact Details component is on the page
+
+  Scenario: Validate initial state
+    Then the header says "Contact Details"
+    And a contact details block of text is present
+
+  Scenario: Ensure component is of correct size/format
+    Then there is at least 1 paragraph
+    And the top paragraph is bold
--- a/testing/features/contact_details.feature
+++ b/testing/features/contact_details.feature
@@ -6,9 +6,6 @@ Feature: Contact Details component
   Background:
     Given a Contact Details component is on the page
 
-  Scenario: Validate initial state
-    Then the header says "Contact details"
-
-  Scenario: Ensure component is of correct size/format
+  Scenario: Validate state of component
     Then there is at least 1 paragraph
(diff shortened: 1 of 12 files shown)


testing/features/contact_details.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Contact Details component
+
+  The Contact Details component provides users with full contact information
+  so that they can contact the relevant authority via different methods
+
+  Background:
+    Given a Contact Details component is on the page
+
+  Scenario: Validate initial state
+    Then the header says "Contact details"
```
panosvoudouris, line 10: The header will be the contact name in bold, eg "Bail for Immigration Detainees".

panosvoudouris (changes requested):

testing/features/contact_details.feature
luke-hill: This is the page title header (h1). And on this page it says "Contact details" - Should it say something different?

panosvoudouris: That is the styleguide header, the component itself is within the red box. Just like all the other ones.

contact details

luke-hill: So the first paragraph which is bold, should always be a title for where to contact the people?

panosvoudouris: It really is up to what the content editor has put in. If they make the first line bold then fine, if not then it is what it is. This will be down to the guidelines that content editors get. The design system itself has no code to enforce styling.

luke-hill: Hi Panos, just realised this is still requiring changes. What needs editing? This line you've highlighted is the assertion at the top of the page. Not the one you've highlighted? Do I need to change the assertion/s below (1+ paragraph and bold), this was on the ticket in JIRA?

panosvoudouris: The check that the title is "Contact details" is irrelevant. The component is the one I highlighted. You're effectively checking part of the styleguide UI, not the component.

panosvoudouris (changes requested):

luke-hill:
Rightio. I've got that on every single component. So they need removing then, no problems.

panosvoudouris (approved):

PR #207, opened 2020-09-02
feature: Enable Browserstack [NP-765]

luke-hill:
Browserstack is now enabled for the design-system

It will only run on master, and will run to begin with 3 combinations of Chrome (Win10,Win7,OSX Mojave).

Once we've had this running for a **couple of weeks** we can do the following...

- Refactor runners to include more browser combinations
- Alter / add extra info dumps to slack
- Monitor timings to ensure we're not going too crazy (As we're still adding more tests)

Once we've had this running for a **few months** we can do the following...

- Start having artifacts publish automatically to slack
- Look into whether we should move the tests elsewhere
- Look into running browserstack directly on branches (Obviously a smaller subset of browsers)
- Research into getting it running on a timer

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,9 @@ quote_type = "single"
 [*.{rb, feature, haml, yml}]
 indent_size = 2
 
+[Jenkinsfile]
+indent_size = 4
+
 [*.md]
 max_line_length = 0
 trim_trailing_whitespace = false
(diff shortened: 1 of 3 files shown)


michelesr (approved):
LGTM

baseonmars (approved):
🚀

PR #208, opened 2020-09-03
bugfix: Browserstack CI fixes [NP-765]

luke-hill:
Since merging the PR for 765 (Introduce and use Browserstack, a couple of small issues have become more apparent. This PR fixes them.

- `Jenkinsfile`
  - Our archiving wasn't working properly because our code sits inside `testing`
  - The newly mounted dir should now be archived correctly
- `docker-compose.yml`
  - This wasn't passing downstream variables
  - The build was always using default driver config (Kind of making browserstack redundant).
- `capybara`
  - Has a bug where the patching for their endpoints for Firefox uses a private method. In Chrome they make the private method public, but don't do this for Firefox. I've patched this and raised an issue (Linked in patch).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,12 +91,12 @@ pipeline {
                 reportName: 'BackstopJS Report',
             ])
             archiveArtifacts([
-                artifacts: 'artifacts/screenshots/*.png, ' +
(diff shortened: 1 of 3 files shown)


michelesr (approved):
LGTM

PR #211, opened 2020-09-07
feat: run pa11y Accessibility tests during Jenkins build [NP-741]

baseonmars:
* Add Docker config for wcag-tests
** installing puppeteer in alpine is not straightforward so use a pre-build image and add pa11y
* Update pa11y scripts to generate docker compatible config
* Add pa11y-test to Jenkinsfile.


davidsauntson (approved):
looking 🔥

PR #212, opened 2020-09-08
feature: Add extra tests for Browserstack [NP-731]

luke-hill:
Add in a bunch of new test candidate drivers for branch testing. Also added to public-website


celonicolosi (approved):
LGTM! Agree with the list of OS/Browsers.

PR #213, opened 2020-09-09
chore: Update node to v14

panosvoudouris:
As public-website is being built with v14 I updated this to be v14 too. v12 will become maintnance only next month.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/erbium
+stable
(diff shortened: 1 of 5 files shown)


davidrapson (approved):
👍 LGTM

panosvoudouris:
The build script was broken due to some changes in the packages. Fixed now.

davidrapson (approved):

PR #214, opened 2020-09-09
hotfix: Remove some drivers whilst we investigate issues

luke-hill:
IE11 has regressed on window resizing
MAC Chrome looks flaky and is not performing some commands as expected.

Will remove triage and re-add extra tickets down the line


celonicolosi (approved):
LGTM!

PR #219, opened 2020-09-09
fix: navigation spacing [NP-816]

baseonmars:
* Site Navigation buttons no longer introduce excess spacing when tabbing
* Site Navigation no longer outline links in blue causing a confusing visual effect which does not align with Figma design
* Added visual regression tests so that appearance of Site Navigation button active/focus state. 
** note: tests are currently not run as part of CI due to instability.

**Before (THE OLD VERSION, YUK!)**

menu-before

**After (the new version, much preferred)**

menu-after


testing/visual-regression/backstop.json
```diff
@@ -403,6 +403,54 @@
...
+      "requireSameDimensions": true,
+      "viewports": [
+        {
+          "label": "phone",
+          "width": 320,
+          "height": 400
+        },
+        {
+          "label": "Small",
+          "width": 641,
```
luke-hill, line 439: are these numbers correct? It feels weird that we've gone from 320 x 400 to 641 x 200.

The ones below look weird too. But maybe only the width matters. I'm not particularly clued up on this area, so disregard if not relevant.

luke-hill (commented):
I had a play with this offline, and found the new tabbing to much cleaner and easier to use.

You mention hover states have changed, but they looked the same for me when I was mousing over the menu items (Or the difference was very small).

testing/visual-regression/backstop.json
baseonmars: I coppied these from the other navigation test, I can't comment either way. If I had to guess it's related to the triggering the greedy nav  behaviour.

davidsauntson (changes requested):
Pls can you update the changelog 🥥

soniaturcotte (approved):

davidsauntson (approved):
🦞 noice

PR #222, opened 2020-09-10
bugfix: Tests failing in Safari during jenkins build [NP-855]

celonicolosi:
As discussed in new_platform_builds, some tests were falling during the builds in Jenkins.
Example: https://jenkins.devops.citizensadvice.org.uk/job/design-system/job/master/150/console

This PR fixes the in the problem in some steps of buttons_steps and advice_feedback steps for Safari Browser.

### DIFF
--- a/testing/features/step_definitions/buttons_steps.rb
+++ b/testing/features/step_definitions/buttons_steps.rb
@@ -28,5 +28,10 @@ Then("the text color of the {button-type} Button changes") do |button|
 end
 
 Then("the Tertiary button text is less prominent than the {button-type} Button") do |button|
-  expect(@component.font_weight_of("tertiary")).to be < @component.font_weight_of(button)
+  if safari?
+    expect(@component.font_weight_of("tertiary")).to eq("normal")
+    expect(@component.font_weight_of(button)).to eq("bold")
+  else
+    expect(@component.font_weight_of("tertiary")).to be < @component.font_weight_of(button)
+  end
 end
--- a/testing/features/buttons.feature
+++ b/testing/features/buttons.feature
@@ -39,6 +39,6 @@ Feature: Buttons components
     Then the background color of the Tertiary Button changes
     And the text color of the Tertiary Button changes
 
-  Scenario: Tertiary button has less prominent text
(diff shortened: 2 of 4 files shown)


testing/features/step_definitions/buttons_steps.rb
```diff
@@ -28,5 +28,10 @@
 end
 
 Then("the Tertiary button text is less prominent than the {button-type} Button") do |button|
-  expect(@component.font_weight_of("tertiary")).to be < @component.font_weight_of(button)
+  if safari?
+    expect(@component.font_weight_of("tertiary")).to eq("normal")
+    expect(@component.font_weight_of(button)).to eq("bold")
```
baseonmars, line 33: why do we test appearance in the behavioural tests? This makes them brittle. Would using Backstop for this make more sense?

baseonmars (commented):
ne comment about testing approach, but otherwise looks good.

testing/features/step_definitions/buttons_steps.rb
davidsauntson: 💯 agree

Maybe we can just delete it from this suite of tests? It will already be covered by the VR tests, right?

panosvoudouris: Yes, behaviour like that will be much better tested with the VRTs as they check what the font actually looks like vs whether the CSS attemptes to make it bold or not.

celonicolosi: I commented this non-behaviour scenario about the font-weight, just in case Luke wanted to use it for another purpose. If not, we are going to remove it permanently. In this PR there is another fix in a behavioural scenario and some changes in local.rb, I'm keeping these changes.

davidsauntson (approved):
@luke-hill  - please review when you return.  We should look to remove any visual testing from the behavioural tests.

testing/features/step_definitions/buttons_steps.rb
luke-hill: I'm happy to remove this scenario. Think we need a follow-up PR to tidy up the safari bit below.

PR #223, opened 2020-09-11
bugfix: add focus styles to error summary [NP-856]

davidsauntson:
Reasons for change:
Error summary must be focusable and have correctly styled links with focus.  This PR address those two issues.

Changes:
 - correct focus styles on error summary and links within
 - add accessibility attrs and links into haml template
 - documentation updates to cover how the error summary should behave

### DIFF
--- a/haml/_error_summary.html.haml
+++ b/haml/_error_summary.html.haml
@@ -3,4 +3,9 @@
     There is a problem with
     %span.cads-error-summary__field_count 2 fields
   %ul.cads-list__no-bullet.cads-error-summary__list
-    %li Item 1
+    %li 
+      %a{:href => "#"}
+        Enter an email address in the correct format, like name@example.com
+    %li
+      %a{:href => "#"}  
+        Enter your full name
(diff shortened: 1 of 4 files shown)


soniaturcotte (approved):
💯

haml/_error_summary.html.haml
```diff
@@ -1,6 +1,11 @@
-.cads-error-summary{"aria-live": "assertive"}
-  %p.cads-heading-small
+.cads-error-summary{"aria-live" => "assertive", "aria-labelledby" => "error-summary-title", :role => "alert", :tabindex => 0}
+  %p.cads-heading-small{:id => "error-summary-title"}
     There is a problem with
     %span.cads-error-summary__field_count 2 fields
   %ul.cads-list__no-bullet.cads-error-summary__list
-    %li Item 1
+    %li 
+      %a{:href => "#"}
```
davidrapson, line 7: These links have no href, should they? Would they normally link to the field with an error?

davidrapson (approved):
👍 With a question about error links

haml/_error_summary.html.haml
davidsauntson: They should give focus to the erroring field (I put this in the usage notes section for this component).  I left it out here as I figured it was an implementation detail (eg I am doing it in React in the Email This Page component, but that might not be the case for other consumers).

I could make it more explicit here tho, if you think that would be useful? eg `<a href="#give-focus-to-input">...</a>`

PR #224, opened 2020-09-11
Upgrade Storybook to v6 [NP-714]

davidrapson:
Upgrade Storybook to v6. Thanks to the `npx sb upgrade` command for sorting the package versions for me automatically.

From the docs:

> presets.js has been renamed to main.js. main.js is the main point of configuration for storybook.
> config.js has been renamed to preview.js. preview.js configures the "preview" iframe that renders your components.
> addons.js has been renamed to manager.js. manager.js configures Storybook's "manager" UI that wraps the preview, and also configures addons panel.

In addition to structural changes the following config changes were needed:

- v6 support zero-config Typescript so `storybook/preset-typescript` is no longer required
- `withA11y` is no longer required, this code is injected by default with the add-on


package.json
```diff
@@ -11,6 +11,7 @@
         "README.md"
     ],
     "scripts": {
+        "start": "npm run styleguide",
```
davidrapson, line 14: Added this by convention as I expected to be able to run `npm start`, aliases the existing styleguide command.

davidsauntson (approved):
Looks good to me.  

Only thing I noticed when running locally is that when you click on the 'Open this in new tab link', it opens something else.  Fine after I cleared my cache, just a heads up.

davidrapson:
> Only thing I noticed when running locally is that when you click on the 'Open this in new tab link', it opens something else (ie a different component). Fine after I cleared my cache, just a heads up.

Cheers, will see if I can replicate that.

This upgrade has already solved one bugbear for me. Prior to v6 storybook seemed to open in Chrome whenever you ran it. I don't use Chrome. Thankfully the upgrade now respects your default browser when running it (@soniaturcotte might appreciate this)

davidrapson:
@davidsauntson Have reliably been able to replicate the open in new tab issue. I noticed this was our own custom code but storybook provides a great default toolbar with this and a lot more. I've switched to use that.

**Before**

image

**After**

image


The second button from the right opens in a new tab, and reliably works.

davidsauntson (dismissed):
🥮 nice

panosvoudouris:
> @davidsauntson Have reliably been able to replicate the open in new tab issue. I noticed this was our own custom code but storybook provides a great default toolbar with this and a lot more. I've switched to use that.
> 
> **Before**
> 
> image
> 
> **After**
> 
> image
> 
> The second button from the right opens in a new tab, and reliably works.

We removed that bar and switched to the text as it was taking up a bit of space for effectively just that button. Also, non devs were finding it a bit confusing and where not sure how to open a component in a new window vs the text under the header. Check with Sonia to confirm which of the two is acceptable.

panosvoudouris (changes requested):
See my comment about the bar.

Also add an entry to the top of the changelog for this ticket.

The addon panel has disappeared? Has the a11y checker.

panosvoudouris (approved):

PR #226, opened 2020-09-15
Button active state [NP-289]

davidrapson:
Revises button states. Heavily inspired by https://service-manual.nhs.uk/design-system/components/buttons

buttons

~Basic approach for now but going to pair with @minniessuyuhuang to get something we are happy with.~ Edit: Updated with revised version.

- [x] Update primary and secondary button styles based on https://codepen.io/serenitybeam/pen/NWNBvbw and https://service-manual.nhs.uk/design-system/components/buttons
- [x] Review tertiary button styles in relation to new primary and secondary buttons
- [x] Add / update visual regression tests
- [x] Re-test cucumber scenarios
- [x] Test in context of `public-website`
- [x] Update changelog


davidsauntson (commented):
🙂 looks good!

Do you want to write a VR test for this too?

davidrapson:
@davidsauntson Yep! Definitely will, but planning to hold off until I've paired on this with Minnie.

davidrapson:
Revised button animation based on work with @minniessuyuhuang 

buttons

We are going to take a second look at tertiary button styles as they relate to the new button animation. For now these changes focus on primary and secondary buttons.

.stylelintrc.json
```diff
@@ -1,49 +1,50 @@
...
-        "string-quotes": "single",
-        "value-no-vendor-prefix": true,
-        "plugin/selector-bem-pattern": {
-            "preset": "bem",
-            "componentName": "ds-(([a-z0-9]+(?!-$)-?)+)",
-            "componentSelectors": {
-                "initial": "\\.{componentName}(((__|--)(([a-z0-9\\[\\]'=]+(?!-$)-?)+))+)?$"
-            }
-        }
+  "extends": "stylelint-config-standard",
```
davidrapson, line 2: Prettier auto-formatted this (from 4 spaces => 2 spaces). The notable change is to disable the pedantic `declaration-empty-line-before` rule. Particularly where there are larger numbers of styles, as in _buttons.scss, some deliberate spacing around properties is useful for grouping rules.

scss/4-elements/_forms.scss
```diff
@@ -123,8 +123,8 @@ textarea,
 
   .cads-radio-button-checkmark {
     @include cads-interactive-element();
-    @include cads-transition-animation(none);
 
+    transition: none; /* stylelint-disable property-disallowed-list */
```
davidrapson, line 127: Calling `cads-transition-animation(none)` was a long way of saying `transition: none`

scss/6-components/_search.scss
```diff
@@ -33,6 +34,14 @@
     .cads-lang-cy &::after {
       content: 'Chwilio';
     }
+
+    &:active {
```
davidrapson, line 38: This resets styles from the new default buttons. This is probably fine, but do wonder if it's preferable to special-case the search bar and inline just the required styles rather than unsetting things. Not a big deal.

scss/6-components/_targeted-content.scss
```diff
@@ -88,9 +88,6 @@
   .cads-targeted-content__expand-all {
     display: inline-block;
 
-    @extend .cads-button;
```
davidrapson, line 91: Noted this in the commit but this caused problems with refactoring. Extend is tricksy and best avoided, see https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/. Simple fix here was to explicitly add the required classes to the expand button in the JS.

styleguide/tokens/animations.md
```diff
@@ -1,10 +1,10 @@
 # Animations
 
-All animations are 0.3s long. This is defined in the <span class="html inline">$cads-animation-speed</span> variable. However you should beware of reduced motion user preferences. These can be checked using the <span class="html inline">prefers-reduced-motion: reduce</span> media query. For defining transitions use the <span class="html inline">cads-transition-animation</span> mixin. This will automatically ignore transition animations if the user has set the reduce motion option. You can supply the prop to animate or if none is supplied <span class="html inline">all</span> is used. Example:
```
davidrapson, line 3: This previously read:

> "You can supply the prop to animate or if none is supplied 'all' is used". 

This wasn't true even before my changes. Not sure this is intentional. Updated the docs to reflect the defaults are `background` and `border`

testing/features/support/pages/buttons.rb
```diff
@@ -4,9 +4,9 @@ module DesignSystem
   class Buttons < SitePrism::Page
     set_url "/iframe.html?id=3-components--buttons&viewMode=story"
 
-    element :primary, "[type='button']:not(.cads-button)"
-    element :secondary, "[type='button'].cads-button.cads-button__secondary"
-    element :tertiary, "[type='button'].cads-button.cads-button__tertiary"
+    element :primary, ".cads-button.cads-button__primary"
```
davidrapson, line 7: Updated these tests because these selectors are incorrect based on my latest changes. However these seem to be doing the job of visual regression testing but by introspecting styles which seems weird?

luke-hill: So these are doing some functional tests (Which backstop basically doesn't do). I've not dived heavily into backstop but these tests aren't ones we have replicated there.

The styling here is just used as locators (Backstop looks at either a large portion of a page or the page proper).

scripts/vr-test.js
```diff
@@ -6,7 +6,7 @@ const { log, error } = console;
 const ci = process.argv.some((arg) => arg === '--ci');
 
 const baseCommand =
-  'cd testing/visual-regression && npx backstop test --docker';
+  'cd ../testing/visual-regression && npx backstop test --docker';
```
davidrapson, line 9: Had to update this path to be able to run backstop tests.

scss/4-elements/_buttons.scss
```diff
@@ -1,71 +1,111 @@
...
+
+  font-weight: $cads-font-weight__bold;
+  font-size: 1.125rem;
 
   @include cads-media-breakpoint-down(sm) {
     font-size: 1rem;
   }
 
-  &,
-  &__primary {
```
davidrapson, line 21: Notable change here is moving to longhand selectors instead of `&__primary`. This had a notable bug because we had `<button>` in the parent selector in that is generated classes like `button__secondary` which is not a real element. Removed this for two reasons:

1. To avoid generating unexpected classes as mentioned above
2. Making it possible to do a basic search in Sass files for things like `.cads-button__secondary`

A secondary change is no longer styling `<button>` as a "Button". As these revised button styles have more prominent active states and a raised effect hooking this on to button elements conflicted heavily with places where `<button>` is used separately to "Buttons" e.g. link-buttons.

scss/4-elements/_links.scss
```diff
@@ -22,12 +27,13 @@ a,
...
-.cads-linkbutton {
+button.cads-linkbutton {
+  // Reset button styles when styled like a link link
+  -webkit-appearance: none;
   border: none;
   background: transparent;
   padding: 0;
   box-shadow: none;
 
-  @extend .cads-heading-small;
```
davidrapson, line 31: `.cads-heading-small` was being almost 100% overrides by `. cads-hyperlink`. Another reason why extends are tricksy. Decoupled this from the main button styles completely as it was overriding so much already.

scss/6-components/_pagination.scss
```diff
@@ -5,35 +5,40 @@
...
-  button {
-    color: $cads-language__link-colour;
-    border-color: $cads-language__border-colour;
-    box-shadow: none;
-    font-weight: $cads-font-weight__regular;
-    background-color: transparent;
-    min-width: $cads-interactive-target-size;
-    margin-bottom: 0;
+.cads-paging__button {
+  @include cads-button-defaults();
```
davidrapson, line 9: New button styles conflicted with pagination styles as they were inheriting from that. Rather than unsetting lots of things I opted to use a mixin to share _just_ the common details for these sorts of buttons. It also fixes a bug as a result as pagination now has a proper focus style.

.stylelintrc.json
davidsauntson: Not too sure what ruleset you're after but `always first-nested` might be better than turning it off completely?  [https://stylelint.io/user-guide/rules/declaration-empty-line-before](https://stylelint.io/user-guide/rules/declaration-empty-line-before)

scss/4-elements/_buttons.scss
davidsauntson: see comment in #231

luke-hill:
Just looking at the animated gif. Are some of the behavioural tests valid now (I know they're passing on node-chrome internally on the grid).

3 of the tests state the button changes after a click - Are these still present?
3 of the tests state the button changes after a hover - These seem still present

davidrapson:
@luke-hill Hopefully screenshots are a little clearer than the animated gif:

**Primary button**

Default:
primary-button-default

Hover:
primary-button-hover

Active:
primary-button-active

Focussed:
primary-button-focussed

**Secondary button**

Default:
secondary-button-default

Hover:
secondary-button-hover

Active:
secondary-button-active

Focussed:
secondary-button-focussed

luke-hill:
Good stuff. Probably worth monitoring once we release and re-test using browserstack. But I think it should be ok if the behaviour is unchanged (i.e. there is a click change as well as hover).

davidsauntson (approved):

PR #228, opened 2020-09-15
Bug: fix page scroll [NP-806]

panosvoudouris:
The site had a horizontal scroll at certain sizes. This was due to the invisble show/hide navigation button still having a size.

The bevaiour can be replicated both in storybook (launch the navigation into its own page) and `public-website`. See the currently hosted styleguide vs this branch to compare.


styleguide/styles.scss
```diff
@@ -220,3 +220,7 @@
   border: 1px solid $cads-language__border-colour;
   margin-bottom: 1rem;
 }
+
+.sb-show-main {
```
panosvoudouris, line 224: Looks like the new storybook version adds extra padding to the body. Removed it to keep the look (and VRTs) consistent.

celonicolosi (approved):
Tested and approved.

davidsauntson (approved):

PR #231, opened 2020-09-17
Update stylelint config

davidrapson:
Ran into two common Sass pitfalls when working on https://github.com/citizensadvice/design-system/pull/226. One relating to use of `@extend` and one relating to union selectors e.g. `.class { &-union  { } }`.

- Use stylelint.config.js to allow for commenting rules
- Add rules for @extend and union selectors with rationale

Happy to chat through my rationale in more detail, but have included some context in comments in the stylelint config.

Union selectors in particular caused the biggest issue. As well as IMO making the codebase harder to search they can generate some wonky CSS when combined with element selectors. For example the scss:

```scss
button,
.cads-button {
  // Some styles

  &__primary {
    // Some more styles
  }
}
```

Generates the following class names:

```css
button {}
button__primary {}
.cads-button {}
.cads-button__primary {}
```

One of those (`button__primary`) is an invalid element.


davidsauntson:
My tuppence worth:

- I've used union selectors in large code bases and haven't had a problem finding stuff - source maps and well named files should be enough.  I'm a bit reluctant to generate a warning every time it's used as it is useful in some circumstances, especially for BEM.  This rule is in a file named `_buttons.scss` which is where I would expect to find it
- Agree we shouldn't be generating selectors for things that don't exist - looks like the solution here is to use `%` placeholder and `@extend`?
- I'd like to know roughly how many warnings this generates?  Long term I'd like to get to 0 warnings for DS and PW (but obvs this isn't really a priority)

Would like to hear what  @panosvoudouris  and @baseonmars think too 😃

davidrapson:
On the `@extend` point, placeholder selectors are still allowed with this rule. It only warns against extending random classnames. The full list of warnings for extends is:

```
scss/4-elements/_links.scss
 31:3  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 32:3  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/4-elements/_misc.scss
 12:3  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/4-elements/_tables.scss
 18:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 33:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 44:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_callout.scss
 14:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_footer.scss
  7:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 13:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_navigation.scss
  18:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 142:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_notice-banner.scss
  6:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 22:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_related-content.scss
 18:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

scss/6-components/_targeted-content.scss
 29:7  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 91:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 92:5  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder

styleguide/styles.scss
 174:3  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
 181:3  ⚠  Expected a placeholder selector (e.g. %placeholder) to be used in @extend   scss/at-extend-no-missing-placeholder
```

Most of these warnings are really examples where mixins should be used instead. Either things like `@extend .cads-heading-large` or genuine issues like `@extend .cads-list__no-bullet` which is extending a child selector of some list styles.


FWIW I've found exactly the opposite with union selectors and so am fairly against them for that reason. Couple of gnarly examples from the codebase

```scss
.cads-callout {
  &.cads-callout- {
    &example {}
    &standard { }
 }
}
```

and 

```sass
.cads-greedy-nav {
  &-has-dropdown {
    .cads-greedy-nav {
      &__dropdown {
        &-toggle {
          position: relative;
        }
      }
    }
  }
}
```

Personally have to really think about what classnames that was generating vs writing them out explicitly.

The full list of warnings for union selectors is:

```
scss/_print.scss
 31:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

scss/4-elements/_buttons.scss
 21:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 32:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 44:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 59:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 60:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 61:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 63:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 64:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 65:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

scss/6-components/_advice_feedback.scss
 20:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 21:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 22:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

scss/6-components/_callout.scss
 26:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 54:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 59:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 86:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

scss/6-components/_navigation_greedy-nav.scss
 30:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 45:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 47:7  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 55:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 59:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 63:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

scss/6-components/_targeted-content.scss
  6:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 49:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
 71:3  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name

styleguide/styles.scss
 156:5  ⚠  Unexpected union class name with the parent selector (&)   scss/selector-no-union-class-name
```

Bear in mind that I've removed quite a few of these with https://github.com/citizensadvice/design-system/pull/226 so it's not such big list. 

Hope that helps clarify what I'm thinking here :neckbeard: 

Agree on the 0 warnings point, ideally we'd set these rules to error eventually.

panosvoudouris:
If we think this kind of code:
```
.cads-callout {
  &.cads-callout- {
    &example {}
    &standard { }
 }
}
```
is problematic we should refactor it. Most of those probably do not need to be nested tbh as the `cads-` takes care of that. However do consider how they are used in the templates, as a consumer of the design system you should probably never really have to set these up separately on components, or if you do there's a template and/or example.

panosvoudouris:
How would you propose we fix this issue:
```
button {}
button__primary {}
.cads-button {}
.cads-button__primary {}
```

We could probably remove the `button`? Thought bear in mind that for some elements we do want default styles, when the content parser in ruby generates the markup we want to have as few classes as possible, eg `ul` tags or `p` etc. Buttons are probably not the case though so a default style is probably not necessary.

panosvoudouris:
Another thing, while something like 
```
.cads-greedy-nav {
  &-has-dropdown {
    .cads-greedy-nav {
      &__dropdown {
        &-toggle {
          position: relative;
        }
      }
    }
  }
}
```
looks pretty rubbish beware that some stuff was copied over from whatever external library/old code/etc. So making changes without looking at why they are what they are will cause issues. It would be good to improve but is likely not a simple thing (or maybe it is).

davidrapson:
> We could probably remove the button? Thought bear in mind that for some elements we do want default styles, when the content parser in ruby generates the markup we want to have as few classes as possible, eg ul tags or p etc. Buttons are probably not the case though so a default style is probably not necessary.

I did exactly this in https://github.com/citizensadvice/design-system/pull/226. Removed `<button>` from the default styles as it's definitely a case where having it by default causes unwanted elements to inherit. No objections to styling default elements in general though, especially where generated markup is concerned.

davidrapson:
Going to close this as I don't think we have consensus so happy to chalk this up to personal preference.

PR #233, opened 2020-09-21
Re-enable jest and backstop in CI

davidrapson:
Whilst working on https://github.com/citizensadvice/design-system/pull/226, @davidsauntson and I noticed that visual regression tests were not running in CI. Upon further digging it doesn't look like Jest is being run either.

Attempts to (re)enable `jest` and `backstop` testing in CI.

- Exclude `vendor` from jest test paths to ensure `jest` can run
- Split visual regression CI scripts from linters and unit tests
- Run `jest` and `backstop` in CI
- Update backstop snapshots which are out of sync since last CI run

May want to view this with [`?w=1`](https://github.com/citizensadvice/design-system/pull/233/files?w=1) when reviewing as there is some prettier auto-formatting going on here.


davidrapson:
Have one difference between generated snapshots and those run in CI:

image

This is a test which is generated based on a click event. The change is a small difference in the resulting hover state and which radio item is moused-over. Need to account for this.

luke-hill:
If these are proving problematic. We have this behaviour covered in the Regular Behavioural tests in a fair bit more detail. So removing this scenario doesn't feel like a loss to me.

See https://github.com/citizensadvice/design-system/blob/master/testing/features/advice_feedback.feature

Also @celonicolosi did some work setting up Backstop, he might be able to shed light on how we mitigate these issues.

davidrapson:
@luke-hill Testing a _couple_ more things, as it mostly looks like it needs a stable place to take the snapshot. But if that doesn't work work I agree with you, this particular test is far more a behavioural test and can do without the visual test. We have visual tests for all of the form element that make up this component anyway.

davidrapson:
Have removed the one flaky test. Not worth blocking getting these running again over one flaky test that we have covered elsewhere.

davidrapson:
Verified that a failing unit test breaks CI

image


image

davidrapson:
Verified that global css change fails CI by changing the background colour of the page:

image

image

davidrapson:
Pretty happy with where this is at now. Definitely better than not running any of these tests in CI 😄

luke-hill (approved):
I've not reviewed specific js stuff. But all the script stuff seems great (You've also fixed a couple of bugs in there too!)

LGTM

luke-hill (approved):

PR #235, opened 2020-09-22
Revert "Tag button features with future release"

davidrapson:
Reverts citizensadvice/design-system#234

Still failed on `master` on the same tests even though I'd tagged them with `@future_release` 🤷 

image

### DIFF
--- a/testing/features/buttons.feature
+++ b/testing/features/buttons.feature
@@ -6,34 +6,33 @@ Feature: Buttons components
   Background:
     Given the Buttons component is on the page
 
-  @future_release @v1.9.4+
   Scenario: Validate initial state
     Then all the buttons are present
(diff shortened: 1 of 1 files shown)


davidrapson:
Re ran the build on `master` and now the **opposite** tests fail!

image


I think the main issue is because the _selector_ has changed, so probably _all_ button scenarios will be out of sync between master and the released version. Is it super weird to mark the whole lot as `@future_release` for now?

luke-hill:
Sadly you are right @davidrapson ? But we can unmark it once we do a deployment. So it's not a huge issue.

davidrapson:
Closing in favour of https://github.com/citizensadvice/design-system/pull/236

PR #239, opened 2020-09-23
Untag button feature specs

davidrapson:
Now that v1.10.0 has been released I should be able to untag the button feature specs as `@future_release`

### DIFF
--- a/testing/features/buttons.feature
+++ b/testing/features/buttons.feature
@@ -6,37 +6,30 @@ Feature: Buttons components
   Background:
     Given the Buttons component is on the page
 
-  @future_release @v1.9.4+
   Scenario: Validate initial state
     Then all the buttons are present
 
-  @future_release @v1.9.4+
   Scenario: Primary Button changes color when hovered over
     When I hover over the Primary Button
     Then the background color of the Primary Button changes
 
-  @future_release @v1.9.4+
   Scenario: Primary Button changes color when clicked on
     When I click on the Primary Button
     Then the background color of the Primary Button changes
 
-  @future_release @v1.9.4+
   Scenario: Secondary Button changes color when hovered over
     When I hover over the Secondary Button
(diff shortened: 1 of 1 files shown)


davidsauntson (changes requested):
These look like more visual tests in the behavioural test suite.

Let's delete them and replace them with equivalent tests in the vrt suite.

luke-hill (approved):

luke-hill:
These tests are doing actions (Denoted by the When steps). Ideally we should be testing these inside the cucumber suite, as the puppeteer stuff isn't as reliable with actions. (Also we cover more test candidates with browser/os combos).

davidrapson:
Seconding @davidsauntson's suggestion here. The cucumber tests here only assert that the colours are different based on the CSSOM which is not the same as confirming they _actually_ look right with a visual regression test. Should definitely have some behavioural tests but not sure they are an ideal fit for this sort of use-case.

davidsauntson:
@luke-hill I hear what you're saying and agree about the multiple browser issue, but they are testing hover states, which is a visual test.  

If backstop isn't working we should look at replacing it or fixing it - @davidrapson has used Percy before, which would cover the multiple browser issue.

luke-hill:
So we've looked into these tests just now, and not only can you not do them in BackstopJS they actually give you false confidence (As they don't work well with multiple actions).

I suggest for now we keep these tests here as a good source of truth with behavioural tests, and either look down the line to migrate the visual tests into the BDD suite using the Algorithm/Gem I made or look towards something like Percy (But percy would just wrap capybara which is the same thing).

davidsauntson:
OK, let's remove the hover states then and since the button update the backgrounds don't change on click anyway, so they can all go. 👍

luke-hill:
Dealt with in #254

PR #241, opened 2020-09-24
Feature: advisernet related content [NP-756]

panosvoudouris:
Add related content styles and template for Advisernet. The current related content component is distributed just as CSS and the consumer has to build it. As the Advisernet component is a bit more complicated (switches between an equivalent PW view and a Targeted Content in small viewports) this is now provided as a template.

Not sure if this should be in the design system tbh as it is a bit too specific.

Also, there is now a `public`/`advisernet` SCSS file to import as well as an equivalent compiled CSS file. The old `lib` is retained for compatibility, this is the same as importing `public`.


luke-hill (commented):
Given this is a new component? on the design-system. Do we have an idea of the following spec for what we would test on this?

davidsauntson (approved):
🎆

panosvoudouris:
@minniessuyuhuang provided the following points for review/changes:

*Large breakpoint*
[] The margin-bottom for the title ‘related content’ should be 16px
[] The margin-bottom for the top-level link should be 16px
[] There is a 16px padding on the left-hand side of the whole container, what is this for? As this component will be on the left-hand side of the page on AdviserNet, the padding should be on the right-hand side of the container and it should be 24px.
[] The spacing between each link (margin-bottom) should be 16px
[] Do we have the ‘visited’ state for the text links?
[] Can we try adding an adviser box and see what it would look like?

*Small breakpoint*
[] The font-size for the top-level link should be 16px, not 20px
[] The padding for the expanded container should be 16px on all sides (currently the top and the bottom are 12px)
[] There should be no ‘close’ button at the bottom of the component 
[] The spacing between the last-child of the list and the bottom edge of the container should be 16px.

panosvoudouris:
> Given this is a new component? on the design-system. Do we have an idea of the following spec for what we would test on this?

This is a presentation component with different style for advisernet. Not sure what you need to check beyond design review and VRT?

davidrapson (commented):
Given this is only a small amount of CSS might it be worth us sticking with one stylesheet for now and monitoring the file-size?

styleguide/haml_locals.rb
```diff
@@ -243,5 +243,25 @@
   <p><a href="https =>//www.citizensadvice.org.uk/immigration/staying-in-the-uk-after-brexit/keeping-your-family-in-the-uk-after-brexit/">Check if you can apply to the EU Settlement Scheme</a></p>'
   },
 
-  'success_message' => 'Thank you for your feedback'
+  "success_message" => "Thank you for your feedback",
```
davidrapson, line 246: Worth me showing the approach I've tried in https://github.com/citizensadvice/design-system/pull/242? It would be good to show both versions of the related content component and we can avoid the need to put data in this locals file.

styleguide/styles.scss
```diff
@@ -1,5 +1,5 @@
 // These styles are for use in the storybook styleguide
-@import '../scss/lib';
+@import '../scss/advisernet';
```
davidrapson, line 2: This feels like a reason to stick with one stylesheet for now. If we have to use the one with the most stuff in to get storybook working correctly?

davidsauntson:
> Given this is only a small amount of CSS might it be worth us sticking with one stylesheet for now and monitoring the file-size?

Think we all agreed to have one css file for PW and one for AN in the mtg we had?

davidrapson:
@davidsauntson I think so, but seeing the actual output makes me wonder if it's overkill.

> Not sure if this should be in the design system tbh as it is a bit too specific.

Think I agree with this and this could be in advisernet. Feels particularly OK as the component is mostly combining design system components in a unique way. Actual design updates look great just having second thoughts about where it should live.

davidsauntson (dismissed):

panosvoudouris:
I've left the two stylesheets, `public` and `advisernet`, however `lib` now is equivalent to `advisernet`. So all options are available and it makes no difference to consumers right now.

panosvoudouris:
Will wait for design review before requesting dev/qa.

panosvoudouris:
Reviewed and approved by @minniessuyuhuang

davidrapson (approved):
👍 LGTM although we might want to come back and update the stories to use mdx.

(Not sure about having merged lots of package updates into this PR, but not a big deal)

PR #242, opened 2020-09-25
Storybook MDX docs

davidrapson:
As demoed earlier today, this is an experiment to find a way to be able to show multiple variations of a component and document stories using markdown whilst still retaining Haml partials.

## New sidebar structure

image

## Docs view

image

## Usable code snippets

Structured so these could be copied directly into a rails app:

image

## Haml usage notes written in markdown

image

## Technical notes

**Pros:**

- Avoids the need to use `locals.rb`
- Allows multiple variants of each component
- Makes it possible for designers to write documentation for components

**Cons:**

- We are still working _against_ storybook a little by using Haml partials so there's still a little bit of a dance to get some of the previews to work

👩‍💻 👨‍💻  Recommend checking this out and having a play for yourselves

🗒️  I'll add explanatory comments for some of the structural changes

~⚠️ I've included my initial work on OISC warnings in here as a working example. Looking for reviews on the storybook structure  at this stage, not this component. If we are happy with the approach I'll remove that work from this PR so we can merge this without shipping early work on the warnings component.~ Updated to remove OISC warnings from this PR to keep things simple.


styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+/>
+
+# Callouts
+
+There are several different variants of the callout component:
+
+## Default
+
+import templateDefault from './_default.html.haml';
+import templateDefaultCode from '!!raw-loader!./_default.html.haml';
```
davidrapson, line 16: `!!raw-loader!` is what allows us to show useful code snippets. Loads the mini Haml story partials as raw text.

styleguide/components/oisc-warnings/_default.html.haml
```diff
@@ -0,0 +1 @@
+= render partial: "@citizensadvice/design-system/haml/oisc_warning", locals: { title: "This is an OISC level 2 topic", content: "Generalist advisers can give clients the information on this page, but must refer them to an immigration specialist if they need advice. <a href='#'>Check what to do if your client needs level 2 advice.</a>" }
```
davidrapson, line 1: Using tiny Haml partials for the stories themselves allows us to bypass the need for `locals.rb` and write examples exactly as we would in Rails.

styleguide/components/oisc-warnings/oisc-warning.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+- Content designers can input free text into the heading fields
+- Content designers can input free text and links into the paragraph fields
+
+import templateDefault from './_default.html.haml';
+import templateDefaultCode from '!!raw-loader!./_default.html.haml';
+
+<Preview>
+  <Story
+    name="With title and description"
+    parameters={{ docs: { source: { code: templateDefaultCode } } }}
```
davidrapson, line 34: This little dance sets the code snippet. The default behaviour is to return the template string which resolves to a literal `() => templateDefault`. Wish this syntax was simpler but think It's worth it to be able to show good code-snippets.

styleguide/styles.scss
```diff
@@ -103,36 +103,6 @@
   }
 }
 
-.cads-styleguide__breakpoint-label {
```
davidrapson, line 106: `@storybook/addon-essentials` includes the docs addon but also a responsive breakpoint addon so switched to that in place of this breakpoint label.

haml/_oisc_warning.html.haml
```diff
@@ -0,0 +1,19 @@
+-# OISC Warning
```
shahzainb, line 1: Correct me if I'm wrong, but it does not look like this PR is only documentation changes? You seem to be committing OISC feature work in this PR too. Can we split that out into it's own PR?

src/js/oisc-warnings.js
```diff
@@ -0,0 +1,54 @@
+const initOISCWarnings = () => {
+  const supportsInterSectionObserver =
```
shahzainb, line 2: And what happens in browsers that don't support it?

shahzainb (changes requested):

src/js/oisc-warnings.js
davidrapson: > ⚠️ I've included my initial work on OISC warnings in here as a working example. Looking for reviews on the storybook structure at this stage, not this component. If we are happy with the approach I'll remove that work from this PR so we can merge this without shipping early work on the warnings component.

haml/_oisc_warning.html.haml
davidrapson: > ⚠️ I've included my initial work on OISC warnings in here as a working example. Looking for reviews on the storybook structure at this stage, not this component. If we are happy with the approach I'll remove that work from this PR so we can merge this without shipping early work on the warnings component.

davidrapson:
Updated to remove OISC warnings from this PR to keep things simple.

styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+<Preview>
+  <Story
+    name="Default callout"
+    parameters={{ docs: { source: { code: templateDefaultCode } } }}
```
davidrapson, line 21: This little dance sets the code snippet. The default behaviour is to return the template string which resolves to a literal `() => templateDefault`. Wish this syntax was simpler but think It's worth it to be able to show good code-snippets.

davidsauntson (approved):
This is very nice 🌵

davidsauntson (approved):

soniaturcotte:
This is a fantastic improvement to Storybook! Makes it much easier for non-devs to navigate and also see all the different formats for a component. 

I have two questions: 
1. Is it possible to have the "Docs" tab first? I think it will be useful to get an overview first, and then be able to explore an individual variant in more depth if desired. I'm also worried people won't notice the docs tab. It's not that big of a thing, we can always do demos to share the features. 
2. I know you haven't gotten to the forms yet, but will this format allow us to show the different states (with or without hint, with an error, etc)? That would be super useful as well! 

Either way, this is great. Thanks!

soniaturcotte (approved):

minniessuyuhuang (approved):

davidrapson:
> Is it possible to have the "Docs" tab first? I think it will be useful to get an overview first, and then be able to explore an individual variant in more depth if desired

I _think_ so, let me double check

> I know you haven't gotten to the forms yet, but will this format allow us to show the different states (with or without hint, with an error, etc)? That would be super useful as well!

Yep! (I did a throwaway experiment for exactly that when I was trying this new format out)

davidsauntson (approved):

PR #243, opened 2020-09-25
Feature: reconcile search with designs [NP-895]

davidsauntson:
Reason for change:
Designs for search had changed since they were first implemented in the design system.  This PR updates the style of the search bar to match those in the design file.

Changes:
 - remove icon in mobile search button and replace with text
 - added VRT for expanded header
 - update VRT references


baseonmars (dismissed):

scss/6-components/_navigation.scss
```diff
@@ -1,6 +1,5 @@
 @import 'navigation_greedy-nav';
 
-nav,
```
davidrapson, line 3: 💯

davidrapson (dismissed):
👍

soniaturcotte (approved):
Looks good for search page! 👍  thanks for doing that. Just to confirm this won't affect the header?

davidsauntson:
> Looks good for search page! 👍 thanks for doing that. Just to confirm this won't affect the header?

@soniaturcotte Collapsed it is the same, expanded it will show 'Search' instead of the icon:
image

davidsauntson:
> > Looks good for search page! 👍 thanks for doing that. Just to confirm this won't affect the header?
> 
> @soniaturcotte It will show Search instead of the icon in the header on desktop as it currently does.
> 
> On mobile, collapsed it is the same, expanded it will show 'Search' instead of the icon.
> image

I've changed this now so it's back to the search icon on mobile, and added a VRT to cover this.

davidrapson (approved):
👍 Looks great. 😍 Thank you for removing the default element styles

PR #244, opened 2020-09-28
OISC warnings component [NP-472]

davidrapson:
Note: _Branched off https://github.com/citizensadvice/design-system/pull/242 as the documentation support changes are needed to be able to show and document the multiple variants of this component._

Adds OISC warnings component, documenting multiple variants:

- Default warning with title and description
- Title only variant
- Optional sticky behaviour

The sticky variant can be disclosed and uses IntersectionObserver (with a polyfill) to show a close button when the warning is stuck to the screen.

**GIF showing sticky behaviour**
disclose

**GIF showing sticky behaviour in IE 11**

ie11-disclose

## Checklist

- [x] Initial component
- [x] Document variants
- [x] Test in IE 11
- [x] Add visual regression tests
- [x] Test within public website via `npm link`
- [x] Design review

Note: Definitely planning on squashing my changes when merging.


package.json
```diff
@@ -106,7 +106,9 @@
         "typescript": "^3.9.6",
         "webdriverio": "^6.4.6"
     },
-    "dependencies": {},
+    "dependencies": {
+        "intersection-observer": "^0.11.0"
```
davidrapson, line 110: 👀 Open question about how we include these kinds of polyfills. For now I've included it here and imported it directly into `oisc-warnings.js`. This this a reasonable approach?

testing/visual-regression/backstop_data/engine_scripts/puppet/onReady.js
```diff
@@ -1,7 +1,16 @@
...
+  console.log('SCENARIO > ' + scenario.label);
+  await require('./overrideCSS')(page, scenario);
+  await require('./clickAndHoverHelper')(page, scenario);
 
-    // add more ready handlers here...
+  /**
+   * Wait for fonts to be loaded.
+   * Ensure that icon fonts are ready to avoid display inconsitencies
+   */
+  await page.waitFor(() => {
```
davidrapson, line 10: Added this based on https://github.com/garris/BackstopJS/issues/1144#issuecomment-588558285

Needed to ensure that icon fonts have been loaded otherwise icon-only buttons may not display. I actually don't think this is an issue per-se with Backstop and more an issue in general with icon fonts, as they have to load as a separate font request. Icon fonts have [plenty of their own issues](https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/) and longer term I think it'd be worth using inline SVG icons, but for now this is a workaround.

testing/visual-regression/backstop.json
```diff
@@ -714,6 +714,20 @@
       "expect": 0,
       "misMatchThreshold": 0.1,
       "requireSameDimensions": true
+    },
+    {
+      "label": "Components/OISC Warning/With title and description",
```
davidrapson, line 719: One benefit of the new component docs structure is that there's no need to hide elements or use `a11yComponentToTest` as the Canvas is _only_ the component.

testing/visual-regression/backstop.json
```diff
@@ -714,6 +714,20 @@
...
+      "url": "http://host.docker.internal:6006/iframe.html?id=components-oisc-warning--with-title-and-description&viewMode=story"
+    },
+    {
+      "label": "Components/OISC Warning/Title only",
+      "url": "http://host.docker.internal:6006/iframe.html?id=components-oisc-warning--title-only&viewMode=story"
+    },
+    {
+      "label": "Components/OISC Warning/Sticky (scrolled)",
+      "url": "http://host.docker.internal:6006/iframe.html?id=components-oisc-warning--sticky&viewMode=story",
+      "selectors": ["viewport"],
```
davidrapson, line 729: This `["viewport"]` magic selector is needed to test scroll behaviour. Default backstop behaviour is to capture the full element regardless of viewport height.

haml/_oisc_warning.html.haml
```diff
@@ -0,0 +1,19 @@
+-# OISC Warning
+
+-# Set defaults for parameters
+- title ||= ''
+- body ||= ''
+- is_sticky ||= false
```
baseonmars, line 6: Is this a good idea? It strikes me as an opportunity for errors (missing params) to be silently swallowed by a consumer of the design-system.

davidrapson: This is a good point, I think it's correct for `body` which is optional, and `is_sticky` as a silent default is _intentional_.  But let me test without it for `title`, as that should really throw an error if not provided.

package.json
davidrapson: When I run a build only the `require` statement for this polyfill is included, so definitely open to suggestions on how best to include this here (or should it be a requirement for the consumer?) @baseonmars @davidsauntson

davidrapson:
This is at the point where I could with some 👀s on it, tested this against public website and got it to a point where I think we can do an initial release. Particularly interested in views on including the `intersection-observer` polyfill here vs. on the consumer end.

soniaturcotte:
Looks great, and works well with voiceover and safari. One thing I noticed is the heading link thing doesn't work on Safari (though it worked well on Firefox Nightly and Chrome). Not sure why that is, but already a huge improvement from existing version, so thank you.

davidrapson:
@soniaturcotte Forgot to note that down. For the header behaviour I am using `scroll-margin` https://caniuse.com/?search=scroll-margin which is a tidy way of setting an offset for this kind of use-case. However excludes any version of Safari (for now) and old IE. Generally decent support though. Is that a reasonable trade-off for an enhancement like this?

scss/6-components/_oisc-warning.scss
```diff
@@ -0,0 +1,65 @@
+/* =========================================================================
+   The Office of the Immigration Services Commissioner (OISC)  Warnings
+   ========================================================================= */
+
+.cads-oisc-warning {
+  background-color: $cads-palette__light-grey;
+  border: 1px solid $cads-palette__dark-grey;
+  margin-bottom: $cads-spacing-3;
+
+  .cads-oisc-warning__title {
```
davidsauntson, line 10: does this need to be nested?  this will generate:
`.cads-oisc-warning .cads-oisc-warning__title`

If not wouldn't just `.cads-oisc-warning__title` be better? 

so either un-nested or use the union operator:
```
.cads-oisc-warning {
   ...

   &__title {
      ...
   }
}

**Review by davidsauntson — COMMENTED** [review]
👍 small comment about avoiding nested SCSS unless you give in to the dark side and use the union operator 😉

File: scss/6-components/_oisc-warning.scss (hunk 6 above)
Comment (davidrapson) [reply scss/6-components/_oisc-warning.scss:10]: Personally, I think the nesting can be a useful defensive technique. i.e. by nesting it's not possible to accidentally use `cads-oisc-warning__title` without the parent `cads-oisc-warning` class for example. Not such a problem here I guess when the markup is mostly included from a partial, but a useful pattern nonetheless.

If the nesting is undesirable, I'd opt to just not nest things, e.g.

```scss
.cads-oisc-warning {
    margin: 2ch;    
}
.cads-oisc-warning__title {
    color: palegoldenrod;
}
```

davidsauntson: Agreed to avoid keep specificity as low as possible unless we find a reason not to.  Will keep a track of instances where BEM element selectors are used outside of their block.

davidsauntson (approved):
❤️

PR #245, opened 2020-09-28
Avoid need for backstop sync script

davidrapson:
Backstop supports a .js config, so rather than needing to sync .json config to handle different baseUrls I've created a shared config which each variant can call with the appropriate baseUrl.

As a side-effect this also simplifies the pa11y sync script.


testing/wcag/.pa11yci.local.json
```diff
@@ -10,8 +10,6 @@
     "http://localhost:6006/iframe.html?id=2-design-foundations--colour-language",
     "http://localhost:6006/iframe.html?id=2-design-foundations--colour-palette",
     "http://localhost:6006/iframe.html?id=3-components--advice-feedback",
-    "http://localhost:6006/iframe.html?id=3-components--advice-feedback",
```
davidrapson, line 13: Backstop often uses the same URL for multiple scenarios. I updated the pa11y script to convert the urls to a `Set` first to de-dupe this list. Avoids testing the same URL multiple times.

scripts/vr-test.js
```diff
@@ -7,7 +7,9 @@ const ci = process.argv.some((arg) => arg === '--ci');
 
 const baseCommand =
   'cd ../testing/visual-regression && npx backstop test --docker';
-const command = ci ? `${baseCommand} --config=backstop-ci.json` : baseCommand;
+const command = ci
+  ? `${baseCommand} --config=backstop-config-ci.json`
```
baseonmars, line 11: Whis should be `backstop-config-ci.js` right?

davidrapson: Yes it should!

scripts/update-pa11y-config.js
```diff
@@ -11,72 +11,54 @@
  * the visual regression tests. (.pa11yci.ci.json)
  *
  * Options for both configs can be added into the baseConfig object.
+ * @ TODO: Upgrade storybook to 6 and use JSON export
```
luke-hill, line 14: Storybook is v6 isnt it?

luke-hill (approved):

scripts/update-pa11y-config.js
davidrapson: This TODO was already here, I just moved it. We are on v6 so whilst we could probably do this now it was beyond the scope of this PR.

PR #246, opened 2020-09-28
Format package.json

davidrapson:
Original issue described in https://github.com/citizensadvice/design-system/pull/240

This PR brings the format of `package.json` back in line with `prettier` settings.

PR #247, opened 2020-09-29
Update backstop documentation

davidrapson:
Update CONTRBUTING and low-level usage to reference new backstop-config.js.

Thanks to @davidsauntson for the spot.


package.json
```diff
@@ -1,137 +1,135 @@
 {
-    "name": "@citizensadvice/design-system",
```
davidrapson, line 2: Prettier autoformat, see https://github.com/citizensadvice/design-system/pull/246

Actual changes in here are to correct a couple of utility commands. View with [?w=1](https://github.com/citizensadvice/design-system/pull/247/files?w=1)

davidsauntson (approved):

PR #248, opened 2020-09-29
feature:Breadcrumbs Spec [NP-623]

luke-hill:
Added Feature specification for Breadcrumbs

NB: I used the capybara helper `has_link?` for links, as the specific URL's contained aren't validatable (As they link off to invalid URLs) - These are then used as `have_link` for the assertion. But I don't provide more details, as they're irrelevant.

### DIFF
--- a/testing/features/breadcrumbs.feature
+++ b/testing/features/breadcrumbs.feature
@@ -1,0 +1,14 @@
+Feature: Breadcrumbs component
+
+  The Breadcrumbs component allows users to see a navigable path to their
+  previous content pages they have traversed through
+
+  Background:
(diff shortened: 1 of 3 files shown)


panosvoudouris:
In small viewports you only get a single link to the previous page. Does that need to be covered?

luke-hill:
At the moment no because it's only testing at one resolution. But this is a good point that we should cover.

I'll create a breakout ticket to cover the existing scenarios at multiple breakpoints. Thanks for letting me know though. Good to have knowledge of.

panosvoudouris (approved):

PR #249, opened 2020-09-30
bug: Callout headings bug [NP-924]

panosvoudouris:
Remove the code that adds the Expand All targeted content for advisernet. This now lives in public-website as it is too coupled with code there for it to be in the design system.


davidrapson (approved):
👍

PR #252, opened 2020-09-30
Backstop config cleanup

davidrapson:
Backstop generates a lot of boilerplate code when you initialise a new project. This PR removes all unnecessary and un-ran code:

- Removes all unneeded backstop boilerplate from `engine_scripts`. Down to a single `onReady` script
- Removes unnecessary defaults from backstop scenarios
- Emulates `prefers-reduced-motion` instead of overriding all `transform` styles


luke-hill (approved):
I was going to do a load of this after finally starting to get to grips with it. Loads of the json stuff was placeholder / hitting already defined defaults. So good job!

davidrapson:
@luke-hill Cheers! Finally clocked that "this is mostly a wrapper around puppeteer" so what could stay and go all clicked.

PR #271, opened 2020-10-05
[Experimental] i18n proof of concept

davidrapson:
Proof of concept to load rails-i18n as part of the haml loader so we can include translations as part of haml components.

Screenshot 2020-10-05 at 09 24 23


scripts/haml.rb
```diff
@@ -1,6 +1,15 @@
 require 'rubygems'
 require 'bundler/setup'
 require 'haml'
+require 'rails_i18n'
+require 'action_view'
```
davidrapson, line 5: Including actionview so we can use the `t` shorthand found in Rails

scripts/haml.rb
```diff
@@ -1,6 +1,15 @@
...
+
+# Allow shorthand t() helper to be used
+include ActionView::Helpers::TranslationHelper
+
+I18n.available_locales = [:en, :cy]
+I18n.load_path += Dir['locales/*.yml']
```
davidrapson, line 11: AFAIK on the Rails / consumer side we'd need to do a similar thing and include `@citizensadvice/design-system/locales`  in the load path.

styleguide/components/translated/_example-cy.html.haml
```diff
@@ -0,0 +1,2 @@
+- I18n.locale = :cy
```
davidrapson, line 1: This is the one part I'd like to find a better approach for. Because we load haml via a webpack loader, there's not a great way to dynamically _change_ language outside of this inline Ruby hack. Would mean lots of little partials to show welsh variants in storybook. 💭 Thoughts welcome

styleguide/components/translated/_example-en.html.haml
```diff
@@ -0,0 +1 @@
+%h1= t('cads.name')
```
davidrapson, line 1: In Rails-land it looks like it's common to use relative path translations like `t('.name')` but we need to write them out long-hand in this content (possible to hack it with `@virtual_path` but that would be unlikely to be portable). Seems a fair trade-off.

styleguide/components/translated/_example-cy.html.haml
luke-hill: What about altering it when the user changes their language? And setting the locale then for the user based on the language they've switched to.

davidrapson: Because of the way we load haml files in storybook there's not an obvious _place_ to do that. But yeah, that'd be the ideal.

davidrapson:
This has a similar problem to https://github.com/citizensadvice/design-system/pull/272 in that the current docker commands use their own bundle `RUN bundle config --global silence_root_warning 1 && bundle init && bundle add haml && bundle install` so the only gem that works is `haml`. Need to update this somehow to respect the projects Gemfile

scripts/haml.rb
pedantic-git: This is a very elegant approach

davidrapson: To confirm I've tested this via `npm link` and adding `config.i18n.load_path += Dir[Rails.root.join("node_modules/@citizensadvice/design-system/locales/*.yml")]` to the application config. It's able to load design-system bundled locales in addition to any app ones.

image

scripts/haml.rb
```diff
@@ -1,6 +1,15 @@
 require 'rubygems'
 require 'bundler/setup'
 require 'haml'
+require 'rails_i18n'
```
davidrapson, line 4: Noting here that I'm including `rails_i18n` but in practical terms I think we are only really relying on the underlying `i18n` gems behaviour as we don't trigger any of the railties code in this context (as far as I know!). I still think this makes sense though as it's ultimately what will get used on the rails end.

styleguide/components/translated/_example-en.html.haml
pedantic-git: It's probably fine to use a fully-qualified path anyway, because they're more portable that way - I often find the shorthand to be a right pain when I move translations from one partial to another.

Maybe something like `cads.example.name` though so the full path is present, and you could have a `cads.shared.product_name` or whatever if there's content that's shared across components.

pedantic-git (commented):
Nice one - as a proof-of-concept it's great. The solution for sharing locale files between ds and the rails app is very elegant.

styleguide/components/translated/_example-en.html.haml
davidrapson: Ah yeah, I'd have assumed we'd use the exact component/shared format you describe. I was being a bit lax for the proof of concept 👍

davidrapson:
Closing this as I think it does the job for a proof-of-concept. We can implement this properly using a single component as a reference implementation. Will link to this PR in the relevant places so we have this as a reference.

PR #272, opened 2020-10-05
Run rubocop in CI

davidrapson:
As mentioned in https://github.com/citizensadvice/design-system/pull/257 we don't currently run `rubocop` on this project. We probably should.

_Edit: The docker changes in this PR also enable future gems being added like haml-link and i18n_

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,11 @@ inherit_gem:
   citizens-advice-style:
     - default.yml
 
+Style/ClassAndModuleChildren:
+  Exclude:
+    # Monkey patch for Firefox driver, should eventually be removed
+    - 'testing/features/support/core_ext/capybara/selenium/driver/firefox_driver.rb'
+
 AllCops:
   Exclude:
     - 'vendor/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,12 @@
 inherit_gem:
   citizens-advice-style:
     - default.yml
+
+inherit_mode:
+  merge:
+    - Exclude
+
+AllCops:
+  Exclude:
+    - 'vendor/**/*'
+    - 'node_modules/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,10 +2,6 @@ inherit_gem:
   citizens-advice-style:
     - default.yml
 
-inherit_mode:
-  merge:
-    - Exclude
-
 AllCops:
   Exclude:
     - 'vendor/**/*'
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,27 @@
 FROM node:14-alpine
 
+RUN apk update && apk upgrade
+
 # Install dockerize so that we can wait for hubs to be up
-RUN apk update && apk add wget
-
+RUN apk add wget
 ENV DOCKERIZE_VERSION v0.6.1
-
 RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
   && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
   && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
-# Standard Ruby Packages
-RUN apk upgrade && apk add --upgrade ruby
+# Ruby Packages
+RUN apk add --upgrade ruby ruby-dev ruby-json ruby-etc
 
+# nokogiri compilation (C-gem)
+RUN apk add gcc libc-dev libxslt-dev libxml2-dev zlib-dev build-base
+
+# git for repository-based gems
+RUN apk add --no-cache git
+
+# Install latest bundler
 RUN gem install bundler -v '2.1.4'
 
 WORKDIR /app
-
-# we only care about haml in here and the version just doesn't matter.
-RUN bundle config --global silence_root_warning 1 && bundle init && bundle add haml && bundle install
 
 COPY package*.json /tmp/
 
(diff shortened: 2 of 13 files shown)


testing/features/support/drivers/local.rb
```diff
@@ -20,13 +20,13 @@ def register
...
 
-      #Selenium::WebDriver::Safari.technology_preview!
+      # Selenium::WebDriver::Safari.technology_preview!
       Selenium::WebDriver::Service.safari({ args: ["--diagnose"] })
     end
 
     def desired_capabilities
       Selenium::WebDriver::Remote::Capabilities.new.tap do |capabilities|
-        if false
+        if false # rubocop:disable Lint/LiteralAsCondition
```
davidrapson, line 29: `if false` will never run, so should this code even exist?

testing/features/support/core_ext/capybara/selenium/driver/firefox_driver.rb
```diff
@@ -8,7 +8,7 @@
 # for more details/discussion including possible other fixes.
 #
 # LH - Sep 2020
-module Capybara::Selenium::Driver::FirefoxDriver
+module Capybara::Selenium::Driver::FirefoxDriver # rubocop:disable Style/ClassAndModuleChildren
```
davidrapson, line 11: Wasn't sure if I should just change this syntax, but disabled the warning for now.

Gemfile
```diff
@@ -6,10 +6,12 @@ gem "haml", "~> 5.2"
 
 gem "cucumber", "~> 3.2"
 gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby", tag: "v0.3.0"
+gem "rubocop", "~> 0.85.1"
```
luke-hill, line 9: This will conflict with other things as / when we upgrade. So it's best to rely on ca-style gem.

It also can be updated to v0.4 (Think I just did that though?)

testing/features/support/core_ext/capybara/selenium/driver/firefox_driver.rb
luke-hill: So I'm not a fan of in-line stuff. Instead we can use the `--auto-gen-config` stuff. Or if we're comfortable we can add it to an AllCops exclude. The reason being these aren't picked up by the CLI. or easy to track down.

As a brief intro (Although it's written above). This is a copy+paste plus bugfix from an issue in Capybara. I believe it is now fixed in master and will be released in v3.33 (But we're capped at 3.29 currently).

testing/features/support/drivers/local.rb
luke-hill: Not sure about this.... I remember some "debug" code we did where we set this, but I swore we never merged it in. Give me an hour and I'll chase this up.

.rubocop.yml
```diff
@@ -1,3 +1,12 @@
 inherit_gem:
   citizens-advice-style:
     - default.yml
+
+inherit_mode:
```
luke-hill, line 5: This in "theory" shouldn't be needed. As primarily it's used for when we have multiple sources. But I "might" be wrong with this.

luke-hill (changes requested):
Some stuff we need to change. I'll also check up on the weird `if false`

Dockerfile
```diff
@@ -1,24 +1,28 @@
...
 ENV DOCKERIZE_VERSION v0.6.1
-
 RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
   && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
   && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
 
-# Standard Ruby Packages
-RUN apk upgrade && apk add --upgrade ruby
+# Ruby Packages
+RUN apk add --upgrade ruby ruby-dev ruby-json ruby-etc
```
davidrapson, line 13: Needed to include `ruby-json ruby-etc` for `rubucop` to run

Dockerfile
```diff
@@ -1,24 +1,28 @@
...
 
+# nokogiri compilation (C-gem)
+RUN apk add gcc libc-dev libxslt-dev libxml2-dev zlib-dev build-base
+
+# git for repository-based gems
+RUN apk add --no-cache git
```
davidrapson, line 19: Included git as `citizens-advice-style` gem is installed via git. Not published to rubygems (maybe it should be?)

davidrapson (commented):
@luke-hill Mind reviewing this again. Have called out the extra bits I had to add to the Dockerfile in comments. As for whether we switch this to a ruby base image, it's probably six-of-one. Either way we have to do manual work to install whichever language is not the base.

Dockerfile
luke-hill: This is something I remember "briefly" raising. If not rubygems an internal forge. I think this is a very good point to make. I'll message you offline and find out who to speak to next for this.

scripts/haml.rb
```diff
@@ -1,12 +1,13 @@
-require 'rubygems'
-require 'bundler/setup'
-require 'haml'
+# frozen_string_literal: true
+require "rubygems"
+require "bundler/setup"
+require "haml"
 
-require('./styleguide/haml_locals.rb')
+require("./styleguide/haml_locals")
```
luke-hill, line 6: Minor thing, can we make this `require_relative "../styleguide/haml_locals"` - This assumes something being on the load path, which technically speaking it isn't. The repo is kinda set up like a gem and kinda isn't. So we need to make tackle that later on.

luke-hill (approved):
Top stuff here. Minor thing to refactor just so we're consistent

scripts/haml.rb
davidrapson: Will test that as a follow up change

PR #273, opened 2020-10-05
bugfix: Browserstack monitoring / help [NP-986]

luke-hill:
Add some more logging to specifically double check ENV vars
Remove 2 most common problematic os/browser combos (Temporarily)
Optimise docker-compose so we can always dump ENV vars if needed


pedantic-git (approved):
Yep this is cool - and it's not like we're dropping Chrome/Win Mac/Safari altogether from the tests

PR #274, opened 2020-10-05
refactor: Flaky button tests [NP-XXX]

luke-hill:
In collaboration with David R we are going to do a couple of things.

- Try spec out a new fully formed ticket for testing Button states (The combination of BDD / VRT tests both don't cover the purpose and cause issues
- To remove the BDD tests from this portion of the repo **but** retain the assertion logic. So that if we cannot fix them, we can revert back to these with some more helpers (Hover over/Hover off e.t.c.)

NB: I also added `nil-safe` checks for `docker_tag`. This works on anything after ruby 2.3 (We're using 2.6), so no issues here. Was a minor bugfix/refactor.

### DIFF
--- a/testing/features/buttons.feature
+++ b/testing/features/buttons.feature
@@ -3,23 +3,12 @@ Feature: Buttons components
   The Button components gives the current site a variety of means of
   providing means to submit data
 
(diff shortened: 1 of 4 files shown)


davidrapson (approved):
👍

PR #275, opened 2020-10-05
Add haml-lint

davidrapson:
1. To confirm that the docker image changes in https://github.com/citizensadvice/design-system/pull/272 allow us to install and run other gems in CI
2. To ensure haml linting is consistent between projects


.haml-lint.yml
```diff
@@ -0,0 +1,13 @@
...
+
+  ClassesBeforeIds:
+    EnforcedStyle: 'id'
+
+  SpaceInsideHashAttributes:
+    enabled: false
+
+  RuboCop:
+    # @TODO: Probably turn this back on
+    enabled: false
```
davidrapson, line 13: This is also disabled in public-website. Enabling triggers a bunch of small warnings, worth fixing eventually but not work blocking a build for.

Gemfile.lock
```diff
@@ -47,6 +47,14 @@ GEM
     haml (5.2.0)
       temple (>= 0.8.0)
       tilt
+    haml-lint (0.999.999)
```
luke-hill, line 50: 😕

.haml-lint.yml
luke-hill: Question) If this is being ran as a separate script. Does this need to be present here? Or will it auto-run it without it being present?

luke-hill (approved):
Seems legit

Gemfile.lock
davidrapson: Yeah kinda weird, but that's what gets added when running `bundle add haml-lint`.

.haml-lint.yml
davidrapson: It auto-runs as part of haml-lint

PR #284, opened 2020-10-06
Browserstack Partioning Fix

luke-hill:
Fix browserstack partitioning for Design System

As this used a different syntax / naming convention for their DOCKER_TAG env vars, we need a slightly different string sanitization. We also use SHA instead of build ref as that isn't passed through to Docker for some reason. Not really important as we still have the granularity required.

### DIFF
--- a/testing/features/support/drivers/browserstack/payload_values_generator.rb
+++ b/testing/features/support/drivers/browserstack/payload_values_generator.rb
@@ -48,7 +48,7 @@ module Drivers
         # Example: master_ad4b223
         # Example: PR-284_dbbb8b6
         def pr_with_sha
-          @pr_with_sha ||= docker_tag.split("design-system/").last
+          @pr_with_sha ||= docker_tag.delete_prefix("design-system/")
         end
       end
     end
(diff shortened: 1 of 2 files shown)


testing/features/support/drivers/browserstack/payload_values_generator.rb
```diff
@@ -35,23 +35,20 @@ def branch?
...
-          pr_sha_reference.split("_").first
-        end
-
-        # Example: PR-81-1_98eb207
-        def pr_sha_reference
-          @pr_sha_reference ||= docker_tag.split("design-system-").last
+        # Example: master_ad4b223
+        # Example: PR-284_dbbb8b6
+        def pr_with_sha
+          @pr_with_sha ||= docker_tag.split("design-system/").last
```
pedantic-git, line 51: I'd write this (and others) as, e.g.

```ruby
@pr_with_sha ||= docker_tag.delete_prefix('design-system/')
```

But it's mostly stylistic as you're only talking about tiny strings.

I'm happy with this change in general.

pedantic-git (approved):

simonwgill (approved):

PR #289, opened 2020-10-08
Improve callout visual regression tests

davidrapson:
- Add visual regression tests for all callout variants
- Deletes old style documentation
- Bugfix: Make standard the default type of callout. We had a "default" style documented, which was the behaviour when not passing a type. This is not an intended design, so fixing this so that we default to 'standard' if no type if provided.

Screenshot 2020-10-08 at 10 11 35


testing/features/step_definitions/callout_steps.rb
```diff
@@ -1,16 +0,0 @@
...
-end
-
-Then("a callout title and message are present") do
-  expect(@component.initial_form).to have_heading
-
-  expect(@component.initial_form).to have_message
-end
-
-Then("a/an {string} label is present above the callout title") do |label|
-  expect(@component.initial_form.warning).to eq(label)
```
davidrapson, line 15: Like with the button tests, I've covered this using visual regression tests for each type of component.

scss/6-components/_callout.scss
```diff
@@ -22,94 +22,93 @@
...
-        }
+  &,
+  &.cads-callout-standard {
+    border-color: $cads-language__brand-secondary;
+    background-color: cads-colour-to-bg($cads-language__brand-secondary);
+  }
 
-        .cads-lang-cy &::before {
-          content: 'Example';
+  &.cads-callout-example {
```
davidsauntson, line 31: the background color of this yellow, think it should be white?

but what's really weird is that the regression tests are passing with a 0 %diff:
image

davidsauntson (changes requested):
💪 apart from the small background colour comment

scss/6-components/_callout.scss
davidrapson: Ah interesting, yeah that was a mistake, let me double check that.

davidrapson: Fixed that now. I think this was because these are _new_ snapshots, as we've migrated them. But corrected this and set an explicit background on the example callout.

davidsauntson: FAB.

The backstop thing is weirding me out tho - if the reference image and the test image are different it shouldn't pass, even if the image is new.

davidsauntson (approved):

PR #290, opened 2020-10-08
Improve button visual regression tests

davidrapson:
Follow on from https://github.com/citizensadvice/design-system/pull/274

## Add accurate focus and active visual regression tests for buttons

- Adds focus state behaviour so we can accurately test `:focus`
- Replaces click behaviour with custom onRead which forces `:active` state which is what we actually want

## Support multiple eslint formats

- Base eslint file covering style settings
- Client-side js using babel-eslint
- Typescript using typescript-eslint
- Including testing engine_scripts in scope


davidsauntson (approved):
🍓

testing/features/buttons.feature
```diff
@@ -1,14 +0,0 @@
...
-
-  NB: Due to these buttons being very visual and the behaviour changing
-  dependent on each browser, we will validate them with some rigorous VRT
-  tests later on, and not test any behaviour here
-
-  Background:
-    Given the Buttons component is on the page
-
-  Scenario: Validate initial state
-    Then all the buttons are present
```
luke-hill, line 14: Can we leave this feature/scenario present, as it gives us some BDD description of the feature.

testing/visual-regression/backstop-common.js
```diff
@@ -1,8 +1,37 @@
...
+    return [
+      {
+        label: `${labelPrefix}`,
+        url: storyUrlFor(componentId),
+        viewports: [viewport],
+      },
+      {
+        label: `${labelPrefix} (hover)`,
+        url: storyUrlFor(componentId),
+        hoverSelector: 'button',
```
luke-hill, line 17: will this scope correctly to the right button? I know we're passing the selector in as well, shouldn't we use that?

luke-hill (commented):
My JS is a bit ropey, but trying to look through it seems quite clever forcing the states e.t.c. to triage the changes of effect on the buttons.

If it gives you each of the desired effects then seems fine. Can we reinstate the simple test for buttons being present as this will be expanded upon with the new simple state validation / translation.

testing/visual-regression/backstop-common.js
davidrapson: Yep, with the new docs formats the _only_ thing in the scenario is the one button. So this is intentional.

testing/features/buttons.feature
davidrapson: This doesn't _do_ anything really right now, but sure happy to leave it in.

davidrapson: Hmm scratch that, because the new docs format has one page per button variant, these tests don't make sense as is. We could either write three different "page" tests, or leave them out (which is my preference here)

testing/visual-regression/backstop-common.js
luke-hill: Rightio, that's quite cool. Makes it a bit easier.

testing/features/buttons.feature
luke-hill: Ah right, because the button components are on a different page to what was previous right?

davidrapson: Not just on a different page, but technically _three_ different pages (one for each variant). That's the way stories ideally work in storybook (we previously squished everything into a single story)

PR #291, opened 2020-10-08
Point important test at the correct variant (was example)

davidrapson:
Cleanup from https://github.com/citizensadvice/design-system/pull/289


davidsauntson (approved):
🧀 lgtm

PR #292, opened 2020-10-08
bugfix: Add error handling to cucumber tests when getting session [NP-1022]

simonwgill:
BrowserStack occasionally fail to respond to Capybara driver creation with a correct response. Since these are missing the sessionId, selenium raises an exception which kills the test.

- Added retry wrapping in set_session_id


pedantic-git (approved):
This looks familiar!

PR #293, opened 2020-10-08
feature: search design review [NP-895]

davidsauntson:
Updates the following in the design-system:
 - move search bar into new component section, add docs
 - update search button focus state
 - move breadcrumbs into new component section, add docs
 - add collapsible and non-collapsible options for breadcrumbs
 - ensures wrapped breadcrumbs are aligned
 - update vis regression tests


styleguide/components/search/search.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+  </Story>
+</Preview>
+
+## Behaviour
+
+The search component is a small form that submits a `GET` request to the provided `search_action_url`. The search term is encoded into a querystring parameter, `q`.
+
+### Inside a header
+
+When placed inside a `.cads-header` element, the text inside the search button is replaced with a magnifying glass icon.
```
davidrapson, line 29: This sounds like a job for the M in BEM. Could we convert this to a `.cads-search--compact` modifier or similar, rather than styling the search based on if it's inside another component. That would let us _show_ this variant as a story rather than tell.

testing/visual-regression/backstop-common.js
```diff
@@ -88,6 +88,37 @@ module.exports = function backstopCommon(baseUrl) {
         postInteractionWait: 500,
         viewports: [{ label: 'sticky', width: 1000, height: 600 }],
       },
+      {
+        label: 'Components/Search',
+        url: storyUrlFor('components-search--search&viewMode=story'),
```
davidrapson, line 93: You can drop `&viewMode=story` from `stroryUrlFor`

davidrapson (approved):
✍️ 👏  This is great work, particularly the documentation.

Made a suggestion about the cads-header search behaviour, and a tiny style thing on `storyUrlFor` but nothing blocking.

testing/features/support/pages/breadcrumbs.rb
```diff
@@ -2,9 +2,9 @@
 
 module DesignSystem
   class Breadcrumbs < SitePrism::Page
-    set_url "/iframe.html?id=3-components--breadcrumb&viewMode=story"
+    set_url "/iframe.html?id=components-breadcrumbs--breadcrumbs-default&viewMode=story"
 
-    section :initial_form, "#a11yComponentToTest" do
+    section :initial_form, ".cads-breadcrumbs" do
```
luke-hill, line 7: This "might" be the issue, but I need a bit more context.

Also I'm guessing this is a breaking change. Unfortunately due to our infrastructure we can't test things until after they're released. So it's likely you'll need to slap a `@future_release` tag on the feature with `@v1.11.0+`

testing/features/breadcrumbs.feature
```diff
@@ -1,14 +1,19 @@
+@future_release @v1.11.0+
```
luke-hill, line 1: With this tag here you can remove lines7,9,12,16 from this PR.

minniessuyuhuang:
A tiny comment:
On the small breakpoint, when the text is too long and gets truncated, the end of the text is too close to the right-hand side edge of the screen. 
image

PR #294, opened 2020-10-08
Gap analysis [NP-1023]

panosvoudouris:
Contains the following changes from the gap analysis:

- Visited links should have link-visited-colour from colour language, including sidebar navigation 
- Find out what happens in Contentful for bold text, update Storybook component so it’s not using <strong> tag
- Page review: Remove margin-bottom (it’s making the space between main content and footer 96px instead of 80px as design)

### DIFF
--- a/scss/6-components/_contact_details.scss
+++ b/scss/6-components/_contact_details.scss
@@ -4,4 +4,8 @@
   p {
     margin-bottom: 0;
   }
+
+  p:first-child {
+    font-weight: $cads-font-weight__bold;
+  }
 }
--- a/scss/6-components/_contact_details.scss
+++ b/scss/6-components/_contact_details.scss
@@ -4,8 +4,4 @@
   p {
     margin-bottom: 0;
   }
-
-  p:first-child {
-    font-weight: $cads-font-weight__bold;
-  }
 }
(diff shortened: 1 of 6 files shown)


scss/6-components/_contact_details.scss
```diff
@@ -4,4 +4,8 @@
   p {
     margin-bottom: 0;
   }
+
+  p:first-child {
```
davidrapson, line 8: This feels like it's asking for trouble. What happens if a the HTML doesn't start with a `<p>`. We have have no control over that right?

panosvoudouris: I agree, the issue we have currently is that the content model does not support it as the content being migrated has no concept of a title, it is just a blob. Perhaps this is something to be resolved by adding a title field (a proper one, there is one now but that is hijacked by the migration script just putting the episerver id) and the migration script fixing migrated content. But that is a whole ticket in itself. It is an issue with other things, eg titles in Callouts.

luke-hill:
@davidrapson Are you able to take on this ticket or work out what we need to do for this ticket to complete it / close it e.t.c.?

davidrapson:
@luke-hill Yep, I'm happy to pick this up

davidrapson:
Have reverted "Find out what happens in Contentful for bold text, update Storybook component so it’s not using tag" as it needs a little more exploration. Other changes look good though.

davidrapson (approved):

PR #295, opened 2020-10-08
standardise prettier

davidrapson:
Full rationale here https://prettier.io/docs/en/integrating-with-linters.html

Makes changes to prettier config to have minimum conflicts with eslint and to be as close to the default setup as we can be.

- Use eslint-config-prettier instead of eslint-plugin-prettier
- Run prettier --check during lint step
- Prettier auto-format


baseonmars (approved):
Nice simplification. 👍

PR #296, opened 2020-10-09
Update oisc_warning to accept a hash

davidrapson:
Update oisc_warning to accept a hash. See https://github.com/citizensadvice/public-website/pull/353#discussion_r502286115


KPobeeNorris (approved):
One small non-blocking comment, but other than that LGTM!

haml/_oisc_warning.html.haml
```diff
@@ -1,12 +1,12 @@
 -# Set defaults for parameters
-- body ||= '' # Body is optional
-- is_sticky ||= false
+- body = oisc_warning.fetch('body', '') # Body is optional
+- is_sticky = oisc_warning.fetch('is_sticky', false)
```
KPobeeNorris, line 3: When we're setting isSticky in public-website we'll always return true or false, so not sure if you still want the default 'false' here?

davidrapson: Yeah I think so. These components shouldn't know too much about what public-website does, so better to say what we want the default to be explicitly.

PR #297, opened 2020-10-09
feature: Docker Bump to Alpha4 images [NP-791]

luke-hill:
Same as in Public Website Cucumber

Bump images so everything is now running Selenium4
Fix up waiting logic to now use curl endpoint instead of dockerize


pedantic-git (approved):
This is great. As discussed elsewhere we might want to add a separate task to stop this running as root but it's outside the scope of this change. If it passes in CI I approve.

PR #298, opened 2020-10-09
feature: Upgrade Cucumber [NP-980]

luke-hill:
Cucumber has had 2 major version bumps in around a year. Both of which massively rewrote parts of the "common" codebase.

**Previously:**
- We were on v3, with every part custom to cucumber

**Now:**
- We are on v4, with the engine now a common engine, but kind of unconsumed in the background (Protobufs only used for Pickling and some limited message transfer)

**Later:**
- Convert json formatter to use protobuf (Set to be removed in v5, but actually still there until v6)
- Convert html formatter to use protobuf (No set date for removal .... yet)
- Decide if we need junit formatter?


testing/features/support/drivers/browserstack.rb
```diff
@@ -36,7 +36,7 @@ def valid_browserstack_drivers
 
     def browser_type_class
       if specialized_driver?
-        Module.const_get("Drivers::Browserstack::#{browser.to_s.camelize}")
+        Module.const_get("Drivers::Browserstack::#{browser.to_s.pascalize}")
```
pedantic-git, line 39: Just FYI really - you've already got activesupport in your Gemfile so this could be:

```ruby
require 'active_support/core_ext/string'

# ...

Module.const_get("Drivers::Browserstack::#{browser.to_s.camelize(:upper)}")
```

pedantic-git (approved):
Assume this one passes (I know there are problems with the other one!) I'm happy, but see comment below about activesupport for future reference.

testing/features/support/drivers/browserstack.rb
luke-hill: Yeh this was causing a few issues. Happy to look into optimising this later. The fork of protobuf-cucumber might also be losing AS soon as well.

The issue here is the AS (or protobuf-cucumber -> AS version), of this method conflicted with this one, and caused the entire show to fail.

pedantic-git: Makes sense that it conflicted, but of course we can just use the AS method instead of rolling our own.

davidrapson:
"Protobufs only used for Pickling" even though I know what both of these things are this still made me laugh 😄

Dockerfile.cucumber
```diff
@@ -13,4 +13,6 @@ RUN bundle install
 
 RUN mkdir -p /app
 
-COPY . /app
+COPY testing /app
```
simonwgill, line 16: Turns out I was wrong with these, docker-compose is actually setting the build context to the repository root in all the dockerfiles it references. Interesting, but quite confusing.

You didn't actually have to move the files or change this COPY line. It would appear to be a different issue

simonwgill (dismissed):

Dockerfile.cucumber
pedantic-git: Interesting! I didn't think that was possible in Docker but I might be misremembering!

simonwgill: Yeah, it's docker-compose doing something "helpful". Just running `docker build -f Dockerfile.cucumber` in testing/ dies because Gemfile isn't in the local context.

simonwgill (approved):
Looking good now

PR #299, opened 2020-10-09
Add error message component

davidrapson:
Add error message component. We (optimistically) only have a success message and need an error message state.

This will need design review as I've invented the styles based on our existing colour variables, which may not be perfect.

Screenshot 2020-10-09 at 17 05 21

Screenshot 2020-10-09 at 17 05 27


davidrapson:
Discussed with @soniaturcotte  and we are going to handle this use case with the error summary pattern.

PR #307, opened 2020-10-12
Remove advice feedback

davidrapson:
Removes the advice feedback component from the design system (see https://github.com/citizensadvice/public-website/pull/375). Note: We never actually used this component on the public-website end so this is completely safe to remove.

@luke-hill FYI the cucumber tests here will likely need moving over to the public website repo to test against the new version of the component, would be interested in pairing on that so I can see how the test work for that project.


davidsauntson (approved):
🔥 🔥 🔥

PR #309, opened 2020-10-12
auto pa11y

davidrapson:
Converts pa11y config to be a `.js` module so we can directly get the URLs to test for from the backstop config, removing the interstitial step of writing a JSON file to disk.


davidsauntson:
So does this mean we won't be able to look at checked in code and see the config files for pa11y that were used during the test?  Might make debugging harder in the future (I did this orig and Dan raised this issue).

davidrapson:
@davidsauntson I guess so, although you can still do this by looking at the backstop config. As a counter we often accidentally forget to commit the current list as it's only updated + committed if you remember to run the script locally.

davidsauntson (approved):
🦮

PR #311, opened 2020-10-12
v1.12.0

davidrapson:
-  NP-756 Advisernet related content component
-  Update oisc_warning component to accept a hash


davidsauntson (approved):
Are we not moving stuff into the the new components section as we touch on them?  Or are you planning to move all of this out and into AN itself?

davidrapson:
@davidsauntson To clarify I didn't work on NP-756, which is the only new component, but in that case we most likely move it out and into advisernet itself based on our last theming chat. Do you want me to do that before we do a release?

PR #319, opened 2020-10-13
fix: add specific js-* class to greedy nav, config and tests

davidsauntson:
The greedy nav default config used a naked `nav` element to initiate the greedy nav behaviour.  This PR changes this to a specific `.js-cads-greedy-nav` class.


baseonmars (approved):
excellent!

davidrapson (approved):
👍 Good catch

PR #320, opened 2020-10-14
feature: Run nightly design system regression tests [NP-292]

simonwgill:
Adds a full build to run at 2AM every night to catch any regressions that have merged in during the day.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,10 @@ configurationTypes = [
 //     ['OSX_Mojave_12', 'safari'], - These seem to be "too" popular!
 ]
 
+cron_schedule = BRANCH_NAME == "master" ? "0 2 * * *" : ""
+
 pipeline {
+    triggers { cron(cron_schedule) }
     agent {
         label 'docker && awsaccess'
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ configurationTypes = [
 //     ['OSX_Mojave_12', 'safari'], - These seem to be "too" popular!
 ]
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -12,7 +12,10 @@ configurationTypes = [
 //     ['OSX_Mojave_12', 'safari'], - These seem to be "too" popular!
 ]
 
+cron_schedule = BRANCH_NAME == "master" ? "0 2 * * *" : ""
```
luke-hill, line 15: how about ...

`cron_schedule = deployBranches.contains(BRANCH_NAME) ? "H 2 * * *" : ""`

That way we cover ourselves in a bit of load balancing (Any time at 2:xx am), and we also keep up with Line63/66), which means if we build this elsewhere we can monitor each build environment).

simonwgill: Makes sense to match the condition in the Full Regression Test. I can't remember where deployBranches comes from at the moment - can you remind me what it is going to contain?

Don't think the H change is going to do much other than start the job at a random time between 2 and 3. It'll most likely still need an agent provisioning at that time of the morning whenever it starts.

luke-hill: That's set on line 1. Currently it is just `["master"]` but it could change down the line.

Yeh I just meant we don't care that its exactly 2am, so making it more flexible might help 🤷

simonwgill: Ah, blinded by the big array of config types.

If we use H, it'll still be running at the same time every night, just a time that we haven't chosen. Round numbers are nicer in the logs... and if we don't have resource contention, I'd prefer just to leave it at "0 2".

luke-hill (approved):
🎉

PR #316, opened 2020-10-13
feature: TEMPORARY - New browserstack endpoint [NP-1048]

luke-hill:
Use new more stable browserstack endpoint temporarily whilst they fix other issues

cc/ @david-lake / @adriana-rus

PR #321, opened 2020-10-14
Revert "feature: TEMPORARY - New browserstack endpoint [NP-1048]"

luke-hill:
Reverts citizensadvice/design-system#316


luke-hill:
FYI @pedantic-git / @mrdaniellewis This seems to have made firefox crash at a rate of around 40% Which is just plain ridiculous. The errors vary from marionette to connectivity. It's like the firefox browser is a time-bomb.....

Reverting and just carrying on with the existing flakiness being retried for now. Experimented failed!

simonwgill (approved):

PR #322, opened 2020-10-14
Account for optional form hints, document form components

davidrapson:
Companion to https://github.com/citizensadvice/public-website/pull/375

Currently the form components include empty elements for hints and errors, even if it not required for the current form field. When using these classes outside of the haml components (i.e. as a React component) it doesn't make sense to include an empty element just to get the layout working.

This PR:

- Moves form fields to the new documentation format, under a dedicated "Forms" heading
- Includes visual regression tests for multiple variants of each form field
- Accounts for optional hints

**Before**
image

**After**
image

image

**Note:** Noticed whilst doing this that `radio_group` and `radio_group_small` are the same apart from a single class name. These should probably be the same partial with an option, but that's beyond the scope of these changes.


scss/4-elements/_forms.scss
```diff
@@ -211,8 +211,11 @@ textarea,
   }
 }
 
-.cads-form-optional {
-  .cads-form-label::after {
+.cads-form-label {
```
davidrapson, line 214: This is the 🐛 fix. Makes labels block, and accounts for reduced margin when a hint is present 👇

haml/_input.html.haml
```diff
@@ -3,6 +3,8 @@
   .cads-form-content.cads-input-group
     %label.cads-form-label.cads-heading-small{id: "#{input['name']}-label", for: "#{input['name']}-input" }
       = input['label']
-    %p.cads-form__hint= input['hint']
-    %p.cads-form__error-message= input['error_message']
+    - if input['hint']
```
davidrapson, line 6: The lack of if statements here was masking the bug. If you use these classes standalone and don't need a hint then the label would display inline.

testing/visual-regression/backstop-common.js
```diff
@@ -32,6 +32,32 @@ module.exports = function backstopCommon(baseUrl) {
...
+      },
+      {
+        label: `${labelPrefix} (optional field)`,
+        url: storyUrlFor(`${componentId}--optional-field`),
+        viewports: [viewport],
+      },
+      {
+        label: `${labelPrefix} (with error)`,
+        url: storyUrlFor(`${componentId}--with-error`),
+        viewports: [viewport],
```
davidsauntson, line 56: testing focus states would be good

davidsauntson (approved):
🔘 great 

Nothing blocking but it would be good to have headings for each variant as per the other docs (we should probably agree a template for docs or smth with @soniaturcotte too) & focus state tests in the VRTs.

testing/visual-regression/backstop-common.js
davidrapson: Agreed, might leave that to a follow on PR as the main goal of this one is to fix the display bug

PR #328, opened 2020-10-15
Story layout

davidrapson:
Bit of churn on this one sorry! This switches storybook to use the built in `layout` parameter, which is documented, rather than requiring a custom wrapper and overriding styles.

The result is that a chunk of the visual regression tests need updating as they include slightly different padding in mobile views. Doesn't affect the actual component styles, but makes the storybook layout more consistent.


styleguide/component.stories.js
```diff
@@ -75,12 +75,14 @@ export default {
     },
   },
   decorators: [
-    (storyFn) => `<div
-      class="cads-styleguide__language-label"
-      onclick="javascript:styleguideChangeLangugage();"
-      aria-hidden="true"
-    ></div>
-    ${storyFn()}`,
+    (storyFn) => `<div class="cads-styleguide__wrapper">
```
davidrapson, line 78: Kept `cads-styleguide__wrapper` in the old component stories as it's used as a hook for some of the backstop tests. No longer any styling hooked to it though.

davidsauntson (approved):

PR #329, opened 2020-10-15
Constrain error summary with max-width instead of %

davidrapson:
The error summary currently has a % based width. This causes issues when the error summary is used with in a context that is already constrained (the advice feedback form is what motivated this change).

The relates to the recommendation in the gap analysis so have removed the % based width. It we need to constrain it further that can be handled by the context it's used in (i.e. already sitting within a constrained grid column).

~Note: @luke-hill asked me to hold off on migrating this component over to the new docs as they are wrapping up some tests for this component, so to avoid conflicting with this work we can migrate this at a later date.~

☝️ Luke has added these test so I've migrated this over to the new style docs as part of this PR. In a new "Forms" grouping to pair with https://github.com/citizensadvice/design-system/pull/322


testing/visual-regression/backstop-common.js
```diff
@@ -116,6 +116,12 @@ module.exports = function backstopCommon(baseUrl) {
         selectors: ['.cads-breadcrumbs'],
       },
 
+      // Forms
+      {
+        label: 'Forms/Error summary',
+        url: storyUrlFor('forms-error-summary--example'),
```
davidsauntson, line 122: focus state for summary div and links?

davidsauntson (approved):
🍏 fab

not a blocker but focus state for error summary div and links would also be good

testing/visual-regression/backstop-common.js
davidrapson: Good shout, will combine that with focus state tests for the other form elements as a follow on PR

PR #330, opened 2020-10-15
Feature: Error Summary Spec [NP-719]

luke-hill:
Error Summary Specification (Based on Advice Feedback).

NB: This uses the old system which will be migrated to the new state-driven components at a later date


testing/features/support/pages/error_summary.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module DesignSystem
+  class ErrorSummary < SitePrism::Page
+    set_url "/iframe.html?id=3-components--error-summary&viewMode=story"
+
+    section :initial_form, "#a11yComponentToTest" do
+      element :heading, "#error-summary-title"
+      element :error_quantity_label, ".cads-error-summary__field_count"
```
davidrapson, line 9: Is there any way in these kinds of tests to query based on the display text? Using CSS classes isn't ideal

davidrapson: We use https://testing-library.com/ in the client-side tests for this reason, is there an equivalent in cucumber land?

Gemfile
```diff
@@ -8,7 +8,7 @@ gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby"
 gem "cucumber", "~> 4.0"
 gem "dotenv"
 gem "faraday", "~> 1.0"
-gem "haml-lint"
+gem "haml_lint"
```
davidrapson, line 11: 🙈  (thanks!)

davidrapson (approved):
👍 LGTM aside from a comment about selectors on the tests, but not worth blocking on

testing/features/support/pages/error_summary.rb
luke-hill: I'd prefer not to query based on text. Given we often assert based on text. Although having said that, this is a sample testing library, so perhaps we could?

If there are ID's to be hooked into they're perfect. I cannot remember if there was. I'll do a quick pass to check

PR #332, opened 2020-10-15
feat: design-system translations [NP-974]

baseonmars:
This is a draft, don't go too heavy on the code review right now (if at all) but do feel free to check it out and try it out.

This PR introduces a language switcher to the design system, and more importantly rails translations to the hall templates.

What's not here:

- [x] documentation
- [x] ~handling code transformations in the right place (should we use a json loader after the haml one?~ No more JSON loader required 
- [x] better way of importing haml templates - would be nice to have a default for templates that contain no translations? Success - English is now returned as the default.
- [x] better haml.rb code structure. it was fine when it was a scrappy little script but it'd be nice to make it act like an adult. little more structure, some help text, error messages. that sort of thing.


davidrapson:
This looks great! Glad the approach we sketched out actually works! Couple of things I noticed whilst playing around:

- The `default` property comes back as a buffer, rather than a string. I think we want this to be raw text so we can use it to get the source code view working.
- Maybe we could write our own little wrapper around `Story`. Might take a little configuration, but I _think_ the doc blocks are little React components so we could have something like:

```tsx
import { Story } from '@storybook/addon-docs/blocks';
import { translate } from '../../../scripts/story-helpers';

const HamlStory = ({ haml, ...storyProps }) => {
  return (
    <Story {...storyProps}>{(_, options) => translate(haml, options)}</Story>
  );
};

export default HamlStory;
```

This looks great so far though!

baseonmars:
> * Maybe we could write our own little wrapper around `Story`. Might take a little configuration, but I _think_ the doc blocks are little React components so we could have something like:

So... I added a version as both a decorator and a component. I think either works... the component hides a lot of complexity automating the docs side of things, I'm not 100% sure I like it. Using the decorator leads to more explicit code which might work in the projects favour in the future.

baseonmars:
Quite a few changes to the build to get this to pass:

- docker ignore vendor/bundle (this caused nokogiri and ffi gem failures).
- properly configure nokogiri installation.
- install ruby-bigdecimal which is needed for activesupport to run on alpine.
- build and then serve the styleguide rather than use webpack hot loader
- serve docs from 6006 so test urls match up.

baseonmars:
> * Maybe we could write our own little wrapper around `Story`. Might take a little configuration, but I _think_ the doc blocks are little React components so we could have something like:

Replying here for future reference. We looked into creating a nicer API for writing stories that contain HAML but there's a fair amount of complexity in writing a DocBlock component that goes beyond creating a simple wrapper. We've decided to park it until there's a clearer understanding of what's required.

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     build: .
     image: 'ca-styleguide${CA_STYLEGUIDE_VERSION_TAG}'
     container_name: ca-styleguide.test
-    command: npm run styleguide:ci -- --quiet
+    command: npm run docs:serve
```
davidrapson, line 8: 💪

scripts/translated-haml-loader.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Webpack loader, takes a haml file as input and outputs
+ * an object keyed by translation.
+ *
+ * Note: Changing this file will trigger hot reloading but changes will
+ * now be included, you must restart webpack.
```
davidrapson, line 6: Should this be "but changes will **not** be included"?

styleguide/translations/translated.stories.mdx
```diff
@@ -0,0 +1,46 @@
+import { Meta, Story, Preview, Source } from '@storybook/addon-docs/blocks';
+import { translate } from '../../scripts/story-helpers';
+
+<Meta title="Translations/How To" />
+
+# Translations
```
davidrapson, line 6: 😍 This doc is a work of art.

styleguide/translations/translated.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+Translations are handled using the [Rails i18n API](https://guides.rubyonrails.org/i18n.html). To view translated versions of components, select the globe icon in the toolbar and choose between English and Cymraeg.
+
+Rather than including English or Welsh strings in a component's HAML code they are placed in one of the `yml` files located in the `locales` directrory.
+
+For Example `en.yml` may contain:
+
+```yaml
+en:
```
davidrapson, line 15: This should probably be:

```suggestion
en:
  cads:
    example:
      name: Citizens Advice
      body: Loren ipsum dolar sit
    search:
      button: Search
```

To match the actual format of the yml files.

davidrapson (approved):
👍 This is wonderful stuff. I'm going to try out the instructions and `npm link` this to public website to confirm I can get translations working there. Will report back. I especially love the how-to document 💪

davidrapson:
>  I'm going to try out the instructions and npm link this to public website to confirm I can get translations working there. Will report back.

For our future selves. I tested this using `npm link` and adding `= t('cads.example.body')` to a template.

Needed to move the i18n config out of `config/application.rb` and into `config/initializers/locales.rb` with the following config:

```rb
# frozen_string_literal: true

I18n.available_locales = [:en, :cy]

# Add design-system locales to the load path
I18n.load_path += Dir[
  Rails.root.join("node_modules/@citizensadvice/design-system/locales/*.yml"),
]
```

That allowed our design-system translations to a) load correctly and b) be overridable by the application if needed.

Screenshot 2020-10-22 at 12 39 28

PR #340, opened 2020-10-16
Component unit tests and lint fixes

davidrapson:
Adds testing-library unit tests for:

- callout heading swap
- header search toggle
- targeted content

Also fixes lint warnings and general pattern of using `try/catch` and `console.logs` in favour of using guard statements. Removed lots of `[0]` code by using `querySelector` which returns the first match by design.

### DIFF
--- a/src/js/callout.js
+++ b/src/js/callout.js
@@ -9,10 +9,12 @@ const initCallouts = () => {
    */
   if (firstCallout && h2s.length === 0) {
     const h3ToReplace = firstCallout.querySelector('h3');
-    const heading = h3ToReplace.innerHTML;
-    const newHeading = document.createElement('h2');
-    newHeading.innerHTML = heading;
-    h3ToReplace.parentNode.replaceChild(newHeading, h3ToReplace);
+    if (h3ToReplace) {
+      const heading = h3ToReplace.innerHTML;
+      const newHeading = document.createElement('h2');
+      newHeading.innerHTML = heading;
+      h3ToReplace.parentNode.replaceChild(newHeading, h3ToReplace);
+    }
   }
 };
 
--- a/src/js/callout.js
+++ b/src/js/callout.js
@@ -1,20 +1,18 @@
 const initCallouts = () => {
-  try {
-    const h2s = document.getElementsByTagName('h2');
-    if (!h2s || !h2s.length) {
-      // Only do this if we don't have an H2 in the page
-      const callouts = document.getElementsByClassName('cads-callout');
-      if (callouts && callouts.length > 0) {
-        const callout = callouts[0];
-        const h3ToReplace = callout.getElementsByTagName('h3')[0];
-        const heading = h3ToReplace.innerHTML;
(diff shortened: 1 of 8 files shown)


src/js/callout.test.js
```diff
@@ -0,0 +1,21 @@
...
+import { screen } from '@testing-library/dom';
+
+import initCallouts from './callout';
+
+const mockCallout = `<div class="cads-callout cads-callout-standard">
+  <h3>Callout title</h3>
+  <p>Example callout text</p>
+</div>`;
+
+test('should convert h3 in callout to a h2 if there is no existing h2', () => {
```
davidrapson, line 11: I don't fully understand why this behaviour is in JS vs. passing in a `heading_level` to the component but good to have some tests for it.

src/js/targeted-content.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+test('allow toggling targeted content', () => {
+  document.body.innerHTML = componentHtml;
+  initTargetedContent();
+
+  const detailsEl = screen.getByTestId('targeted-content');
+  const summaryEl = screen.getByText('Summary title').parentElement;
+
+  function expectOpen() {
+    expect(detailsEl).toHaveClass('is-open');
```
davidrapson, line 31: I was hoping to fix the fact targeted content doesn't work without JS by using the inbuilt `open` attribute for details/summary but storybook docs don't work without JS! So need to think more about how we validate that.

src/js/header.js
```diff
@@ -1,27 +1,29 @@
...
-      'cads-search-reveal'
-    )[0];
+  const CLASS_NAMES = {
+    showSearch: 'cads-header-show-search',
+    iconSearch: 'cads-icon_search',
+    iconClose: 'cads-icon_close',
+  };
 
+  const header = document.querySelector('.cads-header');
+  const controlButton = header.querySelector('.cads-search-reveal');
```
baseonmars, line 9: Do we want to capture when these are null and log out an error?

src/js/callout.js
```diff
@@ -1,20 +1,18 @@
...
+  const h2s = document.querySelectorAll('h2');
+  const firstCallout = document.querySelector('.cads-callout');
+
+  /**
+   * Only run if we:
+   * - Have a callout on the page
+   * - Don't already have a h2
+   */
+  if (firstCallout && h2s.length === 0) {
+    const h3ToReplace = firstCallout.querySelector('h3');
```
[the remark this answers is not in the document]
davidrapson: Cool, I'll put a guard in here, rather than try/catch but will reinstate a warning if nothing is found

src/js/callout.test.js
davidrapson: https://citizensadvice.atlassian.net/browse/NP-1096

davidrapson:
@baseonmars Mind giving this another look with the latest changes?

baseonmars (approved):
LGTM :+1:

PR #341, opened 2020-10-16
feature: OISC Spec [NP-472]

luke-hill:
This adds the OISC Spec to the tests. But also in the new format which we will adopt for all newly migrated "stated" components.

**New bits**:
- The top level namespace will be the component name
- The bottom level will be the state
- We will take advantage of the new `Rule` keyword to split by state inside the feature.
- Inheritance from other components to simplify things like load validations

**Bits retained**:
- One feature per component
- One set of steps per component

**Bits to be wary of**:
- We now have 2 ways of doing the tests (Whilst we migrate over other things)
- I've coupled this PR with some tidy ups of requires as `env.rb` was getting unruly. Now we have an `all.rb` file in the relevant places.
- The test for "stickiness" is a bit unique, but it makes usage of Seleniums WebElement locator strategy.
  - The reason for this is that the W3C definition of visible / present, does not take into account viewport, just whether it "could" be there. So instead we assert based on its **current** position.


testing/features/oisc_warnings.feature
```diff
@@ -0,0 +1,46 @@
...
+  to warn users about something important; in a variety of ways
+
+  Rule: OISC With Title and Description
+    Background:
+      Given a Title and Description OISC component is on the page
+
+    Scenario: Validate initial state
+      Then the OISC title is visible
+      And the OISC message is visible
+      And the OISC message contains a link
```
davidrapson, line 13: I'd maybe not make the link part of the spec. As in usage links are totally optional. Not a big deal as this is a static state, but just reflecting that the link isn't _required_.

davidrapson (approved):
👍 Nice! Really glad the new story format allows to write more valuable tests like this. The sticky behaviour is exactly the kind of gap that these tests can help with, and we can't cover with visual regression tests alone.

testing/features/oisc_warnings.feature
luke-hill: Ah oki. No probs, I'll remove this. I wasn't 100% sure. So that main bit is just plain content then?

davidrapson: Yep, limited formatting available in contentful, but consider that a bit of a black box of content.

KPobeeNorris (approved):
LGTM

PR #349, opened 2020-10-19
Experimental: Translate search label

davidrapson:
Quick experiment using https://github.com/citizensadvice/design-system/pull/332 to move the search label translation out of CSS and into a locale file 👍


davidrapson:
Pretttty good

Screenshot 2020-10-19 at 15 40 34
Screenshot 2020-10-19 at 15 40 31

PR #350, opened 2020-10-19
Migrate sample pages to new docs format

davidrapson:
Groups stories by page type in a new "Sample pages" section

Screenshot 2020-10-19 at 17 18 15


styleguide/samples/_advice_collection_advisernet.html.haml
```diff
@@ -1,39 +0,0 @@
-.cads-advisernet
```
davidrapson, line 1: Removed this sample on the rationale that as part of the upcoming theming work we'll shift advisernet specific theming into the website repo

davidsauntson (approved):
🍏

PR #357, opened 2020-10-20
bugfix: Add example forms page [NP-973]

soniaturcotte:
### Add example forms page

This is so design can play around with spacing and review UI changes in a realistic environment. 

- Added a new page
- Included some form components

The code in include the buttons component is different to that from the form ones, so I'm not sure if this is correct. It _looks_ right on the page.


styleguide/samples/_forms.html.haml
```diff
@@ -0,0 +1,53 @@
...
+              %a{:href=> 'https://www.citizensadvice.org.uk'} A to Z of advice
+            %li
+              %a{:href=> 'https://www.citizensadvice.org.uk'} Finding information
+            %li
+              %a{:href=> 'https://www.citizensadvice.org.uk'} Search for your local Citizens Advice
+            %li
+              %a{:href=> 'https://www.citizensadvice.org.uk'} Browse our campaign pages
+            %li
+              %a{:href=> 'https://www.citizensadvice.org.uk'} Browse our research pages
+= render partial: "@citizensadvice/design-system/haml/footer"
```
davidrapson, line 53: Sadly the build failed for a pedantic reason. It wants you to put an empty new line at the end of this file

davidrapson (approved):
👍 Looks great. FYI I've raised https://github.com/citizensadvice/design-system/pull/350 which moves the sample pages over to the new docs format, but I can sort that out afterwards.

You'll need to fix a tiny pedantic warning from a 🤖  (left a comment) but otherwise looks good.

styleguide/samples/_forms.html.haml
```diff
@@ -0,0 +1,53 @@
...
+%main.cads-main-content
+  .cads-grid-container
+    .cads-grid-row
+      .cads-grid-col-md-8
+        = render partial: "@citizensadvice/design-system/haml/notice_banner"
+    .cads-grid-row
+      .cads-grid-col-md-8
+        %h1 Consumer Query
+        %p From 1 April 2019 we are only able to provide general consumer advice to residents of England and Wales. If you are resident in Scotland and have a general consumer enquiry please contact Advice Direct Scotland / 0808 164 6000 
+
```
davidrapson, line 16: Sadly the build failed for a pedantic reason. You have some trailing whitespace at the end of these two lines that the 🤖  wants you to get rid of.

soniaturcotte: The absolute embarassment. Think it's fixed now.

davidrapson: Ugh now it's moaning about formatting in page.stories.js. Without getting you to set up a bunch of tools if you run `npx prettier --write styleguide/page.stories.js` in your terminal it should fix that. We might need to think about a way to have the CI tool tell you _all_ the code-style warnings in one go, rather than failing on the first thing.

davidrapson:
It passed! 🚢 it!

PR #358, opened 2020-10-20
refactor: Testing code tidy / refactor [NP-1067]

luke-hill:
- Used base page inheritance so all pages can have a common load validation.
  - Base Load Validation is interrogated through `#validate_initial_state!`
- Altered OISC components to correctly inherit from most generic to most specific
- Moved all driver logic into `support/driver.rb` to clean up `support/env.rb`
- Removed files/logging that is no longer needed / helpful for codebase


testing/features/support/components/all.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+require_relative "base"
+require_relative "oisc_warning/title_only"
```
davidrapson, line 4: Not quite sure why "oisc_warning/title_only" is special cased here?

davidrapson (approved):
👍 LGTM with a small query

testing/features/support/components/all.rb
luke-hill: Ah right, so yeh urm.... So when you run cucumber, the auto-loader loads all files for you. Like all good auto-loaders strictly speaking you shouldn't trust it 100% (If it matters).

In ruby because things can be scripted / ran almost verbatim, the following code would throw an error.

```
class Two < One
end

class One
end
```

You'd get something like this:

```
2.6.5 :001 > class Two < One
2.6.5 :002?>   end
Traceback (most recent call last):
        4: from /usr/share/rvm/rubies/ruby-2.6.5/bin/irb:23:in `<main>'
        3: from /usr/share/rvm/rubies/ruby-2.6.5/bin/irb:23:in `load'
        2: from /usr/share/rvm/rubies/ruby-2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):1
NameError (uninitialized constant One)
```

So to mitigate this, we always load the parent classes first. It gets even weirder if you have sections/pages (From SitePrism), but we don't here. So we just need to ensure we load parents first.

TL;DR - Ruby 💎

luke-hill: EDIT: We could place every single file in here to be required. But all we care about is `oisc_warning/title_only` is loaded **before** `oisc_warning/title_description` and `oisc_warning/sticky`

PR #359, opened 2020-10-20
bugfix: Cucumber Hooks [NP-1099]

luke-hill:
- Fix cucumber hooks by using the feature file name as opposed to feature name (No longer in pickle).
  - This has been broken since the cucumber4 upgrade

NB: This is copied from the `public-website-cucumber` fix done a week or so ago.


davidrapson (approved):
```rb
module This
  module Looks
    module Good
       module To
         module Me  
            puts "👍"
         end
      end
    end
  end
end
```

luke-hill:
abc

PR #365, opened 2020-10-21
bugfix: Remove incorrect acutes [NP-973]

soniaturcotte:
The code example in the new components includes 2 acutes ` 

When copying the component from storybook, they are included in the clipboard.
We should consider removing that, so bugs like this don't go by unnoticed.

This pull request just removes the acute from the forms example page.


davidrapson (approved):
👍

PR #366, opened 2020-10-21
feature: Port over Callout Tests [NP-1060]

luke-hill:
Like the PR before for breadcrumbs. This ports over the old code / logic for Callout tests into the new Rule / state format.

I've done 2 refactors in this PR also

1) Tidied up a load of the naming of things, to try make things a little less vanilla, and a bit more informative (What the format of cucumber is meant to be).

2) Moved the clever script for obtaining `::before` content into a `Helpers::Scripts` module, which can take an input argument of the locator string. This means we don't need to copy 4 lines of verbose script. It also allows us to add extra scripts at a later date if needs be.


davidrapson (approved):
👍 LGTM

davidrapson:
Although build is failing on the lint step

luke-hill:
Rubocop is clever and even looks inside heredocs. Damn it!

PR #369, opened 2020-10-22
feature: Enable iOS Driver (Not in CI!) - [NP-748]

luke-hill:
This PR allows us to use an iOS driver locally and ad-hoc. We'll look to integrate this driver into the CI flow in the coming weeks/month

Primarily this facilitated refactoring the temporary code which was put in place when we introduced all browserstack drivers (Some of this was wrong).

I've also massively updated our support doc, including removing Catalina (macOS 13), as this had too many caveats and workarounds to be ran locally / remotely. Big Sur (macOS 11), will be enabled and should work fine.

The devices and iOS versions enabled are a bit prescriptive. But I figured instead of trying to work against Browserstack I'd just work with what worked.


BROWSER_SUPPORT.md
```diff
@@ -13,23 +13,20 @@ combinations are supported.
 - Windows 7
 - Windows 8.1
 - Windows 10
-- OSX Catalina (13) - \_NB: This doesn't like working on certain pages, there are several issues with `safaridriver`
-- OSX Mojave (12)
+- OSX Mojave (12) - We **don't** support Catalina (13), as it is too flaky (Big Sur will be supported).
```
davidrapson, line 16: This is more about framing, but we could we frame this document so that it's clear that we are talking about _automated testing_ support? I don't want to accidentally imply that the components wouldn't work fine in these OS combinations. This is about what we test against in BrowserStack right?

davidrapson (approved):
👍 To getting some mobile device testing in here (I'd still like to see fewer Firefoxes and more general spread, but I know you are reviewing that generally).

Minor comment about the framing of the browser support document. But nothing blocking.

BROWSER_SUPPORT.md
davidrapson: Actually looking at the full doc I can see you've kinda covered that in the preamble. So 👍

PR #370, opened 2020-10-22
Prepare stylesheets for theming [NP-1062]

davidrapson:
**Prepare stylesheets for theming**

- Add `!default` to all colour variables
- Re-merge stylesheets into a single entrypoint

**Simplify sass build step**

As part of these changes I took the opportunity to simplify the sass build by replacing the custom build script with equivalent cli commands. I couldn't see anything special in build.js that the clis don't already support built in.

**Switch to sass (instead of node-sass)**

According to the docs:

> Dart Sass is the primary implementation of Sass

node-sass is a wrapper around libsass which requires native extensions to be built where as dart sass compiles directly to javascript so should make installs quicker and is inline with the current recommendation.

Happy to pull this part out into another PR?


scss/6-components/components-imports.scss
```diff
@@ -15,3 +15,4 @@
 @import './error-summary';
 @import './pagination';
 @import './related-content';
+@import './advisernet/related-content';
```
davidrapson, line 18: Planning to pull this out into the advisernet code as a follow on PR

davidsauntson (approved):
F.A.B ⚡ 🐦

PR #371, opened 2020-10-22
feature: Port over Buttons tests [NP-1059]

luke-hill:
Same as the other port over tests. We've migrated the existing tests over from the old legacy code to the new "state-d" format.

Only thing worth noting here is because these are primitive tests, I've tried to "zazz" them up a little, without straying over into the VRT territory


davidrapson (approved):
👍 Looks good to me. Appropriate level of "zazz".

PR #376, opened 2020-10-23
feature: Automation: Language Testing Skeleton [NP-974]

luke-hill:
**Previously:**

A variety of strings were tested in one variant (English), and hard-coded to be tested

**Now:**

A variety of strings are (still), only tested in one variant. But we have
- ~The steps ready to switch to Welsh when we are able to (Working with Dan on how to do this)~ The steps to switch language will now work using a query-string once the relevant ticket is merged
- The yaml files have gotten slightly updated to provide some of the strings ~(Not translated yet)~ (Will sit ontop of the translations work being done by Dan)
- We interrogate these yaml files correctly during our tests (these are the same yaml files served by the DS proper)

~NB: This is a Skeleton because we don't _actually_ change language yet - That step does nothing. But the components we have translated are now covered and "ready". It's likely I may have made some mistakes or mis-assumptions here, so feel free to correct.~


testing/features/support/components/base.rb
```diff
@@ -6,6 +6,10 @@ class Base < SitePrism::Page
     [validate_initial_state!, "Initial #{self.class} component didn't load correctly!"]
   end
 
+  def switch_language(_language)
+    "TBD"
```
luke-hill, line 10: This is the only non-functional piece of this PR.

testing/features/oisc_warnings.feature
```diff
@@ -7,23 +7,38 @@ Feature: OISC Warning components
     Background:
       Given a Title and Description OISC component is on the page
 
-    Scenario: OISC component has a title and message
+    Scenario: English OISC component has a title and message
```
baseonmars, line 10: The part of OISC that's translated is the "Close this warning" label, the rest of the content is user supplied, so I don't believe this is the right set of tests for OISC.

testing/features/callout.feature
```diff
@@ -7,30 +7,50 @@ Feature: Callout component
     Background:
       Given a Standard Callout component is on the page
 
-    Scenario: Callout contains a title and message
+    Scenario: English Callout contains a title and message
```
baseonmars, line 10: The translated parts of the Callout are the labels (currently in CSS and due for extraction), the rest of the content is user provided, so similarly to the OISC warnings you might find that some of these tests aren't needed.

locales/cy.yml
```diff
@@ -3,3 +3,11 @@ cy:
...
       name: 'Cyngor ar Bopeth'
       body: 'Dyma ychydig o destun y corff'
+  components:
+    callout:
+      warning:
+        adviser: Adviser
+        example: Example
+        important: Important
+    oisc_warning:
+      title: This is an OISC level 2 topic
```
baseonmars, line 13: These are currently provided by the CMS so won't be included in this translation file. Need to remove to avoid confusion.

testing/features/support/components/base.rb
baseonmars: Still figuring out an straight forward solution on this. Will get back to you on it soon.

baseonmars (changes requested):
This is a good start, I've left some comments that need to be addressed. Hopefully I can get the translations in today and that will make it easier to write this.

Do feel free to ask any questions, and as always happy to collaborate if it saves either of us re-doing work.

testing/features/callout.feature
luke-hill: This test just validates presence of title / message. The label is the one interrogated. So these should be fine.

testing/features/oisc_warnings.feature
luke-hill: I couldn't find that label anywhere on the current version. So I'm guessing that's a new piece of functionality?

locales/cy.yml
```diff
@@ -3,3 +3,9 @@ cy:
     example:
       name: 'Cyngor ar Bopeth'
       body: 'Dyma ychydig o destun y corff'
+  components:
+    callout:
```
baseonmars, line 7: The structure on this is gonna be:

```    
callouts:
    example: Example
    important: Important
    adviser: Adviser```

baseonmars (changes requested):

baseonmars (approved):

PR #377, opened 2020-10-23
Remove the need for cads-styleguide__language-label

davidrapson:
A bit of clean up following https://github.com/citizensadvice/design-system/pull/332

Remove the need for `cads-styleguide__language-label` in favour of using the new local global. Uses this to add the appropriate cads-language-* class for components that still rely on this (.i.e. the logo).


.storybook/preview.js
```diff
@@ -25,3 +25,10 @@ export const globalTypes = {
     },
   },
 };
+
+const withLocaleClassname = (Story, { globals }) => {
+  console.log(globals.locale);
```
baseonmars, line 30: console.log

baseonmars (changes requested):

baseonmars (approved):

.storybook/preview.js
davidrapson: 🙈 Try again now!

PR #378, opened 2020-10-23
Resolve targeted content crossed streams 

davidrapson:
Pairs with https://github.com/citizensadvice/public-website/pull/440

There's a bit of confusion with the targeted content code that crosses the streams heavily between public website and design system. This PR is an attempt to improve things.

**Migrates styles for expand all button**

These styles were included in design system despite the actual functionality living in public website

**Moves anchor link expansion code to design system**

This behaviour was included in public website but really belongs to the core targeted content behaviour.


styleguide/components/targeted-content/targeted-content.stories.mdx
```diff
@@ -0,0 +1,65 @@
...
+## JavaScript behaviour
+
+Targeted content requires some additional JavaScript behaviour which can be initialised with:
+
+```js
+import initTargetedContent from '@citizensadvice/design-system/lib/targeted-content';
+initTargetedContent();
+```
+
+**Note**: If you are supporting IE 11 or earlier make sure you include an [Element.prototype.closest polyfill](https://www.npmjs.com/package/element-closest).
```
davidrapson, line 65: The equivalent code from public website had a lot of very specific traversal to do this. Using .closest is the most stable way of finding a parent by selector.

scss/6-components/_targeted-content.scss
```diff
@@ -79,18 +79,3 @@
     }
   }
 }
-
-.cads-targeted-content__expand-all {
```
davidrapson, line 83: This was included here but the actual code for this was in public website. Still not sure exactly which side this should live on, but we need to at least pick _a_ side!

davidsauntson (approved):

PR #379, opened 2020-10-23
feature: Header Spec [NP-734]

luke-hill:
Adds all code required to perform header testing for new format (Currently not yet complete)
Adds in simple test/s for header against existing component


davidrapson (commented):
Not sure these tests make sense at the design-system level. The language switcher and log in / log out links are handled by the consumer (i.e. public website). We also wouldn't expect you to be able to search from the design system pages. The header in this context is simply static markup. It might make more sense for these behavioural tests to live in public-website-cucumber as that's where things are _actually_ hooked up.

We might still want some here but a basic markup smoke test would be enough.

luke-hill:
I think these tests here just validate that there are some elements. We have much more in-depth functionality on the regular public-website-cucumber repo for logging in / out using usernames e.t.c.

If/When we migrate these to the new state-d functionality, if we don't have multiple states it's fine by me.

davidrapson (approved):
👍 Gotcha! In which case, crack on!

PR #382, opened 2020-10-23
 feature: DS - Automated Tests - Footer [NP-733]

celonicolosi:
Add new feature to test the footer component. Because the footer component has a link that users can leave feedback on the pages that they are on, it also adds a new helper to handle newly opened tabs.

### DIFF
--- a/testing/features/support/components/footer.rb
+++ b/testing/features/support/components/footer.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module DesignSystem
+  class Footer < ::Base
+    set_url "/iframe.html?id=3-components--footer&viewMode=story"
+
+    sections :category_titles, ".cads-grid-col-md-3 > h2" do
+      elements :links, :xpath, "./following-sibling::ul[1]/li/a"
+    end
+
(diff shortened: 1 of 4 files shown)


testing/features/support/components/footer.rb
```diff
@@ -0,0 +1,18 @@
...
+    set_url "/iframe.html?id=3-components--footer&viewMode=story"
+
+    sections :category_titles, ".cads-grid-col-md-3 > h2" do
+      elements :links, :xpath, "./following-sibling::ul[1]/li/a"
+    end
+
+    section :initial_form, ".cads-footer" do
+      element :website_feedback, ".cads-website-feedback a"
+      element :logo, ".cads-logo"
+      element :copyright, ".cads-footer__company-info div p:nth-of-type(1)"
```
davidrapson, line 14: Don't need to worry about this now, but we'll be doing a bit of design review work on the footer so can hopefully improve some of these selectors at that point.

davidrapson (approved):
👍 LGTM

luke-hill (approved):

PR #393, opened 2020-10-26
Grid docs [NP-1076]

davidrapson:
Adds some initial grid documentation using story examples. Focussed on documenting _only_ the styles of grid we use on public-website at the moment. Consider this a starting point, will work with @soniaturcotte to refine this.

Screenshot 2020-10-26 at 12 41 10

Also make some supporting changes to grid styles:

- Removes any commented out styles to avoid confusion
- Removes any unused mixins
- Inlines all remaining mixins into the grid object styles as this is the only place they are used. Reduce the number of places we need to look to understand the grid styles


davidsauntson (approved):
👍

davidrapson:
@davidsauntson Cheers! Going to keep this open until @soniaturcotte has had a chance to review

soniaturcotte:
Looks great!

davidrapson:
For reference this is the revised version:

image

PR #392, opened 2020-10-26
Page review spacing [NP-1030]

davidrapson:
Fixes spacing on page review date. Previously the page review styles were fighting with paragraph-s styles. As part of this PR I've:

- Migrated the page review component to the new docs format
- Added paragraph style mixins to allow more granular sharing of text styles (making a distinction between shared _formatting_ and shared _spacing_)
- Fixed BEM conventions within the page review component

Fixes NP-1030

PR #394, opened 2020-10-26
feature: translations [NP-974]

baseonmars:
This PR replaces hard coded strings and CSS translations with Rails I18n compatible translation strings.

- [x] actually translate the welsh


haml/_footer.html.haml
```diff
@@ -77,10 +77,6 @@
       = render "@citizensadvice/design-system/haml/logo_clickable", locals: { root_path: root_path }
       %div
         %p.paragraph-s
-          Copyright ©2020 Citizens Advice. All rights reserved.
+          t('cads.shared.copywrite')
```
davidrapson, line 80: `copyright` instead of `copywrite`?

locales/en.yml
```diff
@@ -3,3 +3,38 @@ en:
...
+      important: Important
+      adviser: Adviser
+    form:
+      optional: optional
+    logo:
+      title: Citizens Advice homepage
+    oisc_warning:
+      close: Close this warning
+    page_review:
+      body_html: Page last reviewed on <span class="cads-page-review-date">%{date}</span>
```
davidrapson, line 19: Could we avoid having HTML in the translations? I've also just updated this markup as part of https://github.com/citizensadvice/design-system/pull/392 which would conflict with this. This could probably be:

```
page_review
  last_reviewed: Page last reviewed on
```

davidrapson (changes requested):
👍 Looks great, couple of :neckbeard: nit picks but fab work. (Hope this isn't too pre-emptive)

locales/en.yml
baseonmars: It's a good idea to have HTML in the translation so that the full context of the translation is taken into account by the translator. 

Are you sure that the correct localisation in welsh would put the date as the last item in the sentence?

haml/_footer.html.haml
baseonmars: yes! 🤦‍♂️

locales/en.yml
davidrapson: Ah interesting. Fairly sure in the case of Welsh the construction is the same, but I take your point. I still think it makes things brittle, but I reckon the part I object to most is having _class names_ in here. Looking at the component all `cads-page-review-date` does is make the text bold, so perhaps this could be:

page_review:
      body_html: Page last reviewed on <strong>%{date}</strong>

haml/_website_feedback.html.haml
```diff
@@ -1,3 +1,3 @@
 %p.paragraph-s.cads-website-feedback
-  %a{href: 'https://www.research.net/r/J8PLH2H', title: "Let us know", target: "_blank"}> Is there anything wrong with this page? Let us know
+  %a{href: 'https://www.research.net/r/J8PLH2H', target: "_blank", rel: "noopener"}= "#{t('cads.website_feedback.body')} "
```
davidrapson, line 2: Just saw `rel: "noopener"` ❤️

locales/en.yml
baseonmars: Sure that's not a problem.

scss/6-components/_page_review.scss
```diff
@@ -2,7 +2,7 @@
   color: $cads-language__secondary-text-colour;
   margin-bottom: 0;
 
-  .cads-page-review-date {
+  strong {
```
davidrapson, line 5: This can go as `<strong>` is already bold

davidrapson (approved):
👍

baseonmars:
I've submitted translations for review.

davidsauntson (approved):
🚀 lush

baseonmars:
@luke-hill hey Luke, you mentioned these tests might have trouble - what should I do? Mark them for future_release or are they simple enough to fix?

luke-hill:
So it's 2-fold actually. Because of the way the browserstack tests will run.

So they need fixing "and" tagging as future_release. But for now if you want to get this ticket done quickly, just tag them as future release and I can fix them up later.

PR #373, opened 2020-10-22
refactor: Tidy design-system dockerfiles [NP-1037]

simonwgill:
Tidying of dockerfiles and general improvements 

Since this is going to be a chunky PR, I'm going to drop checkpoint commits in that gather up related changes in a (hopefully) more organised fashion than trying to dump the inevitable iterative commits that come from playing with a Jenkinsfile. It'll stay in draft until that last step is complete.

PR #402, opened 2020-10-27
feature: Enable cache-from feature to speed up builds (NP-1159)

simonwgill:
Mostly the same as https://github.com/citizensadvice/design-system/pull/373, but rebased to master and adding the code to stop the build after collecting the docker-compose logs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,15 +46,15 @@ pipeline {
                         // Pull the master images and any previous builds if we're on a different branch
                         // docker-compose only looks in the local images and doesn't try to pull when building
                         ['story-server', 'backstop', 'wcag', 'ruby'].each {
-                            docker.image("${docker_registry}/design-system:${it}").pull()
-                            if (env.BRANCH_NAME != 'master') {
-                                // Ignore failures from docker - it's probably an Image Not Found.
-                                // Other errors like out of disk space will cause problems in other commands
-                                try {
-                                    docker.image("${docker_registry}:${it}-${env.CA_STYLEGUIDE_VERSION_TAG}").pull()
-                                } catch (Exception e) {
-                                    echo "Error pulling ${docker_registry}:${it}-${env.CA_STYLEGUIDE_VERSION_TAG}"
+                            // Ignore failures from docker - it's probably an Image Not Found.
+                            // Other errors like out of disk space will cause problems in other commands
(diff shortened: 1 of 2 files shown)


pedantic-git (approved):
Nice. Not seeing anything to whinge about here (except maybe the name of the PR?) so ✅

Jenkinsfile
```diff
@@ -37,7 +41,40 @@ pipeline {
...
+                        ['story-server', 'backstop', 'wcag', 'ruby'].each {
+                            // Ignore failures from docker - it's probably an Image Not Found.
+                            // Other errors like out of disk space will cause problems in other commands
+                            try {
+                                docker.image("${docker_registry}/design-system:${it}").pull()
+                                if (env.BRANCH_NAME != 'master') {
+                                        docker.image("${docker_registry}/design-system:dev_${it}_${env.CA_STYLEGUIDE_VERSION_TAG}").pull()
+                                }
+                            } catch (Exception e) {
+                                    echo "Error pulling ${docker_registry}/design-system:dev_${it}_${env.CA_STYLEGUIDE_VERSION_TAG}"
```
baseonmars, line 57: Does this one also need to re-throw?

simonwgill: Not this one. When you run the first build, this error will always trigger because the image doesn't exist. It's possible that there are other errors - so it logs the attempt to make it easier to find things in that case.

baseonmars: It'd be cool to provide a different error message for a 404 vs other errors... perhaps something for another PR tho.

baseonmars (approved):
Looking good!

Jenkinsfile
simonwgill: It looks like this in the logs - so the actual message is provided by the result of docker pull

```
+ docker pull 979633842206.dkr.ecr.eu-west-1.amazonaws.com/design-system:dev_story-server_pr-402_5c4f209
Error response from daemon: manifest for 979633842206.dkr.ecr.eu-west-1.amazonaws.com/design-system:dev_story-server_pr-402_5c4f209 not found: manifest unknown: Requested image not found
[Pipeline] echo
Error pulling 979633842206.dkr.ecr.eu-west-1.amazonaws.com/design-system:dev_story-server_pr-402_5c4f209
```

PR #403, opened 2020-10-27
bugfix: Firefox new tab issue [NP-1172]

luke-hill:
**Previously:**

When firefox loads up a new tab. It first loads `about:blank`. Because this is a static white page. Selenium detects the page has loaded correctly and continues on.

Shortly after a "redirect" then changes the URL. And selenium then goes into "load" mode again and waits for the page to load. However by this time, it's too late, as we've checked the url and triggered a failure.

NB: This doesn't appear to affect chrome.

**Now:**

Explicitly wait until the url "isn't" `about:blank` if on firefox, and we've asked it to wait for a new url.

### DIFF
--- a/testing/features/footer.feature
+++ b/testing/features/footer.feature
@@ -10,7 +10,7 @@ Feature: Footer component
   Scenario: Footer is present
     Then a report problem with this page link is present
     And each header item has at least 1 link below it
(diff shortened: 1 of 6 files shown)


davidrapson (approved):
👍

PR #404, opened 2020-10-27
refactor: Split static ruby and js testing [NP-1161]

simonwgill:
Split up the ruby linting and node linting/testing so that they can use two different container images. The ruby image is now down to using ~430MB from the ~1GB it was before. This should speed up that part of the setup and reduce bandwidth costs as well.

Decided against using multi-stage builds in the end, as it seems to bust the cache in docker-compose with few mechanisms to reuse existing images. This would be fine for a final build container (like Maddie's rubocop image) but is less useful for getting quick testing done.

- Collected image names in Jenkinsfile into a map to reduce messy and repetitive string substitution
- Broke out the contents of bin/test into the Jenkinsfile and ca-styleguide Dockerfile
- Removed the seemingly unnecessary scripts in bin
- Renamed to cucumber dockerfile to ruby_host
- Moved ruby_host to use an alpine base image
- Vendored in wait-for as a script to this repository
- Grouped gems to split for use between testing and running the ca-styleguide web server
- Replaced local use of `bin/test` with `bundle exec rake check`


bin/test
```diff
@@ -1,11 +0,0 @@
-#!/bin/bash
```
davidrapson, line 1: Would be great if we could keep this top level `bin/test` even if it's not used for Docker. It's useful to run locally.

davidrapson (changes requested):
Small request from me to keep `bin/test` around. Can't comment too much beyond that other that 💯 to that size improvement and that this looks way simpler!

simonwgill:
> Small request from me to keep `bin/test` around. Can't comment too much beyond that other than to say 💯 to that size improvement and that this looks way simpler!

How do you feel about moving these commands into the rakefile so the commands can be stored in one place and used reasonably from both the dev command line and from the Jenkinsfile?

docker/Dockerfile.ruby_host
```diff
@@ -0,0 +1,23 @@
+FROM ruby:2.6-alpine as build
+
+ENV BUNDLER_VERSION 2.1.4
+ENV LANG=C.UTF-8 \
+    BUNDLE_JOBS=6 \
```
luke-hill, line 5: are these env keys used anywhere?

docker/Dockerfile.ruby_host
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /app
+
+RUN gem install bundler -v $BUNDLER_VERSION && bundle install --with test && gem cleanup
+
+RUN addgroup --gid 1000 ruby && \
+    adduser --home /home/ruby --uid 1000 --ingroup ruby --disabled-password ruby
+
+COPY . /app
+
+RUN rm ./testing/artifacts/*/.keep && chown -R 1000:1000 /app
```
luke-hill, line 21: Do we need to remove these files as they preserve the folder structure (Or will they still preserve structure after removing them?)

Rakefile
```diff
@@ -10,8 +10,51 @@ namespace :design_system do
...
-desc "Rubocop Linting"
-task :rubocop do
-  puts "Running rubocop linter"
-  system("bundle exec rubocop") || raise
+desc "Run all checks in the local context"
+task check: ["ruby:lint", "npm:test"]
+
+namespace :ruby do
+  desc "Lint ruby files"
+  task lint: [:rubocop, :haml_lint]
```
luke-hill, line 18: Unsure if this'll crop up but you may wanna go `%i[rubocop haml_lint]` here. Might fall foul

Rakefile
```diff
@@ -10,8 +10,51 @@ namespace :design_system do
...
+    system("npm run lint")
+  end
+
+  desc "Install Node modules"
+  task :install do
+    puts "Installing modules"
+    system("npm install") || raise
+  end
+
+  desc "Build Node files"
```
luke-hill, line 55: Minor OCD, order of these. Lets have `install` at the bottom as the one relied on by all others.

Rakefile
```diff
@@ -10,8 +10,51 @@ namespace :design_system do
...
+  end
+
+  desc "Run linting jobs in node"
+  task lint: :build do
+    puts "Running Node lint"
+    system("npm run lint")
+  end
+
+  desc "Install Node modules"
+  task :install do
```
luke-hill, line 50: Is there a reason for dual dependencies, i.e. could these be ran separate. If not why don't we combine them as `npm install && npm build`

luke-hill (commented):
Mainly style comments. All seems really good though.

I've not looked at the wait-for.sh file as that just looked 🤮

docker/Dockerfile.ruby_host
simonwgill: They're parameters for bundle to manage the download jobs.

docker/Dockerfile.ruby_host
simonwgill: They're only removed from the folder structure inside the container after the copy, so the mount point is empty and should allow writes properly from the non-root processes inside the container.

Rakefile
simonwgill: Fall foul of the style rules? It passes happily on the branch at the moment.

Rakefile
simonwgill: Sure, easy to do.

Rakefile
simonwgill: Possibly - I can easily imagine other tasks which depend on install but not on build.

docker/Dockerfile.ruby_host
luke-hill: I'm guessing they're auto used then. No probs, I guess I've never configured bundler like this before. I half expected them to be used like the bundler version is.

simonwgill: That was an extra I added over Maddie's rubocop example - to absolutely ensure it's installing the version we are expecting.

bin/test
simonwgill: @davidrapson Are you happy with the changes?

davidrapson: Just tried this out locally and words great. Lovely stuff

davidrapson (approved):
👍

luke-hill (approved):
Think afterwards it would be good to see how we're meant to use these manually e.t.c.

PR #405, opened 2020-10-27
bugfix: Mark failing test [NP-1172]

luke-hill:
The Footer test mentioned here fails on Firefox

I have created a breakout ticket in NP-1173 to tackle this issue with more details in it. TL;DR - On Firefox <80 this functionality isn't working as specified.

### DIFF
--- a/testing/features/footer.feature
+++ b/testing/features/footer.feature
@@ -14,6 +14,7 @@ Feature: Footer component
     And a copyright notice is present
     And a company info is present
 
+  @failing @NP-1173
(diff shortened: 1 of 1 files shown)


luke-hill:
@pedantic-git We may need to add extra info to the ticket or take an approach of whether we should fix this e.t.c. - More details on Jira 1173

pedantic-git (approved):
Approved. I've added the frontend team to that case so they can discuss the failure over there.

PR #406, opened 2020-10-27
feature: locale switch url [NP-1165]

baseonmars:
This features allows the design-system locale to be set via the `locale` query string parameter.

This works with both the main site and the component iframe pages.

Add `&locale=cy` to the end of a url to test. Currently the translation documentation is the only translated feature (pending updated strings).

I need to test whether the polyfill is needed, I'm fairly sure it will..

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -1,3 +1,5 @@
+require('url-polyfill');
+
 import { addons } from '@storybook/addons';
 import { create } from '@storybook/theming/create';
 import { version } from '../package.json';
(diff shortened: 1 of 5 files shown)


davidrapson (approved):
👍

luke-hill (approved):
All seems good. Is it worth having a bog-standard unit test for this (Not sure if this is possible).

baseonmars:
> All seems good. Is it worth having a bog-standard unit test for this (Not sure if this is possible).

I'm gonna say no on this one as the tests will be using it, and the test infrastructure required to test this would be many many lines of code. I think we get the feedback we need from whether the tests run or not. 

Hope that's ok.

baseonmars:
Screenshot 2020-10-27 at 15 46 52

PR #407, opened 2020-10-27
bugfix: Footer: Report this Page tests working! [NP-1173]

luke-hill:
So we now do a 3 step check during out wait for switch context

1) Wait until 2 windows have been detected
2) Once we have 2 windows (Note that a Tab and window are treated the same in this subsection of the Capybara API), detect which window is **not** the one we have current browsing context in (Always the other one) - Then switch to it
3) Wait until the current windows url is not `about:blank`

This should ensure we are always on the correct window, and that we have the correct loaded context.

### DIFF
--- a/testing/features/support/helpers/page.rb
+++ b/testing/features/support/helpers/page.rb
@@ -75,7 +75,7 @@ module Helpers
     end
 
     def second_window
-      (page.windows.to_a - [page.current_window]).first
+      page.windows.detect { |window| !window.current? }
     end
 
     def wait_for_new_url
--- a/testing/features/support/helpers/page.rb
+++ b/testing/features/support/helpers/page.rb
@@ -25,7 +25,8 @@ module Helpers
     end
 
     def switch_to_newly_opened_window!(new_page: false)
-      page.switch_to_window(page.windows.last)
+      wait_for_second_window
+      page.switch_to_window(second_window)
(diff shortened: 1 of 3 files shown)


testing/features/support/helpers/page.rb
```diff
@@ -69,6 +70,14 @@ def height(fallback: 800)
       ENV.fetch("BROWSER_HEIGHT", fallback)
     end
 
+    def wait_for_second_window
+      Selenium::WebDriver::Wait.new.until { page.windows.length == 2 }
+    end
+
+    def second_window
+      (page.windows.to_a - [page.current_window]).first
```
luke-hill, line 78: The array casting with the first object is because Capybara Collections aren't Enumerable. This is also present in some other areas of the codebase

davidrapson (approved):
👍

pedantic-git (approved):
I think this is fine. I assume it still works even for the browsers (e.g. non-Firefox?) that _don't_ open a new window?

luke-hill:
So I actually found a way of querying the windows much easier when I dug into the docs `#current?` returns the handle that is equal to the current window context.

As for new window vs tab, Selenium treats them all as "Windows" with a "handle id". When you call `switch_to` with an id, it just switches and makes that the top level context

PR #409, opened 2020-10-27
Forms design review [NP-1032]

davidrapson:
**Breaking changes**:

* Forms: Don't style raw `input` and `textarea` elements
* Forms: Updated all forms classnames to use consistent BEM conventions
* Forms: Removed `radio_group_small`, `radio_group` now accepts a modifier
* Forms: Refactor radio component. Fixes line-height for wrapping text

image

**New features**:

* 🆕 Forms: Added character width modifiers to inputs: `2ch`, `4ch`, `8ch`, `16ch`

image

**Bugfixes**:

* Forms: Make success message a block element, no width constraints
* Forms: Increase spacing between labels and hints
* Forms: Normalise spacing between fields
* Forms: No width constraints on inputs or textareas


scss/2-tools/_interactive-elements.scss
```diff
@@ -1,6 +1,4 @@
 @mixin cads-interactive-element() {
-  @include cads-transition-animation();
```
davidrapson, line 2: Removed this as the places that called `cads-interactive-element` all then disabled `transition` afterwards. So this seems to be unintentional. Or better to require explicit use when needed.

davidrapson:
After chatting with @soniaturcotte realised that the "compact" style should actually be two variants. Inline and small.

image

CHANGELOG.md
```diff
@@ -1,7 +1,23 @@
-**Breaking changes**
 
-* Rails I18n API translated haml templates - NP-974
-* NP-1074 Typography updates (typography scoping and placeholders)
+**Breaking changes**:
```
davidsauntson, line 2: This is really nice! 📝

davidsauntson (approved):
📻  lgtm

PR #412, opened 2020-10-27
fix: the sample pages

baseonmars:
Sample pages were not loading, as result of importing haml needs to be cast to a string when not passed through the translation helper.

This is due to the the default export of the haml–loader not being recognised as a string by the HTML Storybook App (it's an augmented string object but it's still a string deep down inside!)


davidrapson (changes requested):
👍 Cool beans, although I think we can remove this from the changelog (as it doesn't affect the published package) and instead run `npm run docs:build` and commit those changes here. We only need to fix the github pages build right?

davidsauntson (approved):
🧶 
might be worth adding the sample pages into VRT but not in this PR if want to get the fix in ASAP.

baseonmars:
> 👍 Cool beans, although I think we can remove this from the changelog (as it doesn't affect the published package) and instead run `npm run docs:build` and commit those changes here. We only need to fix the github pages build right?

I guess so... cool beans.

davidrapson (approved):
image

davidrapson (approved):
🚀

PR #413, opened 2020-10-27
fix: typography design system [NP-1074]

davidsauntson:
Refactored and rescopes typography styles for maximum BEM:
 - remove utility header classes and directly styled `hX` and `p` elements
 - replaced directly styled elements with placeholders
 - extend placeholders into classed selectors where required
 - add `.cads-prose` scope to contain classless elements from CMS
 - add scoping class to example pages
 - update `cads-adjacent-heading-margin` mixin to be scoped to `.cads-prose`
 - add initial typography documentation & update changelog


scss/2-tools/_typography.scss
```diff
@@ -4,35 +4,15 @@
     update the list of elements that we want to check against.
 */
 @mixin cads-adjacent-heading-margin($m) {
```
davidrapson, line 6: Should this set a default spacing? Seems to be `$cads-spacing-6`?

scss/2-tools/_typography.scss
```diff
@@ -4,35 +4,15 @@
...
+  .cads-prose ul + &,
+  .cads-prose img + &,
+  .cads-prose details + &,
+  .cads-prose table + &,
+  .cads-prose .cads-callout + & {
     margin-top: $m;
   }
 }
-
-@mixin cads-paragraph() {
```
davidrapson, line 20: Guessing this is because we're standardising on placeholder extends? Fine if so!

scss/4-elements/_typography.scss
```diff
@@ -65,15 +50,55 @@ h4,
   }
 }
 
-p,
-.paragraph {
-  @include cads-paragraph();
-
+%p {
```
davidrapson, line 53: :neckbeard:  Having `%p` but then `%paragraph--small` feels a touch confusing. Could this be `%cads-paragraph` and `%cads-paragraph-small`. Don't think we need the `--` modifier style as it's not really a modifier. 💯 to sharing the styles though.

scss/6-components/_callout.scss
```diff
@@ -119,3 +123,10 @@
     margin-top: $cads-spacing-4;
   }
 }
+
+.cads-prose .cads-callout__title {
```
davidrapson, line 127: Guessing this is because callouts mainly appear in body content? Makes sense, but could we add an explanatory comment?

scss/6-components/_contact_details.scss
```diff
@@ -1,4 +1,6 @@
 .cads-contact-details {
+  @extend %prose;
```
davidrapson, line 2: Rather than extending here would it make sense to add `cads-prose` explicitly to the contact details template? Personally find that clearer for these kinds of scoped styles.

scss/6-components/_oisc-warning.scss
```diff
@@ -34,32 +42,36 @@
       background-color: rgba(0, 0, 0, 0.05);
     }
   }
+}
+
+.cads-prose .cads-oisc-warning__title {
```
davidrapson, line 47: With the change of scope in https://github.com/citizensadvice/public-website/pull/458 we can likely avoid the need for this override?

scss/6-components/_page_review.scss
```diff
@@ -1,6 +1,6 @@
-/** @define page-review */
-.cads-page-review {
-  @include cads-paragraph-s();
+.cads-page-review,
+.cads-prose .cads-page-review {
```
davidrapson, line 2: With the change of scope in citizensadvice/public-website#458 we can likely avoid the need for this override?

styleguide/foundations/typography.stories.mdx
```diff
@@ -0,0 +1,56 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Foundations/Typography" />
+
+# Typography
```
davidrapson, line 5: ❤️ ✍️ 🔠

styleguide/sample-pages/advice-collection/_advice_collection.html.haml
```diff
@@ -4,7 +4,7 @@
   .cads-grid-row
     .cads-grid-col-12
       = render partial: "@citizensadvice/design-system/haml/breadcrumb", locals: {breadcrumbs: {"breadcrumb_links" => [{"url" => "/link", "title" => "Tables"}]}}
-%main.cads-main-content
+%main.cads-main-content.cads-prose
```
davidrapson, line 7: Could we update this to reflect the narrow scoping from https://github.com/citizensadvice/public-website/pull/458?

davidrapson (changes requested):
A few 🤓 comments but nothing major. Lovely work. Really satisfying to see.

scss/2-tools/_typography.scss
davidsauntson: I can do - wonder if forcing a consumer to specify is better in the long run though? 🤷

scss/4-elements/_typography.scss
davidsauntson: agree - I'll change all typography placeholders to `cads-`

scss/2-tools/_typography.scss
davidrapson: I guess it comes down to if the adjacent margin rules are a fixed design, or something that should be flexible. Reckon it's fine go with this for now though

scss/6-components/_oisc-warning.scss
```diff
@@ -2,16 +2,24 @@
    The Office of the Immigration Services Commissioner (OISC)  Warnings
    ========================================================================= */
 
+%cads-oisc-warning__title {
```
davidrapson, line 5: This placeholder is only used once so we could (re)inline this now.

davidrapson (approved):
👍 Nice!

davidsauntson:
@soniaturcotte agreed to review this work with second PR for NP-1074

scss/6-components/_oisc-warning.scss
davidsauntson: doh!  will pick up in the list work I'm doing now

PR #415, opened 2020-10-28
Fix: lists design system [NP-1074] 

davidsauntson:
- Removes list utility classes from the design-system, with the exception of `.cads-list-no-bullet` which is widely used in the public-website repo
- Remove uses of `.cads-list--no-bullet` from design-system
- Align list markers with grid in prose content
- Fix issue with paragraphs in list items from migrated content


scss/4-elements/_lists.scss
```diff
@@ -1,5 +1,4 @@
-li,
```
davidrapson, line 1: ❤️

scss/4-elements/_lists.scss
```diff
@@ -10,42 +9,80 @@ li,
...
+  margin: 0 0 $cads-spacing-4 0;
+  list-style-type: none;
+
+  li {
+    display: inline-block;
+    margin-right: 1rem;
+  }
+
+  // Reverse margins if this is right aligned
+  .cads-align-right & {
```
davidrapson, line 42: Is this only used in the header? If so maybe we should inline these styles there?

scss/4-elements/_lists.scss
```diff
@@ -10,42 +9,80 @@ li,
...
+  li {
+    @extend %cads-list-item;
+  }
+}
+
+.cads-list-no-bullet {
+  @extend %cads-list-no-bullet;
+}
+
+.cads-prose {
```
davidrapson, line 64: Style preference: Does it make sense to have a single file for the prose styles, or happy defining this inline like this?

scss/6-components/advisernet/_related-content.scss
```diff
@@ -66,3 +66,11 @@
     }
   }
 }
+
+// :-( this whole file is a bit much
```
davidrapson, line 70: Yeah, fine for this to be a bit gnarly for now as we're going to pull this out anyway

styleguide/sample-pages/advice-collection/_advice_collection_adviser.html.haml
```diff
@@ -27,6 +27,26 @@
                 %p The important callout should be used for any important snippet of text that has serious and/or legal implications if the client does not follow the advice.
             %p
               Every page should have some lorem ipsum. So here it is.
+            %ul
```
davidrapson, line 30: Maybe we should create a "kitchen sink" sample page with this and a few more prose specific examples?

davidrapson (approved):
👍 Nice! A few comments but nothing blocking (although unsure if anything changes based on your mixin suggestion?)

davidsauntson:
Thanks! - I think I will do the mixin stuff as a second PR and work on the advisernet template stuff first.  That way there won't be another breaking change later on 👍

scss/4-elements/_lists.scss
davidsauntson: yup good idea, I've made a new objects file `_prose` in section 5 👍

styleguide/sample-pages/advice-collection/_advice_collection_adviser.html.haml
davidsauntson: I think the public advice collection one covers most things, I've added lists to that one too.  We can add more tho!

scss/5-objects/_prose.scss
```diff
@@ -0,0 +1,49 @@
...
+    li p {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.cads-prose {
+  @extend %cads-prose;
+
+  .cads-list--no-bullet {
```
davidrapson, line 46: Does this need to be `cads-list-no-bullet` too?

davidsauntson: 🤦

soniaturcotte (approved):
LGTM

luke-hill (approved):

PR #419, opened 2020-10-29
Sass output tests

davidrapson:
Uses the technique used in govuk-frontend outlined at https://obyford.com/posts/testing-sass-with-jest/ to add some basic unit tests for sass output. Thought this was worthwhile particularly as we are making breaking changes.

Does two things to begin with:

1. Asserts that settings and tools don't output any CSS. Needed to update some comments and move where font styles lived to make this work
2. Runs a snapshot test on the grid styles. There's a lot of generated CSS here so useful to have a snapshot for what this _should_ be.

If this looks useful I can add something to the backlog to look at adding unit tests for our sass functions too.


scss/5-objects/_grid.scss
```diff
@@ -1,3 +1,8 @@
+@import '../1-settings/breakpoints';
```
davidrapson, line 1: This allows us to assert the output of grid styles, and allows grids to be imported standalone. Because we also assert that settings and tools don't output any CSS this is safe to do.

baseonmars (approved):
smart. I like it.

PR #422, opened 2020-10-30
Split cads-main-content into multiple classes

davidrapson:
Whilst working on box lists I ran into an issue where we need to break out of the grid (to allow a background colour to spread full-width).

Currently the `cads-main-content class` prevents this:

- It constrains the width at the page level. This doubles up what `cads-grid-container` does so is both unnecessary and limiting (literally)
- It applies arbitrary styles to images
- It sets styles on `h1`s

Based on this I've split `cads-main-content` out into two classes:

- `cads-page-title`: only concerned with setting h1 styles. It also fixes an issue with the current typography styles where h1s are typically _outside_ `cads-prose` classes (in order to keep the scope as narrow as possible) so don't have the correct sizes
- `cads-page-content`: Only concerned with setting page level margins. Can be applied contextually based on the context of the page rather than assuming it's a top-level class
- Moved the img styles into cads-prose

### DIFF
--- a/scss/5-objects/_page-content.scss
+++ b/scss/5-objects/_page-content.scss
@@ -1,0 +1,15 @@
+// ============================================================================
+// Page content
+// ============================================================================
+
+/** @define page-title */
+.cads-page-title {
+  @extend %cads-h1;
+
+  margin-top: $cads-spacing-5;
+}
+
+/** @define page-content */
+.cads-page-content {
+  padding-bottom: cads-rem(80px);
+}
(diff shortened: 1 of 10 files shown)


scss/5-objects/_page-content.scss
```diff
@@ -0,0 +1,15 @@
...
+
+/** @define page-title */
+.cads-page-title {
+  @extend %cads-h1;
+
+  margin-top: $cads-spacing-5;
+}
+
+/** @define page-content */
+.cads-page-content {
```
davidrapson, line 13: I could go one step further and make this a margin utility class? But this seemed like a reasonable middle ground as this ~80px margin is specific for the bottom of page.

davidsauntson (approved):
👍

PR #426, opened 2020-10-30
refactor: Move root dockerfile [NP-1162]

simonwgill:
This should now give us a clean and fast build of the ca-styleguide container.

Removing .bundle/config turned out to be important to resolve conflicts between ca-styleguide and ruby_host. If the setting is still necessary for individual devs when working locally, then user-wide settings or environment variables will work for that.

- Replaced /Dockerfile with a cleaner cache-friendly docker/Dockerfile.ca-styleguide
- Consolidated naming around the old ca-styleguide name instead of the new story-server name
- Pushed .bundle/config into container configurations
- Fixed potential surprises by replacing npm install with npm ci in the Rakefile


davidrapson (approved):
👍 Nice!

pedantic-git (approved):
Looks good, I think!

.bundle/config
```diff
@@ -1,2 +0,0 @@
----
-BUNDLE_PATH: "vendor/bundle"
```
luke-hill, line 2: This alters the way in which gems are installed (Unless it's set elsewhere). Are we happy to do this?

EDIT: Seen we've altered the dockerfile. But this will affect people running locally. Can we keep this, and the dockerfile changes, or would that be the issue you were describing?

luke-hill (commented):
One quick question.

.bundle/config
simonwgill: From the running process in the container's point of view, it doesn't mean anything. So long as the path matches between install and exec.

Heroku's recommendation is not to check in the .bundle/config file (https://devcenter.heroku.com/articles/bundler-configuration#:~:text=bundle%2Fconfig%20file.,to%20configure%20Bundler%20on%20Heroku.) and I agree with them.

It could go in the .dockerignore and .gitignore files so people can still use a local copy of the file if they want, instead of user-config or environment variables?

luke-hill: I think that sounds a really good idea. In gitignore. Then it can be just like `.env` I'm happy with that approach.

luke-hill (approved):

PR #435, opened 2020-11-02
build: Tidy visual regression dockerfile [NP-1163]

simonwgill:
The wait-for-it script has been reported as unnecessary, and since that was the only reason we were building an image, it could be scrapped.

One thing I did notice is that it was tricky to match the version of BackstopJS being used by developers through the `bin/visual_regression` script installed by npm and the version grabbed by docker. The old method didn't use any version pinning, and BackstopJS doesn't supply major/minor/patch tags on their image repo.

I did try to use the Rakefile from the Jenkins agent context, but I forgot that the agents do not provide a version of ruby without using a container.

- Removed unnecessary image creation
- Added a way to match the currently installed version of BackstopJS


luke-hill (approved):
Some of this is a bit abstract for me. But I can't see any glaring errors.

pedantic-git (commented):
Nice.

PR #437, opened 2020-11-02
build: Final clean-up pass [NP-1224]

simonwgill:
This PR captures some of the issues that I've noticed with the way I've done other tickets under NP-1037.

- rake will now properly return an error if one of the subtasks fails - but will still run all of the expected subtasks
- Docker images are now tagged with just the branch instead of with branch+commit SHA so the PR builds will actually be reused in the cache
- Removed an unnecessary traefik config that wasn't doing anything
- Added a service to the compose file for running ruby source testing which doesn't require the whole grid setup to be started
- Removed the npm install from the Rakefile entirely - this is most likely to be run inside an already built container, and it was always replacing the node_modules which would mean potentially not testing the code running in the actual container.
- Fixed the sandbox images list to reference all the images in use.


Jenkinsfile
```diff
@@ -26,7 +26,8 @@ pipeline {
     }
     environment {
         DOCKER_TAG = "${env.BRANCH_NAME}_${getSha()}"
-        CA_STYLEGUIDE_VERSION_TAG = "${DOCKER_TAG.toLowerCase()}"
+        // Using the commit SHA would mean that every build gets a different tag and this busts the cache between PR builds.
+        CA_STYLEGUIDE_VERSION_TAG = "${env.BRANCH_NAME}"
```
pedantic-git, line 30: Does this mean that 'master' will always be the same version forever? I'm not saying that's a bad thing - it's pretty common practice in Docker, iirc.

pedantic-git (approved):
Amazing stuff 🎉

Jenkinsfile
simonwgill: No, there's a dev_$containername_master tag which will get updated to the latest build on master after each setup step.

Since it's not a project released through Docker and just a build cache, there's not much call for keeping old versions around.

docker-compose.yml
```diff
@@ -79,6 +79,14 @@ services:
       - testing/browserstack_ci.env
     depends_on: []
 
+  ruby-tests:
```
luke-hill, line 82: could we call this `ruby-style-tests` So we know it's not the physical ruby code being ran, but just checking the style.

luke-hill (commented):
This is really good stuff!! Minor nit on name but other than that looks 🆗

docker-compose.yml
simonwgill: I considered that, but I figured that if there's any rspec or similar tests going in here in the future, this would clearly be the container to run them in.

luke-hill: Hmmm oki. I don't think there would be, but seems sensible.

luke-hill (approved):

PR #438, opened 2020-11-02
Use npm version for release script

davidrapson:
Fixes a bugbear with the current release script:

We currently manually update the package.json version but this a) causes issues with formatting and b) always misses out the version in package-lock.json. `npm version` does this for us.

The default include making a git tag of the version too which is think is a good thing to do, but if we want to disable that for now we can add `--no-git-tag-version`.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -171,7 +171,7 @@ prompt([
               git
                 .add('.')
                 .commit(releaseMessage)
-                .push('origin', newVersionBranch, (gitErr2) => {
+                .push(['origin', '--tags'], newVersionBranch, (gitErr2) => {
                   if (gitErr2) {
                     showError(gitErr2, true);
                   }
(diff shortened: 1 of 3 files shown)


scripts/release.js
```diff
@@ -130,7 +114,7 @@ prompt([
         process.exit(1);
       }
 
-      const newVersionBranch = `v${newVersion}`;
+      const newVersionBranch = `release/v${newVersion}`;
```
davidrapson, line 117: Scoping this branch because the default behaviour of `npm version` creates a git tag at `vx.x.x`

baseonmars (commented):
Very nice!

You'll need to update the build script to push tags (`--tags`) if we do keep them included (no big opposition to them, but they do require a little more overhead to correct if the release goes wrong).

davidrapson:
@baseonmars Done. I think it's worth keeping them as we can then create a github release with the changelog which would be nice.

davidrapson:
@baseonmars Updated this to document pushing the tags at the publish stage. Doing it at the release state would publish the tag too soon probably.

baseonmars (approved):
Seems reasonable to me. 👍

davidrapson:
Closes https://github.com/citizensadvice/design-system/pull/454

PR #450, opened 2020-11-03
bugfix: Local Firefox Driver [NP-1065]

luke-hill:
Creates a patch that allows Firefox drivers to be created both locally and through grid (Browserstack appears to be unaffected).

The issue stems from the fact that the key for the browsername when created directly using `browser :firefox` appears to not be camelised, whereas the chrome one is being camelised. In Se4 this case mismatch is then bubbling up as an invalid key error.


luke-hill:
Related work to follow from this:

https://citizensadvice.atlassian.net/browse/NP-1234?filter=15271

cc/ @mrdaniellewis / @adriana-rus

celonicolosi (approved):
LGTM!

PR #452, opened 2020-11-04
feature: Automation Candidate Triage [NP-991]

luke-hill:
Now we have the Design-System running stably again (Admittedly with some belt-n-braces approaches). We should look to re-enable some more testing candidates and also streamline our existing ones.

This is still a work in progress and browsers/versions are very much up for discussion. But fundamentally this time we introduce iPhone's into our testing matrix.

The code for enabling these was done last Sprint, but I've only just wanted to introduce them now after running them privately for a little while. You'll note the format of BROWSERSTACK_CONFIG_OPTIONS is slightly different. Although this is not mandatory (Essentially the 3rd parameter on iPhone's is always the same as the 2nd, and as such it can be omitted - i.e. `iPhone8_12_12 == iPhone8_12`)

### DIFF
--- a/testing/features/oisc_warnings.feature
+++ b/testing/features/oisc_warnings.feature
@@ -32,6 +32,7 @@ Feature: OISC Warning components
       Then the OISC title is visible
       And the OISC message is visible
 
+    @not_mobile @NP-1230
     Scenario: Sticky component follows you down the page
       When I scroll to the bottom of the page
       Then the OISC component is visible at the top of the viewport
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,14 +1,13 @@
 deployBranches = ['master']
 
 configurationTypes = [
-    //     ['Windows_10_83', 'chrome'], - These seem to be "too" popular!
     ['Windows_10_85', 'chrome'],
-    ['Windows_10_80', 'firefox'],
-    ['Windows_10_76', 'firefox'],
+    ['Windows_10_81', 'firefox'],
     ['Windows_7_80', 'chrome'],
     ['Windows_7_78', 'firefox'],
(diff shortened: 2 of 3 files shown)


testing/features/oisc_warnings.feature
```diff
@@ -32,6 +32,7 @@ Feature: OISC Warning components
       Then the OISC title is visible
       And the OISC message is visible
 
+    @not_mobile @NP-1230
```
luke-hill, line 35: This appears to contradict the W3C spec. And needs further clarification from myself. Furthermore some of the XCUI actions appear to not be liked by iPhones. It's likely there will be others on this PR too.

baseonmars: Surely there is a way we/you can find this out before merging?

luke-hill: Finding if there are other's out - There aren't any more, so that's good. We're green.

Why is this one a failure. I'm waiting on feedback from senior selenium committers. I have made a ticket to investigate this for next sprint. I cannot explain why though. It only breaks one test on mobile, so I don't wanna hold off too much. It's to do with the way in which the DOM is being rendered or at least "viewed" as rendered.

baseonmars (commented):
Final question... you mentioned about running these in parallel... is that something that can be included in this PR? Adding these three extra builds will increase build times quite significantly

luke-hill:
I've created NP-1237 to do that next Sprint. Given the issues we had on public-website-cucumber. I'd rather let that run for a bit, and this still linearly, and then refocus back in a week or so if that's ok?

baseonmars (approved):
Gonna approve on the basis we review build times over this week and if productivity is negatively impacted we work on a solution.

PR #455, opened 2020-11-04
fix: spacing between bottom of related content and footer. [NP-1029]

baseonmars:
Fix the spacing between related content, the toolbar and footer. This should be 40px (2.5rem) at the smallest breakpoint.

Spacing between the two titles and each list item is already 16px/1rem - assuming this was fixed in a previous PR related to typography gap analysis?

**With toolbar, smallest/mobile breakpoint (logged in):**

(will need additional PR on public-repo to remove top padding from toolbar)

Screenshot 2020-11-04 at 15 00 40

**Logged out, smallest/mobile breakpoint (logged out):**

Screenshot 2020-11-04 at 14 59 31

**Advisernet related content**

This change also affects advisernet related content - spacing on the list is now a uniform 1rem/16px (VRTs updated to reflect this)

Screenshot 2020-11-04 at 16 48 42


baseonmars:
related: https://github.com/citizensadvice/public-website/pull/488

davidsauntson (commented):
I think the VRTs will pick this change up

baseonmars:
I've updated the VRTs - they did highlight a problem with my initial PR. I fixed that and realised there was an opportunity to normalise the spacing around the Adviser content in the advisernet related content - so made that change and updated VRTs.

davidsauntson (approved):
🌞 nice

scss/6-components/advisernet/_related-content.scss
```diff
@@ -9,6 +9,10 @@
       display: none;
     }
 
+    ul {
+      margin-bottom: 1rem;
```
davidsauntson, line 13: lol, should be a var

baseonmars: fml... yes.

davidsauntson: <3

PR #457, opened 2020-11-04
v2.0.0 release

davidrapson:
**Breaking changes**:

* i18n: Rails I18n API translated haml templates (NP-974)
* Typography: removed typography utility classes (NP-1074)
* List: removed list utility classes (with exception of `.cads-list-no-bullet`) (NP-1074)
* Forms: Don't style raw `input` and `textarea` elements
* Forms: Updated all forms classnames to use consistent BEM conventions
* Forms: Removed `radio_group_small`, `radio_group`. Inline and small are now two separate modifiers
* Forms: Refactor radio component. Fix line-height for wrapping text
* Layout: Split `cads-main-content` out into `cads-page-title` and `cads-page-content`

**New features**:

* 🆕 Forms: Added character width modifiers to inputs: `2ch`, `4ch`, `8ch`, `16ch`

**Bugfixes**:

* Forms: Make success message a block element, no width constraints
* Forms: Increase spacing between labels and hints
* Forms: Normalise spacing between fields
* Forms: No width constraints on inputs or textareas
* Typography: fixed adjacent header margin for headers with 'Add reference' buttons
* Lists: align list markers with grid
* Lists: cater for `p` inside `li` in prose content

PR #459, opened 2020-11-04
Build docs

davidrapson:
Missed out from https://github.com/citizensadvice/design-system/pull/457 due to a minor error in the release script.


scripts/release.js
```diff
@@ -136,7 +136,7 @@ prompt([
         fs.writeFileSync(changelogPath, newChangelog, 'utf8');
 
         // Rebuild the docs
-        if (!newVersion.indexOf('alpha')) {
```
davidrapson, line 139: This was checking if the result of `indexOf` is falsey, but `indexOf` returns `-1` or a positive index, both of which resolve to false:

```
> !(-1)
false
> !(1)
false
```

baseonmars (approved):

PR #462, opened 2020-11-05
Docs tweaks

davidrapson:
A few small docs tweaks:

- Fixes an issue where the page title was always "CA Design System" rather than including the current story title
- Inline contact details into welcome guide. It was one sentence.
- Link to releases and changelog instead of inlining into Storybook. This is a partially generated doc so the formatting is a little troublesome in storybook and better to link off to github and the release page now we have it.

### DIFF
--- a/styleguide/component.stories.js
+++ b/styleguide/component.stories.js
@@ -53,7 +53,8 @@ ${usage || ''}`,
 
 // Storybook section setup
 export default {
-  title: '3: Components',
+  title: 'Components',
+  id: '3: Components',
   parameters: {
     options: {
       showPanel: true,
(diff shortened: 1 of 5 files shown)


baseonmars (commented):
Great stuff! Whilst you're in here tweaking, do you know of a way to reorder the sidebar?

Screenshot 2020-11-05 at 10 13 10

baseonmars (approved):
Will approve anyways as the PR itself good to go.

davidrapson:
Merged the old components section into the new one (whilst maintaining existing URLs)

image

baseonmars (approved):

styleguide/component.stories.js
```diff
@@ -53,7 +53,8 @@ ${usage || ''}`,
 
 // Storybook section setup
 export default {
-  title: '3: Components',
+  title: 'Components',
```
davidrapson, line 56: This was the trick. Sets the title to the new name but keeps the `id` as the old one for the URLs.

PR #463, opened 2020-11-05
Migrate asset_hyperlink docs

davidrapson:
Migrate asset_hyperlink docs to the new format. Applies haml-lint rubocop checks to the file and moves the styles out of the generic "misc" file.


davidrapson:
@luke-hill Tagged this test as `@future_release` as it's moving to the new format. Any way we can update this ahead of time?

davidsauntson (commented):
👍 

Small nitpick re: accessibility.  Voice over reads `mb` as millibars, but will read `MB` as megabytes.  This might just be macOS behaviour, but as `MB` is the standard unit for megabytes we should probably use it.  I know you haven't changed it but seeing as you're there... 😄 

Q: does the type actually do anything?  If so, should we list what types are supported?

davidrapson:
Good shout re: voiceover. And great catch on the `size` param. It existed in the previous example data, but you are right I don't think it's actually used.

davidrapson:
@davidsauntson Have made both of those updates.

davidsauntson (approved):
👍

testing/features/asset_hyperlink.feature
```diff
@@ -1,3 +1,4 @@
+@future_release
```
luke-hill, line 1: can we pop a `@v2.1.0+` tag on this. So I know when to look at it again.

PR #464, opened 2020-11-05
Callout markup bug

davidrapson:
**Bugfix**: Don't render empty .cads-callout-label for standard callouts (no label)

Also enables haml-lint rubocop checks for this component and fixes any warnings. Plus removes some misleading examples of nested callouts (we have a backlog task to document nested callouts more generally)


luke-hill (approved):
This was a JIRA ticket I just made. Can you move it into this Sprint and just mark it as ready to deploy.

np-1244

PR #465, opened 2020-11-05
v2.1.0 alpha.0

davidrapson:
Alpha release including:

**Deprecations**:

* Removed standalone `logo_clickable` component. Only used within header and footer
* Removed standalone `website_feedack` component. Only ever used within footer

**Bugfixes**:

* Header: Fix spacing and grid sizing around search form (NP-1024)
* Footer: Fix spacing and responsive styling (NP-1033)
* Callout: Don't render empty `.cads-callout-label` for standard callouts (no label)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## <sub>v2.1.0-alpha.0</sub>
+
+#### _Nov. 5, 2020_
+
 **Deprecations**:
 
 * Removed standalone `logo_clickable` component. Only used within header and footer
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## <sub>v2.1.0-alpha.0</sub>
```
baseonmars, line 1: Does the release tooling manage re-titling when we flip over to non-alpha?

baseonmars (approved):
Good stuff!

CHANGELOG.md
davidrapson: No, have to handle that manually. Might almost be better to not do the titling for alpha releases.

PR #466, opened 2020-11-05
feature: Design System Test Automation for Input [NP-916]

celonicolosi:
Adding a few scenarios to cover Input and its variants.

- Basic Input
- Input with hint
- Optional Input
- and Input with and error message


testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Input
+  As a non-technical person at Citizens Advice
```
luke-hill, line 2: Lets alter this to be more specificatory to the component proper (We would usually do this style for the regular features on public-website-cucumber).

Take a look at the other components for examples.

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+  Rule: Basic
+    Background:
+      Given I have a Basic Input component on page
+
+    Scenario: Input is on the page and I can type on it
+      Then there is a label
+      But there is no hint
+      And there is no optional field
+      And there is no error message
+      Then I can type "Hello" into the text box
```
luke-hill, line 16: lets re-order these slightly so it reads better. If we place this one first, then the other four in the same order, I think it will read better

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+  As a non-technical person at Citizens Advice
+  I want to know the full specification of what an “Input” component can / should do
+  So I can check whether it is being used correctly
+  And I can check if is buggy
+
+  Rule: Basic
+    Background:
+      Given I have a Basic Input component on page
+
+    Scenario: Input is on the page and I can type on it
```
luke-hill, line 11: Lets re-word this to something like.

Input form is on the page without any assistance. We can also re-word the other ones accordingly.

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+      Then there is a label
+      But there is no hint
+      And there is no optional field
+      And there is no error message
+      Then I can type "Hello" into the text box
+
+  Rule: With Hint
+    Background:
+      Given I have an Input with hint component on page
```
luke-hill, line 20: Whilst the name space for this will be `Input::Hint` lets make the text read a bit better. So we can say. An Input component with a hint

testing/features/support/components/input/basic.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Input
+  class Basic < ::Base
+    set_url "/iframe.html?id=forms-input--basic&viewMode=story"
+
+    element :input, ".cads-input"
+    element :label, "label"
+    element :hint, ".cads-form-field__hint"
+    element :optional, ".cads-form-field__optional"
```
luke-hill, line 10: lets re-name this to `optional_label` so it's a bit more descriptive.

testing/features/support/components/input/basic.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Input
+  class Basic < ::Base
+    set_url "/iframe.html?id=forms-input--basic&viewMode=story"
+
+    element :input, ".cads-input"
+    element :label, "label"
```
luke-hill, line 8: lets call this `title_label` so it's a bit more descriptive.

testing/features/step_definitions/input_steps.rb
```diff
@@ -0,0 +1,53 @@
...
+Then("there is an error message") do
+  expect(@component).to have_error_message
+end
+
+Then("there is no error message") do
+  expect(@component).not_to have_error_message
+end
+
+Then("I can type {string} into the text box") do |string|
+  @component.input.send_keys(string)
```
luke-hill, line 52: This has no assertion. I think you can actually re-use the existing test from the other step file. Perhaps moving it out into a `common_steps.rb` file?

luke-hill (changes requested):
A lot of the review points are just naming. For the main body description in gherkin try encapsulate what the specification of the component should be.

testing/features/input.feature
celonicolosi: I reorganized all the steps, so **Then** came first and **But** came last.

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  The Input component represents a field for text input, the control associated
+  with this field is a text box that allows users to edit a single line of plain
+  text, providing means to submit data
+
+  Rule: Basic
+    Background:
+      Given I have a Basic Input component on page
+
+    Scenario: Input is on the page and I can type on it
```
luke-hill, line 11: `Scenario: Input does not require any additional information`

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+      But there is no hint
+      And there is no optional field
+      And there is no error message
+      Then I can type "Hello" into the text box
+
+  Rule: With Hint
+    Background:
+      Given I have an Input with hint component on page
+
+    Scenario: Input is on the page and has a hint
```
luke-hill, line 22: `Scenario: Input can have a hint`

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+      And there is a hint
+      And there is no optional field
+      But there is no error message
+      Then I can type "Hello" into the text box
+
+  Rule: With Optional field
+    Background:
+      Given I have an Optional Input component on page
+
+    Scenario: Input is on the page and it is optional
```
luke-hill, line 33: `Scenario: Input can have an optional label`

testing/features/input.feature
```diff
@@ -0,0 +1,49 @@
...
+      And there is a hint
+      And there is an optional field
+      But there is no error message
+      Then I can type "Hello" into the text box
+
+  Rule: With Error
+    Background:
+      Given I have an Input with error component on page
+
+    Scenario: Input is on the page but there is an error message
```
luke-hill, line 44: `Scenario: An Input can show an error message`

luke-hill (commented):
Couple of renames to do. Also some of the technical changes don't appear to be pushed up.

testing/features/step_definitions/input_steps.rb
celonicolosi: There is only one step that uses send_keys in other steps files. And it is:
    **When** I type "Immigration" into the search box

Is the idea here to create a common_steps.rb with a step that can be used in both tests?

luke-hill (approved):
Made final change for Test assertion. Now RtM

PR #467, opened 2020-11-06
Hotfix: Safari CI test

luke-hill:
Safari has less unusable space so it appears at a relative lower space

Safari windows have different "usable" spaces. So we need to account for that in our sticky test.


davidrapson (approved):
These numbers feel a bit ✨ magic ✨ but OK 👍

luke-hill:
Fair point. So we scroll to the bottom of a 1280 window. And they appear at the top.

PR #468, opened 2020-11-06
Hotfix/failing ci

luke-hill:
This fixes 2 additional issues that have now surfaced (SHAs)

- The Jenkinsfile was misconfigured. We're using ios drivers which are "similar" but different to safaridrivers
- The breadcrumb tests needs a re-think for iPhones. So we're going to fix it for them!


davidrapson (approved):
👍

PR #470, opened 2020-11-06
scss cleanup

davidrapson:
- Extract cads-spacers into storybook styles
- Remove cads-spacing-0
- Remove unused cads-interactive-target-size__small
- Move forms styles to components
- Remove old cads-iconfont partial (moved to generics)


scss/6-components/_header.scss
```diff
@@ -1,5 +1,5 @@
 // ============================================================================
-// Footer
+// Header
```
baseonmars, line 2: :)

styleguide/styles.scss
```diff
@@ -62,6 +62,19 @@
   }
 }
 
+$cads-spacers: (
+  (
+    0: 0,
```
baseonmars, line 67: `$cads-spacing-0` isn't used by any macros etc?

baseonmars (approved):
Looks fine - I'm gonna assume you're confident in removing cads-spacing-0, if not I don't see the harm in leaving it it.

styleguide/styles.scss
davidrapson: It's never used. Plus having a variable name for 0 is a bit much. The Mayans already sorted this for us 🙃

baseonmars: I figured it'd be handy in a loop, but if it's not used and brings us no joy. then 🧹. What year is the world meant to end again? ;)

PR #471, opened 2020-11-06
Update local backstop version

davidrapson:
Local backstop versions were out of date with the package version. Also adds a dependabot config to keep this up to date in future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,12 @@ updates:
     labels:
       - 'npm dependencies'
     target-branch: 'master'
+  - package-ecosystem: npm
+    directory: '/testing/'
+    schedule:
+      interval: daily
+      timezone: Europe/London
+    open-pull-requests-limit: 10
+    labels:
+      - 'npm dependencies'
+    target-branch: 'master'
--- a/bin/docker/rebuild_vr_references
+++ b/bin/docker/rebuild_vr_references
@@ -5,7 +5,4 @@ set -e
 # start styleguide in background
 docker-compose up -d ca-styleguide
 
-# Get version string from npm inside a context with npm accessible
-version_string=`docker-compose run ca-styleguide npm view backstopjs version`
-# Use the version string and strip the additional \r that docker-compose output adds
-BACKSTOPJS_VERSION=${version_string//[[:cntrl:]]/} docker-compose run visual-tests reference --config=backstop-config-ci.js
+docker-compose run visual-tests reference --config=backstop-config-ci.js
--- a/bin/docker/visual_regression
+++ b/bin/docker/visual_regression
@@ -6,7 +6,4 @@ set -e
 docker-compose up -d ca-styleguide
 
 # run visual regression tests
-# Get version string from npm inside a context with npm accessible
-version_string=`docker-compose run ca-styleguide npm view backstopjs version`
-# Use the version string and strip the additional \r that docker-compose output adds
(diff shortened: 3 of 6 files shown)


luke-hill:
Only thing to bare in mind with this is that @simonwgill looked into this recently, something about dockerfiles. Not sure if same thing.

baseonmars (commented):
Any idea what the breaking change was between 4 and 5? Their release notes are pretty bare.

davidrapson:
FWIW the docker files are already running v5

davidrapson:
Ideally though I'd get rid of this and only use the docker image as that's always how we run it

simonwgill:
It's in `bin/docker/visual_regression` and `bin/docker/rebuild_vr_references` - those scripts will select the backstopjs version referenced by the root package.json for that run. If the environment variable isn't set, it'll just use latest instead.

davidrapson:
Interesting the _root_ package.json doesn't reference backstop at all. Can we get rid of this nested version and pin the docker version to v5? Or latest?

simonwgill:
> Interesting the _root_ package.json doesn't reference backstop at all. Can we get rid of this nested version and pin the docker version to v5? Or latest?

So, what I didn't realise was happening was that `npm view $package version` gets the version from the global index. Technically it will always be the latest version getting downloaded by docker-compose then.

I wouldn't pin it to latest (despite my having accidentally done so) because there is a chance of tests just randomly starting to fail for no good reason.

davidrapson:
Ah yeah I missed that. `npm view` looks up a package from the registry. Does tagging it to the latest released version in one place work e.g. `backstopjs/backstopjs:5.0.6`?

simonwgill:
Yeah, if you take out the BACKSTOPJS_VERSION stuff which is completely useless, you can just use the version pin in the docker-compose file. Unless there's a sensible way of getting the version from the lock file?

Pinning in docker-compose would mean that we have to make a config change for every patch version update we wanted unfortunately. There's no major or major.minor tags on the backstopjs image repo.

davidrapson:
> Pinning in docker-compose would mean that we have to make a config change for every patch version update we wanted unfortunately. There's no major or major.minor tags on the backstopjs image repo.

I think that is fine for this case as backstop release cadence is fairly slow. Plus if we get dependabot updating the package file at least that's a signal when it updates (unless there is a way for dependabot to update Docker file too?)

davidrapson:
Updated this now

simonwgill:
> > Pinning in docker-compose would mean that we have to make a config change for every patch version update we wanted unfortunately. There's no major or major.minor tags on the backstopjs image repo.
> 
> I think that is fine for this case as backstop release cadence is fairly slow. Plus if we get dependabot updating the package file at least that's a signal when it updates (unless there is a way for dependabot to update Docker file too?)

Not sure I understand your meaning of "plus" here, David. If we don't reference backstopjs in the package.json then I can't see how dependabot will fire for it? If there's a way to get it to update docker tags in a docker-compose file though, that would work out well.

davidrapson:
We do, see testing/package.json and the new dependabot config I added.

simonwgill:
Ahhh, gotcha.

So... this command works correctly to determine the backstopjs out of the package-lock in /testing/ `npm list backstopjs --depth=0 2>/dev/null | grep -i "backstopjs@" | grep -oEi "\d\.\d\.\d"`. It'll work whether or not the module is installed as well.

davidrapson:
Going to come back to this as we need to review with Simons change and this isn't essential at the mo

PR #476, opened 2020-11-09
feature: Feedback link and footer links are now supplied by template locals

baseonmars:
This PR is in preparation for advisernet. The feedback link at the top of the footer and the 4 sections of links are now supplied via template locals.

Note that the colour change of the footer is made by varying SCSS variables in the public-website, not in the design system.


styleguide/component.stories.js
```diff
@@ -9,7 +9,6 @@ import initTables from '../src/js/tables';
 import GreedyNav from '../src/ts/greedy-nav';
 
 // eslint-disable-next-line no-unused-vars
-import locals from './haml_locals.rb'; // just used to watch
```
baseonmars, line 12: Watching this file has caused intermittent build errors and after adding the footer links I found it 100% of the time caused build errors.

Will need to find a better way to watch this, and other dependant files (it is possible to extend the loader to do this).

styleguide/haml_locals.rb
```diff
@@ -182,6 +182,62 @@
     HTML
   },
 
+  "footer" => {
```
davidrapson, line 185: Any way we can do this without the locals? Nice if we can reduce dependance on this file over time

davidrapson (approved):
👍 LGTM. Named column slots make sense I think.

styleguide/haml_locals.rb
baseonmars: There's a couple of ways and the pain on working on this feature is making them seem like something I might explore sooner rather than later. In the end I decided to stick with using this until the Header is finished then dedicate some time to improving the dev experience.

PR #479, opened 2020-11-09
Pagination review

davidrapson:
- Migrates pagination and paging info to new docs format. Includes a little experiment with a hybrid docs format to allow us to remove some of the boilerplate _out_ of the markdown code.  Using this approach for embedding stories https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/mdx.md#embedding-stories
- Update pagination to match designs: border-radius and uses button primary colours for current state

Screenshot 2020-11-09 at 16 48 00

https://citizensadvice.atlassian.net/browse/NP-1031


styleguide/components/pagination/_basic.html
```diff
@@ -0,0 +1,14 @@
+<ul class="cads-paging">
```
baseonmars, line 1: I think we should wrap pagination in `nav` elements with `role="navigation"`. Thinking about that also prompted me to look up best practices around pagination accessibility and there's also recommendations to use `aria-current="page"` and a label such as `aria-label="Page three, current page"` on the active element and `aria-label="Pagination Navigation"` on the nav element.

What do you think about improving the documentation to include accessibility features?

baseonmars (changes requested):
Looking good, but perhaps we can also add accessibility recommendations?

davidrapson:
Have updated pagination with appropriate aria roles and labels, included documentation to explain how they are used:

image

baseonmars (approved):

PR #482, opened 2020-11-10
feature: Add firefox to Local grid test runs [NP-1174]

luke-hill:
Same as sister PR for public-website-cucumber

NB: I added parallel functions for the grid tests. I will add parallelisation for the browserstack ones in the next sprint too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,11 +19,11 @@ def ecr_credential = 'ecr:eu-west-1:cita-devops'
 
 def images = [:]
 
-def withCucumberNode(String description, Array imageArray, Closure body) {
+def withCucumberNode(String description, Map imageMap, Closure body) {
     node("docker && awsaccess") {
         stage (description) {
             checkout scm
-            withDockerSandbox(imageArray) {
+            withDockerSandbox(imageMap) {
                 // Call closure
                 body()
             } // withDockerSandbox
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -106,26 +106,67 @@ pipeline {
                 }
             }
         }
-        stage('Sanity Test') {
+        stage('VRT / Ally Tests') {
             steps {
-                script { env.BUILD_STAGE = 'Sanity Test' }
+                script { env.BUILD_STAGE = 'VRT / Ally Tests' }
                 withDockerSandbox([
                     images['ca-styleguide'],
                     'backstopjs/backstopjs',
                     images['ruby'],
-                    images['wcag'],
-                    'selenium/hub:4.0.0-alpha-6-20200609',
-                    'selenium/node-chrome:4.0.0-alpha-6-20200609',
-                    'selenium/node-firefox:4.0.0-alpha-6-20200609' ]) {
+                    images['wcag'] ]) {
                     script {
                         try {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -106,22 +106,18 @@ pipeline {
...
         }
-        stage('Sanity Test') {
+        stage('VRT / Ally Tests') {
             steps {
-                script { env.BUILD_STAGE = 'Sanity Test' }
+                script { env.BUILD_STAGE = 'VRT / Ally Tests' }
                 withDockerSandbox([
                     images['ca-styleguide'],
                     'backstopjs/backstopjs',
                     images['ruby'],
```
luke-hill, line 115: wasn't sure if this was needed for these two tests. So left it in.

Jenkinsfile
```diff
@@ -106,22 +106,18 @@ pipeline {
                 }
             }
         }
-        stage('Sanity Test') {
+        stage('VRT / Ally Tests') {
             steps {
-                script { env.BUILD_STAGE = 'Sanity Test' }
+                script { env.BUILD_STAGE = 'VRT / Ally Tests' }
```
davidrapson, line 111: Use A11y here or the full Accessibility, rather than Ally.

luke-hill: Should they be grouped together. Or split out into their own stages? Whilst I'm here I can tackle it.

davidrapson: Granular steps makes sense to me. Don't need to be grouped together

luke-hill:
@simonwgill We get a name conflict on container spin up here: https://jenkins.devops.citizensadvice.org.uk/blue/organizations/jenkins/design-system/detail/PR-482/2/pipeline/144

Thoughts?

simonwgill:
> @simonwgill We get a name conflict on container spin up here: https://jenkins.devops.citizensadvice.org.uk/blue/organizations/jenkins/design-system/detail/PR-482/2/pipeline/144
> 
> Thoughts?

Looks like they're both trying to run on the same agent. There was a thing for that... Pretty sure I was using the label to force it. Which repo did we do again so I can go back and check?

luke-hill:
https://github.com/citizensadvice/public-website-cucumber/blob/master/Jenkinsfile

simonwgill:
Yeah, @luke-hill, using something similar to the withCucumberNode function in public-website-cucumber will definitely help. Obviously you won't need the same crazy env layering in this repo.

Jenkinsfile
```diff
@@ -118,17 +118,34 @@ pipeline {
...
-        stage('VRT / Ally Tests') {
+        stage('Visual Regression Tests') {
             steps {
-                script { env.BUILD_STAGE = 'VRT / Ally Tests' }
+                script { env.BUILD_STAGE = 'Visual Regression Tests' }
                 withDockerSandbox([
                     images['ca-styleguide'],
                     'backstopjs/backstopjs',
-                    images['ruby'],
                     images['wcag'] ]) {
```
davidrapson, line 127: This can go too

luke-hill:
This fails to find the defined method. I have no idea why :( :shrug:

luke-hill:
Being dealt with in 491

PR #483, opened 2020-11-10
v2.1.0 alpha.1

davidrapson:
**Bugfixes**:

* Search: Normalise search button styles and use 44px height. Removed `cads-button` from search input.
* Pagination: Update pagiantion designs and add reference aria examples.
* Forms: Add `name` attribute to input haml partial

**New features**:

* ⬆️ Footer: Supply feedback link and links using template locals (NP-662)


baseonmars (approved):
good stuff!

PR #484, opened 2020-11-10
[bugfix] Updating input component to include name attribute [SMP-101]

tonyto:
Using the input component in a rails form and realising the input values are not submitting with the form because the `name` attribute is not set.

This PR set the attribute `name: input['name']` on the Input component.

This change does not change the UI or the way it functions.

### DIFF
--- a/haml/_input.html.haml
+++ b/haml/_input.html.haml
@@ -9,4 +9,4 @@
       %p.cads-form-field__hint= input["hint"]
     - if input["error_message"]
       %p.cads-form-field__error-message= input["error_message"]
-    %input.cads-input{ type: "text", id: "#{input['name']}-input", class: ("cads-input--#{input['size']}" if input["size"]) }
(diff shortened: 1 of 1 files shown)


davidrapson:
Not sure if you can see the build but failing for :neckbeard: reasons

> haml/_input.html.haml:12 [W] RuboCop: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping. (https://rubystyle.guide#consistent-string-literals)

You can run `bundle exec haml-lint` or I think just `rake` to see this locally.

davidrapson (approved):
👍

simonwgill:
> You can run `bundle exec haml-lint` or I think just `rake` to see this locally.

Yep, `bundle exec rake` will run all the checks that can be done just with the source code (linting, unit tests, etc)

PR #489, opened 2020-11-11
Fix error message display on inline radio buttons and fixed width inputs [NP-1275]

davidrapson:
Fix error message display on inline radio buttons and fixed width inputs

| Before | After |
| ----- | ---- |
|  Screenshot 2020-11-11 at 10 12 48 | Screenshot 2020-11-11 at 10 13 13 |

| Before | After |
| ----- | ---- |
| Screenshot 2020-11-11 at 10 16 54 | Screenshot 2020-11-11 at 10 21 37 |

NP-1275


tonyto (approved):
Noice 🚀

davidsauntson (approved):
🚀

PR #491, opened 2020-11-11
jenkins scripted pipeline

zhelyan:
rewrite of https://github.com/citizensadvice/design-system/pull/482 which uses scripted pipeline. Mixing declarative with scripted code in the Jenkinsfile is not a good idea.

Feel free to ignore this PR or use bits and bobs from it as needed.

Copy / pasting my slack message to @luke-hill here:

> I converted to scripted because mixing scripted / declarative is not recommended.
> 13:13
> couple of things
> 13:13
> archiving artefacts is commented out as I didn’t know which stage it applies to.
> 13:15
> the code can be cleaned up quite a lot more to abstract things away
> 13:16
> not sure if stash 'sources' stashes the .git folder. slackNotifyReleaseOnly relies on it
> 13:20
> and lastly - not sure if the different stages need the image you build in Setup. If they do - you need to
> 13:20
> docker.withRegistry(docker_registry_url, ecr_credential) { docker.image(........).pull()
> 13:21
> if the stage runs on different node. Otherwise docker-compose will rebuild it

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,8 +32,8 @@ def images = [:]
 node(nodeLabel) {
   stage('Setup') {
     cleanWs()
-    //checkout scm
-    git credentialsId: 'jenkins-pipeline-gh-access', url: 'https://github.com/citizensadvice/design-system.git'
+    checkout scm
+
     slackNotifyReleaseOnly {
       stash 'source'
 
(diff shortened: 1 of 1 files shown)


luke-hill (commented):
Went through all of this with Simon. We're going to make a lot of tabbing changes, and a couple of minor tweaks. But other than that this is good to go.

Thanks @zhelyan !

simonwgill:
This was finished off in https://github.com/citizensadvice/design-system/pull/531 which fixes some issues that were a problem before adding parallelisation as well.

PR #492, opened 2020-11-11
feat: add variables to allow footer to be styled [NP-662]

baseonmars:
This PR adds variables to the footer to allow it to be styled correctly for AdviserNet.

**The logo is wrong in this screenshot, but here's an example of what's possible**

Screenshot 2020-11-11 at 15 46 05


davidrapson (approved):
🥇 Nice. Lovely work.

PR #494, opened 2020-11-12
bugfix: JS Lag fix [NP-1264]

luke-hill:
**Previously:**

- We had 3 problematic tests in OISC Warnings for Mobiles
- Two of these we believed related to iOS12 not rendering the freeze-pane correctly
- One of these we believed related to Mobiles not rendering a native property of Selenium (The "rect"), correctly.

**Now:**

- We have 2 problematic tests in OISC Warnings for Mobiles
- Both of these relate to Mobiles not rendering a native property of Selenium (The "rect"), correctly.
- One of the 2 iOS12 failures was actually remedied by just waiting longer.

**The Cause:**

- The XCUITest (Appium), JS Driver when executing JS script isn't halting during execution. During the Regular driver usage, it either "appears" to, or is so quick, that it is impossible to notice. Consequently our `0.5` second delay after moving the page was more than enough to cover against "good" browsers.
- When using "bad" or "slow" browsers (Read iOS Safari or Safari proper), this halting is visible/discernable and as such we need to manually wait for it for a short(er), while longer.

**Next**
- Watch Appium repo for fix / updates. https://github.com/appium/ruby_lib_core/issues/285 is the last place I raised it.


pedantic-git (approved):
This all looks good to me.

It's not one for this time but I like to apply the mantra "if you need to sleep in your Capybara tests, you've not got enough stuff in your DOM to test for" so we might like to revisit all this sleeping at a future date.

luke-hill:
I almost wholeheartedly agree @pedantic-git - The thing with this (At least from my knowledge), is that when given a command, selenium has it's own waiting logic.

Because the command in question doesn't navigate or click on anything, but just scrolls, it's a little hard to wait for something in the DOM. Although... We could potentially wait for the close button to show (This would indicate it has stickified), the only issue is on phones this then plays silly beggars.

I'll make a mental note to have a re-think in a few weeks for this.

pedantic-git:
Yeah - scrolling is a very tricky one I know.

PR #496, opened 2020-11-12
feature: allow search button to be styled

baseonmars:
One more change to allow the search button to be styled.

Screenshot 2020-11-12 at 17 11 23

### DIFF
--- a/scss/6-components/_search.scss
+++ b/scss/6-components/_search.scss
@@ -1,5 +1,5 @@
 // ============================================================================
-// Footer
+// Search
 // ============================================================================
 
 /** @define search */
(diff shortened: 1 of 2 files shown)


scss/6-components/_search.scss
```diff
@@ -1,5 +1,5 @@
 // ============================================================================
-// Footer
+// Search
```
davidrapson, line 2: 🙈

davidrapson (approved):
👍

scss/6-components/_search.scss
baseonmars: np. I have done far worse!

PR #498, opened 2020-11-13
Bugfix/more friendly waiters

luke-hill:
Given we were using hard coded sleeps. This "inevitably" failed when the time wasn't long enough in a single situation (The one in which we stipulated the longest 2.5 second sleep).

Now we revert back to a 0.5 second sleep. But add an additional conditional 5 second sleep based on the element appearing in time.

This will have no effect on the fast browsers (They still have a 0.5s delay), but on the slower ones it will wait up to 5.5s +/- a tolerance. If they don't pass here, then I'm not sure what we can do. Probably need to tag them and go back to the drawing board!

:crossed_fingers:

### DIFF
--- a/testing/features/step_definitions/oisc_warnings_steps.rb
+++ b/testing/features/step_definitions/oisc_warnings_steps.rb
@@ -24,7 +24,7 @@ When("I scroll to the top of the page") do
 end
 
 When("I close the sticky component") do
-  @component.wait_until_close_sticky_visible(wait: 5) if ios? || safari?
+  @component.wait_until_close_sticky_visible(wait: 5)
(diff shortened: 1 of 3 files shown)


testing/features/step_definitions/oisc_warnings_steps.rb
```diff
@@ -24,6 +24,7 @@
 end
 
 When("I close the sticky component") do
+  @component.wait_until_close_sticky_visible(wait: 5)
```
pedantic-git, line 27: Would you mind linking me to where this method is defined?

pedantic-git (approved):

PR #502, opened 2020-11-16
bugfix: Sticky issues (Omit testing on all problematic versions) [NP-1297]

luke-hill:
TL;DR - Fixed the issue by not testing the scenarios on iOS11/12

Essentially the code will never work on iOS11/12 because each of those devices / OS' has differing behaviour for the scrollbar appearing "sticky" and the close button appearing.

`safaridriver` version 13 does start to mitigate this issue, and all tested devices on iOS13 seem fine (For scrolling and showing the close button). I suggest given we have a working staple, we just wait until we stop supporting iOS11/12

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,7 +17,7 @@ namespace :design_system do
   task :all do
     puts "Running all design system tests"
     system(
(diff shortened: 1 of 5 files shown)


davidrapson (approved):
👍

PR #503, opened 2020-11-16
feature: Page Review Spec [NP-1183]

luke-hill:
This is a new spec, but a really simple one - As the component is just plain text.

As part of this I refactored all "old" style components to validate the initial form, but the base validation is just the root element is present (And again this is overloaded with most new components).

I did this, as for Page Review there is literally nothing we can wait on we don't already validate on. Because it's such a threadbare component.

NB: I didn't validate date format, as I wasn't sure if we wanted to do this. cc/ @sheldonsmiltnieks Maybe one for you to rule on?

### DIFF
--- a/testing/features/page_review.feature
+++ b/testing/features/page_review.feature
@@ -1,0 +1,12 @@
+Feature: Page Review component
+
+  The Page Review component reassures the public that the content
+  they are viewing is up to date
+
+  Rule: Standard Page Review text
+    Background:
+      Given a Default Page Review component is on the page
+
+    Scenario: Label includes the date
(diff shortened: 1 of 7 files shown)


davidrapson (approved):
👍 

> NB: I didn't validate date format, as I wasn't sure if we wanted to do this. cc/ @sheldonsmiltnieks Maybe one for you to rule on?

On this I'd say _no_ here, as we don't do the date formatting at this end. We simply pass through what's given. The date is formatted on the consumer end, and we have unit tests for that part in public website so I reckon we're good.

PR #504, opened 2020-11-16
Move OISC out of the design system [NP-1285]

davidrapson:
Pairs with https://github.com/citizensadvice/public-website/pull/531

@luke-hill Apologies for the churn, let me know if anything from this needs moving into the public-website tests or if the existing coverage there is enough.


luke-hill (approved):
At the moment we have 2 * 2 OISC tests on Immigration (One sticky on non-sticky), these are the only 2 I am aware of.

Can you confirm these are the only 2 we have atm.

I'm guessing moving this out of DS is because these are going to be more bespoke each time?

Also on code removal SwV0uhep6Uqr5EYZkg

davidrapson:
> Can you confirm these are the only 2 we have atm.

Yeah that covers the main cases from an interaction point of view

> I'm guessing moving this out of DS is because these are going to be more bespoke each time?

Specifically, it was unanimously agreed in the last design-system working group on the basis that we don't _want_ to encourage these to be re-used as the have a specific legal context.

davidsauntson (approved):
💣

PR #505, opened 2020-11-16
Correctly guard for when header is not present

davidrapson:
Spotted in https://github.com/citizensadvice/public-website/pull/531

Correctly guard for when header is not present.


davidsauntson (approved):
👍

PR #506, opened 2020-11-16
Update search partial to accept input_value

davidrapson:
The search partial is currently not used directly by any Rails app (there is a version in the header but that is a separate partial). With https://github.com/citizensadvice/public-website/pull/516 we can start using it. To do that we need to be able to set the current input value. Whilst making this change I also made the search parameter (q) configurable.

image


styleguide/components/search/_search.html.haml
```diff
@@ -1 +1 @@
-= render partial: "@citizensadvice/search", locals: {search_action_url: '/search/page' }
+= render partial: "@citizensadvice/design-system/haml/search", locals: { search: { action_url: "/search/page", search_param: "q" } }
```
davidsauntson, line 1: If we allow a consumer of the design system to specify the name of the search param, should we should do the same with the search box in the header?  Otherwise they could potentially have to deal with catering for two names for the same param.

In reality, I think most consumers will just be pointing back to the search page on the public advice site, so is there any advantage in making this configurable?

davidsauntson (approved):
nice!

question about the search param

styleguide/components/search/_search.html.haml
davidrapson: Good point, I agree. I think I'll remove that change from this PR as it's tangential to what I'm trying to fix here anyway.

PR #509, opened 2020-11-17
[bugfix][SMT-109] Add value prop to Input partial

tonyto:
In the event of an error, I'd also like to populate the value in the input

image


davidrapson (changes requested):
Thanks for doing this! We're not actively using these partials yet so apologies if there's a few of these tweaks to make.

Couple of lint-errors:

```
haml/_input.html.haml:12 [W] RuboCop: Style/IfUnlessModifier: Modifier form of `if` makes the line too long. (https://rubystyle.guide#if-as-a-modifier)
haml/_input.html.haml:12 [W] RuboCop: Layout/LineLength: Line is too long. [154/140] (https://rubystyle.guide#max-line-length)
```

Given how finicky haml is about line-lengths and wrapping we could probably disable these in haml-lint.yml  unless they look easy to fix.

You can see these locally by running `rake` or more directly `bundle exec haml-lint`

Also mind updating the CHANGELOG?

tonyto:
Thanks @davidrapson.

Multiline haml is plain ugly IMO. I've excluded the file for now.
Also added to the CHANGELOG.

davidrapson (approved):

PR #510, opened 2020-11-17
feature: add skip links [NP-1181]

davidsauntson:
Adds three skip links to the header that become visible when they receive focus.

- added markup and styles to present links
- added welsh translations from current site
- updated header documentation
- updated samples pages
- updated changelog

### DIFF
--- a/haml/_header.html.haml
+++ b/haml/_header.html.haml
@@ -6,14 +6,13 @@
 %header.cads-header
   .cads-grid-container
     .cads-grid-row
-      .cads-grid-col-12
-        %nav.cads-skip-links
-          %a.cads-skip-links__link{ href: "#cads-navigation" }
-            = t("cads.header.skip_to_nav")
-          %a.cads-skip-links__link{ href: main_content_anchor_link }
-            = t("cads.header.skip_to_content")
-          %a.cads-skip-links__link{ href: "#cads-footer" }
-            = t("cads.header.skip_to_footer")
+      %nav.cads-skip-links
+        %a.cads-skip-links__link{ href: "#cads-navigation" }
+          = t("cads.header.skip_to_nav")
+        %a.cads-skip-links__link{ href: main_content_anchor_link }
+          = t("cads.header.skip_to_content")
+        %a.cads-skip-links__link{ href: "#cads-footer" }
+          = t("cads.header.skip_to_footer")
       .cads-grid-col-md-6.cads-header__logo-row
         %a.cads-logo{ href: root_path, title: t("cads.shared.logo_title") }
         %button.cads-header__search-reveal.js-cads-search-reveal.cads-icon_search{ title: t("cads.header.search_mobile"), "aria-expanded": "false" }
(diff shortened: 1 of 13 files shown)


haml/_header.html.haml
```diff
@@ -1,9 +1,18 @@
...
-- sign_in_out ||= {}
+:ruby
+  links = header.fetch("links", [])
+  sign_in_out = header.fetch("sign_in_out", {})
+  main_content_anchor_link = header.fetch("main_content_id", "#cads-main-content")
 
 %header.cads-header
   .cads-grid-container
     .cads-grid-row
+      %nav.cads-skip-links
```
davidrapson, line 9: Wondering if this should be completely outside the `header` element? i.e. the first thing on the page.

haml/_header.html.haml
```diff
@@ -1,9 +1,18 @@
-- links ||= []
-- sign_in_out ||= {}
+:ruby
```
davidrapson, line 1: How come you are using the `:ruby` filter rather than sticking with `-` for each line?

scss/6-components/_skip-links.scss
```diff
@@ -0,0 +1,16 @@
+.cads-skip-links {
```
davidrapson, line 1: Mind adding `/** define skip-links */` to this so it triggers the BEM lint rule? Although this looks good so don't expect any warnings.

styleguide/components/header/header.stories.mdx
```diff
@@ -30,21 +30,28 @@ import linksVariable from '!!raw-loader!./links_example.rb';
 
 The Haml partial the following locals:
```
davidrapson, line 31: Worth updating this to mention that these locals are now within a `header` hash?

davidrapson (changes requested):
👍 Nice couple queries and minor tweaks, but this looks great.

haml/_header.html.haml
davidsauntson: Does it make a difference? It's still the first thing to receive focus.

haml/_header.html.haml
davidsauntson: It's a haml-linting rule that error if there are three or more

haml/_header.html.haml
davidrapson: Probably not, it only looked funny to see something loose within a `cads-grid-row`, or do we still want it to sit within the grid? If so then crack on!

styleguide/components/header/header.stories.mdx
davidsauntson: have reverted to previous form as discussed

haml/_header.html.haml
davidsauntson: Yeah it needs to be in the grid 👍

scss/6-components/_skip-links.scss
davidrapson: Sorry! I just realised I typoed this 🙈 , needs an `@` in `@define`

davidrapson (approved):
👍

soniaturcotte (approved):
⭐️⭐️⭐️⭐️⭐️

PR #516, opened 2020-11-18
Accessible targeted content [NP-1093]

davidrapson:
Refactors the targeted content component so that it:

- Is usable if javascript fails or is disabled
- No longer uses details/summary which is unfortunately not accessible
- Exposes heading levels correctly
- Has descriptive messaging for expanded and closed states for screen readers

Thanks to @soniaturcotte for pointing me in the direction of https://github.com/alphagov/govuk-design-system/issues/1379, which referenced https://components.publishing.service.gov.uk/component-guide/step_by_step_nav/default/preview as a reasonable starting point. Also referenced the techniques in https://inclusive-components.design/collapsible-sections/ and https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

Needs plenty of testing, but is hopefully is a good starting point.

## VoiceOver behaviour

voiceover

## Fallback behaviour

A major challenge with testing fallback behaviour in Storybook is that Storybook is one giant client-side app so _doesn't work without JavaScript_. To counter this I've included a variant of the story that doesn't have the javascript code initialised.

Screenshot 2020-11-18 at 10 17 25


locales/en.yml
```diff
@@ -22,7 +22,9 @@ en:
       submit_title: Submit search query
       submit_label: Search
     targeted_content:
-      close_button_aria: Collapse %{title}
+      descriptive_label_show: show this section
```
davidrapson, line 25: Need to decide on the appropriate wording here, i.e. show/hide, close/open, or expand/collapse.

davidrapson:
Component in IE 11:

image

locales/en.yml
davidsauntson: I like show and hide, the GDS stuff mentioned show/hide being more easily understood as it actually describes what is happening, the others are a bit more figurative

davidsauntson (approved):
great! 👍

locales/en.yml
davidrapson: Cool, I went for that for the same reason. I guess my follow up question is if we're happy on sticking with "Close" for the button at the bottom of the UI. i.e. the visual label would read "Close" but the longer form aria text would say hide this section.

davidsauntson:
Oh just had a thought that I didn't check 😬  - does this still work with the 'expand all' button that is at the top of each page? And should we (royal) make a story that show this functionality? This doesn't have to be part of this PR

davidrapson:
> Oh just had a thought that I didn't check 😬 - does this still work with the 'expand all' button that is at the top of each page?

Almost certainly not, but that's currently in public-website (not 100% sure it should be). So yeah, we should possibly do something to show that here, or at least we'll need to update the code in public-website afterwards.

PR #523, opened 2020-11-19
Upgrade storybook to 6.1

davidrapson:
Storybook 6.1 is out. Mostly performance related improvements but also includes a slight UI refresh.

Screenshot 2020-11-19 at 09 15 37

https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md


davidrapson:
Interesting, the clean/build step seems to fail

PR #531, opened 2020-11-20
refactor: Convert to scripted pipeline [NP-1174]

simonwgill:
Adding extra grid tests that would work in an acceptably fast fashion meant that we needed parallelisation in this Jenkinsfile. Zhelyan did a basic rewrite converting everything to scripted pipeline so we weren't mixing the two styles anymore.

That left in a lot of bugs and weirdnesses that had accumulated and emphasised a couple of others. This PR takes that code, makes it fit the NP coding style a bit closer, and fixes some of the bugs

- set the tab width to 2 spaces for consistency with other files
- added a try finally block to replicate the post always section of the declarative pipeline
- restructured the artifacts folder structure so information about what tests were run is encoded

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -127,12 +127,12 @@ def pipeline() {
       } finally {
         sh './bin/jenkins/fix_visual_test_report'
         publishHTML([
-            allowMissing: true,
-            alwaysLinkToLastBuild: true,
-            keepAll: true,
-            reportDir: 'reports/html_report',
-            reportFiles: 'index.html',
-            reportName: 'BackstopJS Report',
+          allowMissing: true,
+          alwaysLinkToLastBuild: true,
+          keepAll: true,
+          reportDir: 'reports/html_report',
+          reportFiles: 'index.html',
+          reportName: 'BackstopJS Report',
         ])
       }
     }
@@ -189,15 +189,15 @@ def withTestingNode(String description, Boolean useBrowserStack, Closure body) {
         withEnv(global_environment_variables) {
           if (useBrowserStack) {
             withForcedDockerUpdate([images['ruby']]) {
-                withVaultSecrets([
-                  BROWSERSTACK_USERNAME: '/secret/devops/public-website/develop/env, BROWSERSTACK_USERNAME',
-                  BROWSERSTACK_ACCESS_KEY: '/secret/devops/public-website/develop/env, BROWSERSTACK_ACCESS_KEY',
-                ])
-                {
-                  // Call closure
-                  body()
-                } // withVaultSecrets
-              }
+              withVaultSecrets([
+                BROWSERSTACK_USERNAME: '/secret/devops/public-website/develop/env, BROWSERSTACK_USERNAME',
+                BROWSERSTACK_ACCESS_KEY: '/secret/devops/public-website/develop/env, BROWSERSTACK_ACCESS_KEY',
+              ])
+              {
+                // Call closure
+                body()
+              } // withVaultSecrets
+            }
           } else {
             withForcedDockerUpdate(
               [
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,11 +32,8 @@ def images = [:]
 node(nodeLabel) {
   stage('Setup') {
     cleanWs()
-    //checkout scm
-    git credentialsId: 'jenkins-pipeline-gh-access', url: 'https://github.com/citizensadvice/design-system.git'
+    checkout scm
     slackNotifyReleaseOnly {
-      stash 'source'
-
       withEnv([
         "DOCKER_TAG=${dockerTag()}",
         // Using the commit SHA would mean that every build
@@ -173,7 +170,7 @@ grid_tests = [:]
 grid_tests['regression'] = {
   stage('Full Regression Test') {
     node(nodeLabel) {
-      unstash 'source'
+      checkout scm
       if (isRelease) {
         env.BUILD_STAGE = 'Full Regression Test'
         withVaultSecrets([
@@ -200,7 +197,7 @@ parallel grid_tests
 def withCucumberNode(def description, def imageMap, Closure body) {
   node('docker && awsaccess') {
     stage(description) {
-      unstash 'source'
+      checkout scm
       slackNotifyReleaseOnly {
         withDockerSandbox(imageMap) {
           // Call closure
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,12 +20,7 @@ ecr_credential = 'ecr:eu-west-1:cita-devops'
 
 images = [:]
 
-global_environment_variables = [
-  "DOCKER_TAG=${env.BRANCH_NAME}_${getSha()}",
-  // Using the commit SHA would mean that every build
-  // gets a different tag and this busts the cache between PR builds.
-  "CA_STYLEGUIDE_VERSION_TAG=${env.BRANCH_NAME}"
-]
+global_environment_variables = []
 
 // Pipeline definition begins here
 
@@ -36,9 +31,16 @@ node('docker && awsaccess') {
       notFailBuild: true
     ])
     checkout scm
-    currentBuild.displayName = "$BUILD_NUMBER: $DOCKER_TAG"
+
+    global_environment_variables = [
+      "DOCKER_TAG=${env.BRANCH_NAME}_${getSha()}",
+      // Using the commit SHA would mean that every build
+      // gets a different tag and this busts the cache between PR builds.
+      "CA_STYLEGUIDE_VERSION_TAG=${env.BRANCH_NAME}"
+    ]
 
     withEnv(global_environment_variables) {
+      currentBuild.displayName = "$BUILD_NUMBER: $DOCKER_TAG"
       slackNotifyReleaseOnly() {
         pipeline()
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,6 @@
 deployBranches = ['master']
+isRelease = deployBranches.contains(env.BRANCH_NAME)
+nodeLabel = 'docker && awsaccess'
 
 configurationTypes = [
     ['Windows_10_85', 'chrome'],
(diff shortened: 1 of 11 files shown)


testing/cucumber.yml
```diff
@@ -1,7 +1,9 @@
 passing: --tags 'not @failing and not @future_release'
 
-html: --format html --out artifacts/reports/report.html
-junit: --format junit --out artifacts/reports
-json: --format json --out artifacts/reports/report.json
+<% base_path = "#{ENV['BROWSER']}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'grid')}" %>
```
luke-hill, line 3: this wants to be evaluated <%=

simonwgill: Does it? It's setting a variable which is then used in the profile statements, so it shouldn't be outputting anything on that line

luke-hill: Hmm ok, maybe I'm wrong here. Because it's doing both evaluation and setting. Maybe not then.

Rakefile
```diff
@@ -14,12 +16,25 @@ end
...
     system(
       "cd ./testing && bundle exec cucumber -p reports && cd .."
     ) || raise
   end
+
+  task :report_folders do
+    puts "Creating folder structure for this test run"
+    raise "BROWSER must be set" unless ENV["BROWSER"]
+
+    base_path = "artifacts/#{ENV['BROWSER']}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'grid')}"
```
luke-hill, line 30: can we call the fallback here either "unknown" or "non-browserstack". As locally we run these rakes sometimes.

Rakefile
```diff
@@ -14,12 +16,25 @@ end
 
 namespace :design_system do
   desc "All Design System Tests"
-  task :all do
+  task all: [:report_folders] do
```
luke-hill, line 19: as this isn't enumerable, lets leave it as `all: :reports_folders do` that way it's slightly more optimal.

Rakefile
```diff
@@ -14,12 +16,25 @@ end
...
+    puts "Creating folder structure for this test run"
+    raise "BROWSER must be set" unless ENV["BROWSER"]
+
+    base_path = "artifacts/#{ENV['BROWSER']}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'grid')}"
+    [
+      "#{base_path}/html_pages",
+      "#{base_path}/logs",
+      "#{base_path}/reports",
+      "#{base_path}/screenshots"
+    ].each { |dir| FileUtils.mkdir_p(dir) }
```
luke-hill, line 36: I'm not familiar with this. But does this complain if the directories already exist?

luke-hill (commented):
Seems good, couple of questions / naming queries.

Rakefile
```diff
@@ -14,12 +16,25 @@ end
 
 namespace :design_system do
   desc "All Design System Tests"
-  task :all do
+  task all: [:report_folders] do
     puts "Running all design system tests"
     system(
       "cd ./testing && bundle exec cucumber -p reports && cd .."
```
luke-hill, line 22: This here, the first step. May be needed also in the report folders test. Or alternatively you can add the testing/ prefix

Rakefile
simonwgill: Done

Rakefile
simonwgill: I've added an unknown fallback for browser

Rakefile
simonwgill: Nope, it just makes sure they are there.

Rakefile
simonwgill: Added the testing prefix.

Rakefile
```diff
@@ -12,14 +14,28 @@ def collect_task_errors(tasks)
   errors.count.positive? && raise("The following tasks failed #{errors}")
 end
 
+def base_cucumber_path
+  "artifacts/#{ENV.fetch('BROWSER', 'unknown')}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'grid')}"
```
luke-hill, line 18: Sorry I meant on the second one. Not the first one. The browser is always set. Browserstack config options are not always set.

I think this needs altering a bit as well in a sub-method. Because even if the options are set. They are ignored if the browserstack var isn't set. Maybe something like

```
if browserstack?
  "artifacts/#{ENV.fetch('BROWSER')}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS')}"
elsif grid?
  "artifacts/#{ENV.fetch('BROWSER')}/grid"
else
    "artifacts/#{ENV.fetch('BROWSER')}/local"
end
```

simonwgill: Not sure about the grid? and browserstack? functions being available in cucumber.yml and the Rakefile, but I can use the backing environment variables instead.

Are the BROWSERSTACK_CONFIGURATION_OPTIONS still optional, even if browserstack is enabled? Seems like something that we would always need to specify

luke-hill: the config options are mandatory if on browserstack. but not otherwise.

simonwgill: ok, so if we're $BROWSERSTACK==true, BROWSERSTACK_CONFIGURATION_OPTIONS will be set, so that's a decent enough proxy for browserstack? for use in the folder path.

Is there any reason why we can't just drop reports run by the dev on their desktop into a folder named grid inside the relevant browser folder? They are basically running the grid tests at that point, right?

luke-hill: not quite. I often run locally. Which just tunnels directly to the driver on your machine, and runs your own instance of chrome/firefox

However, bunching up grid tests and local tests doesn't feel too much of a problem. So we could just call them all "non-browserstack" maybe?

simonwgill: Ah, and they are the same tests running against a local driver, in the grid and in browserstack, aren't they?

Fair enough, I guess local might be the better name then.

luke-hill: Kind of. Essentially selenium runs in 2 ways. Local / Remote.

But for us we have 2 "kinds" of remote. We've got a remote we control (our grid, through docker), and a remote we don't control (Browserstack).

simonwgill: We've gone with "other" as the final choice.

Jenkinsfile
```diff
@@ -1,207 +1,301 @@
...
+      try {
+        sh './bin/jenkins/visual_regression'
+      } catch (Exception e) {
+        sh 'docker-compose logs --no-color'
+        currentBuild.result = 'FAILURE'
+        throw e
+      } finally {
+        sh './bin/jenkins/fix_visual_test_report'
+        publishHTML([
+            allowMissing: true,
```
luke-hill, line 130: -1 tab here.

Jenkinsfile
```diff
@@ -1,207 +1,301 @@
...
+
+def withTestingNode(String description, Boolean useBrowserStack, Closure body) {
+  node('docker && awsaccess') {
+    try {
+      stage(description) {
+        checkout scm
+        withEnv(global_environment_variables) {
+          if (useBrowserStack) {
+            withForcedDockerUpdate([images['ruby']]) {
+                withVaultSecrets([
```
luke-hill, line 192: -1 tab

Rakefile
```diff
@@ -12,14 +14,28 @@ def collect_task_errors(tasks)
   errors.count.positive? && raise("The following tasks failed #{errors}")
 end
 
+def base_cucumber_path
+  "artifacts/#{ENV.fetch('BROWSER', 'unknown')}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'other')}"
```
luke-hill, line 18: Come to think of it this will still fail unless we delete our local configs. We have to have a conditional here. But we can still keep it split in 2.

```
if ENV["BROWSERSTACK"] == "true"
  "artifacts/#{ENV['BROWSER']}/#{ENV['BROWSERSTACK_CONFIGURATION_OPTIONS']}"
else
  "artifacts/#{ENV['BROWSER']}/other"
end
```

testing/cucumber.yml
```diff
@@ -1,7 +1,9 @@
 passing: --tags 'not @failing and not @future_release'
 
-html: --format html --out artifacts/reports/report.html
-junit: --format junit --out artifacts/reports
-json: --format json --out artifacts/reports/report.json
+<% base_path = "artifacts/#{ENV['BROWSER'])}/#{ENV.fetch('BROWSERSTACK_CONFIGURATION_OPTIONS', 'other')}" %>
```
luke-hill: Sorry I know you're going to hate me. But we probably need to set the conditional here too

```
base_path = 
  if ENV["BROWSERSTACK"] == "true"
    "artifacts/#{ENV['BROWSER']}/#{ENV['BROWSERSTACK_CONFIGURATION_OPTIONS']}"
  else
    "artifacts/#{ENV['BROWSER']}/other"
  end
```

luke-hill (changes requested):
The changes requested will allow us to run this locally in all the ways we do currently, as well as on CI with always returning the correct folder structure.

testing/cucumber.yml
simonwgill: Done

Rakefile
simonwgill: Done

Jenkinsfile
simonwgill: Done

Jenkinsfile
simonwgill: Done

Jenkinsfile
```diff
@@ -1,207 +1,301 @@
...
-            description: 'channel to print messages')
-        string(name: 'slackCredentialsId', defaultValue: 'slack-plugin',
-            description: 'id of slack credentials used when posting a message')
+global_environment_variables = []
+
+// Pipeline definition begins here
+
+node('docker && awsaccess') {
+  try {
+    step([
```
zhelyan, line 29: no need to use the generic step syntax, this is already exposed as `cleanWs()`.
`cleanWs()` uses deferred cleanup by default so no need to add it to `try ..finally`

simonwgill: I changed it because I got a no such method error

https://jenkins.devops.citizensadvice.org.uk/view/Design%20System/job/design-system/view/change-requests/job/PR-531/2/console

zhelyan: typo - you used `cleanWS()`, it should be `cleanWs()`

simonwgill: dammit, that list hid the fact it wasn't alphabetically ordered in plain sight

luke-hill (approved):

luke-hill:
I think we should merge this in. And if any issues crop up we can fix them later. This is quite a long running branch now.

Top work by all 🎉

PR #532, opened 2020-11-20
fix: link pseudo selector ordering and wrap footer in nav element [NP-662]

baseonmars:
This PR is connected to the footer work. The link pseudo selectors need to be defined in this order to correctly style visited links which are focused.

Whilst here also wrapped the footer in a nav element. I'm waiting on translations.

### DIFF
--- a/scss/4-elements/_links.scss
+++ b/scss/4-elements/_links.scss
@@ -7,7 +7,6 @@ a,
   color: $cads-language__link-colour;
   text-decoration: underline;
 
-  // Note :visited links should be browser default
   &:visited {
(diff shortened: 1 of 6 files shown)


scss/4-elements/_links.scss
```diff
@@ -8,6 +8,9 @@ a,
   text-decoration: underline;
 
   // Note :visited links should be browser default
```
davidsauntson, line 10: is this still true? I assume not

davidsauntson, line 10:
```suggestion
```

davidsauntson (approved):
, knees and toes

PR #547, opened 2020-11-23
feature: display an icon after footer links [NP-622]

baseonmars:
I've added a generic way of displaying an icon after a footer link - this feature is only used to display an external link icon right now but the effort was more or less identical.

Code in the footer template is also refactored to remove duplication.

Screenshot 2020-11-23 at 12 09 45

@minniessuyuhuang I've added you so you can see a visual preview ahead of this being merged into public website - can you check that visually the link looks right - I've taken size of 12px from the [figma files](https://www.figma.com/file/Im9r6ZM1YWiJK488OWx8a6/AdviserNet-Template-Redesign?node-id=1270%3A345) - by default the icon would appear slightly larger (1rem or 16px).


haml/_footer.html.haml
```diff
@@ -7,39 +7,9 @@
...
-          %h2.cads-footer__section-title= links["column_four"]["title"]
-          - if links["column_four"]["links"]
-            %ul.cads-footer__links
-              - links["column_four"]["links"].each do |link|
-                %li.cads-footer__link
-                  %a.cads-footer__hyperlink{ href: link["url"] }= link["label"]
-
+      - %w[column_one column_two column_three column_four].each do |column|
+        - if links[column]
+          = render partial: "footer_links", locals: { links: links[column] }
```
davidrapson, line 12: Do you mind inlining this partial? Currently using `render` _within_ a partial in design-system severely limits what the `view_path` can be on the Rails side. This problem would go away if/when we move to publishing a rails engine, but for now it's a restriction. See "Allows more granular prepend_view_path
" in https://github.com/citizensadvice/design-system/pull/458 for more context. The loop makes sense though.

davidrapson (changes requested):
👍 To the general approach. With a change / comment about the partial.

davidrapson (approved):
👍 With the latest changes

PR #561, opened 2020-11-24
feat: translated aria label for footer navigation

baseonmars:
This is used to aria-label the footer navigation.


davidrapson (approved):
👍

PR #562, opened 2020-11-24
bugfix: Make mobile work on header tests [NP-761]

luke-hill:
Make sure to account for mobile header differences!

### DIFF
--- a/testing/features/header.feature
+++ b/testing/features/header.feature
@@ -9,8 +9,7 @@ Feature: Header component
     Background:
       Given a Standard Header component is on the page
 
-    # This specification is deliberately removed from Mobile Devices
-    @not_mobile
+    @not_mobile @failing @NP-1235
     Scenario: Header has some quick links
       Then a link to change language is present
       And a login link is present
--- a/testing/features/header.feature
+++ b/testing/features/header.feature
@@ -9,6 +9,8 @@ Feature: Header component
     Background:
       Given a Standard Header component is on the page
 
+    # This specification is deliberately removed from Mobile Devices
+    @not_mobile
     Scenario: Header has some quick links
       Then a link to change language is present
       And a login link is present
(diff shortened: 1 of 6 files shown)


testing/features/step_definitions/header_steps.rb
```diff
@@ -23,6 +23,7 @@
 end
 
 Then("I am able to search for {string}") do |search_term|
+  @component.open_search_pane.click if device?
```
baseonmars, line 26: Might be worth naming this `mobile_phone?` or similar?

Strictly speaking this behaviour is tied to screen size rather than device ID.

baseonmars (approved):
Looks good - left a question about naming `device?` more clearly but don't let that stop you merging this.

testing/features/header.feature
```diff
@@ -9,6 +9,8 @@ Feature: Header component
     Background:
       Given a Standard Header component is on the page
 
+    # This specification is deliberately removed from Mobile Devices
```
davidrapson, line 12: Worth nothing that "deliberately" is too strong. Links _should_ be accessible on small screens but aren't currently we have https://citizensadvice.atlassian.net/browse/NP-1235 to fix this so maybe tag this with that ID?

luke-hill: So do we envisage this behaviour returning in 1235. Or different behaviour appearing? If it's the former I'll tag it as a future release/failing thing. If it's the latter I don't know what to do. maybe just mark it as a future release.

testing/features/step_definitions/header_steps.rb
luke-hill: Good point, will rename. 

Also whilst we don't test on small resolutions, due to % compatibility e.t.c. it might be worth something revisiting later on. I don't think it's worth spending a lot of time on though.

testing/features/header.feature
davidrapson: Future release would be fine, it will be different behaviour

davidrapson (approved):
👍

PR #568, opened 2020-11-26
bugfix: Wrapped pipelines with docker hub credentials [NP-1344]

simonwgill:
Docker is now rate limiting anonymous image pulls by IP address. This adds authentication so that all our pulls inside this build job are associated with our account.


davidrapson (approved):
👍

luke-hill (approved):

PR #571, opened 2020-11-26
feat: Set logo links in header and footer via template local [NP-1274]

KPobeeNorris:
This PR updates the header and footer to accept a "homepage_url" which
allows the consuming code to configure it appropriately.


baseonmars (approved):
Looks good, I'll get an alpha release out once this passes CI.

PR #573, opened 2020-11-26
feature: Spec: Skip links (Header only) [NP-1181]

luke-hill:
I have added spec validation for skip-links. However as this wasn't something directly I fleshed out. I'm not 100% sure I've covered it well. (Hence tagging sheldon as well to look at Business cases).

There's a tiny bit of tech-debt in this PR - Which will be fixed soon (It's in capybara master, but not a versioned candidate). It's something a lot of people have got fixed up in Capybara and it will filter to selenium (When you tell a page to do something native, do it, don't complain).

NB: I've altered a tiny bit of haml/labels. Let me know if this is ok!

### DIFF
--- a/haml/_header.html.haml
+++ b/haml/_header.html.haml
@@ -8,11 +8,11 @@
   .cads-grid-container
     .cads-grid-row
       %nav.cads-skip-links
-        %a.cads-skip-links__link{ href: "#cads-navigation", data: {test_id: "skip-link" } }
+        %a.cads-skip-links__link{ href: "#cads-navigation" }
           = t("cads.header.skip_to_navigation")
-        %a.cads-skip-links__link{ href: main_content_anchor_link, data: {test_id: "skip-link" } }
+        %a.cads-skip-links__link{ href: main_content_anchor_link }
           = t("cads.header.skip_to_content")
-        %a.cads-skip-links__link{ href: "#cads-footer", data: {test_id: "skip-link" } }
+        %a.cads-skip-links__link{ href: "#cads-footer" }
           = t("cads.header.skip_to_footer")
       .cads-grid-col-md-6.cads-header__logo-row
         %a.cads-logo{ href: homepage_url, title: t("cads.shared.logo_title") }
--- a/haml/_header.html.haml
+++ b/haml/_header.html.haml
@@ -8,11 +8,11 @@
   .cads-grid-container
     .cads-grid-row
       %nav.cads-skip-links
-        %a.cads-skip-links__link{ href: "#cads-navigation" }
+        %a.cads-skip-links__link{ data: { test_id: "skip_link", test_id: "skip_to_navigation" }, href: "#cads-navigation" }
           = t("cads.header.skip_to_nav")
-        %a.cads-skip-links__link{ href: main_content_anchor_link }
+        %a.cads-skip-links__link{ data: { test_id: "skip_link", test_id: "skip_to_content" }, href: main_content_anchor_link }
           = t("cads.header.skip_to_content")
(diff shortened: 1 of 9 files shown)


davidrapson:
Controversial suggestion but I would strongly recommend only testing on the public-website end.

As the skip links within the design system are unlikely to point anywhere useful as the component is disconnected from the full context of the page so the skip links will never work. We could potentially update one of the sample pages to do that, but I genuinely think this is best tested within the context of the full website.

haml/_header.html.haml
```diff
@@ -8,11 +8,11 @@
   .cads-grid-container
     .cads-grid-row
       %nav.cads-skip-links
-        %a.cads-skip-links__link{ href: "#cads-navigation" }
-          = t("cads.header.skip_to_nav")
-        %a.cads-skip-links__link{ href: main_content_anchor_link }
+        %a.cads-skip-links__link{ data: { test_id: "skip_link" }, href: "#cads-navigation" }
```
davidsauntson, line 11: I might be missing something but you don't seem to be using these ids now?  If not just remove them.

davidsauntson (changes requested):

haml/_header.html.haml
luke-hill: Oh yes. I am not.

luke-hill:
Sister PR for Public-Website is incoming @davidrapson - Which will validate a similar thing to here but obviously in the setting of a full website.

davidrapson:
In which case I'd say we don't need this. Keep it to public-website.

luke-hill:
We've got the feature tests on the public-website for validating the component mixed in with others and here for isolation tests. That way we have both the integration as well as the component tests, same as the other components. Plus we're going to start using these for documentation down the line (The plan is to have every component here with some sort of spec).

davidsauntson (changes requested):
I've just seen this: https://github.com/citizensadvice/public-website-cucumber/pull/131/files

Looks like you've copied and pasted the code over?

If so agree with David R that these tests should only exist in one place.  Feels like you could test more thoroughly with content to skip to, ie in public-website

luke-hill:
So here the tests as a documentation tool. ie this covers the spec of what the component will do.

The tests on here "shouldn't" ever change. The ones on the main website might. That's the reason for having them in both places. This acts as a spec for the design-system (and each component).

For the main website, we might be able to add some extra validations later. But currently there is no meaty stuff that changes on the site (Other than the native scrolling to a js section, which is untestable in automation). The ones on the site also partially don't run yet (Not until we make some other changes), and will take a lot longer to run (~20s compared to ~1s)

davidrapson:
Yep, still think this only needs to be in public website. On the "here for isolation tests" point, this is precisely why I don't think this test is helpful because this component _can't_ be meaningfully tested in isolation, only in context. We could get around that by setting up a page like https://citizensadvice.github.io/design-system/?path=/story/sample-pages-generic--tables to be a full test case, but at that point you're just duplicating what we'd do in public-website wholesale.

luke-hill:
The purpose of this file isn't just to perform tests. It's also to act as a barometer for the specification of the design system. 

Eventually we'll have a test for each of the components in isolation. Even if that test is just a primitive - The component exists test. Because alongside that we'll have a description of what each component does which can be exported to the Specification Document we'll be providing either on demand or online.

The tests here just validate the form of the component. And on the main site we may introduce more tests later as mentioned.

The code being copied over is symptomatic of gherkin. That's the point of it, to describe the behaviour which is identical for both the site and the component but functionally different (Clicking different things / links).

davidrapson (approved):

PR #574, opened 2020-11-26
Feature: /np 1263 ds footer migration

celonicolosi:
Migrate the existing Footer scenario into a new variant scenario.
- create a folder footer in support/components;
- separate each type of footer in a different file, default.rb and minimal.rb;
- change elements and sections to its respective support file;
- create a new scenario for minimal footer
- adjust some step_definitions

### DIFF
--- a/testing/features/step_definitions/footer_steps.rb
+++ b/testing/features/step_definitions/footer_steps.rb
@@ -21,15 +21,15 @@ Then("each header item has at least 1 link below it") do
 end
 
 Then("a footer logo is present") do
-  expect(@component.initial_form).to have_logo
+  expect(@component).to have_logo
 end
 
 Then("a copyright notice is present") do
-  expect(@component.initial_form).to have_copyright
+  expect(@component).to have_copyright
 end
 
 Then("a company info is present") do
-  expect(@component.initial_form).to have_company_info
+  expect(@component).to have_company_info
 end
 
 Then("I am presented with a form to report the issue about the page I am on") do
--- a/testing/features/step_definitions/footer_steps.rb
+++ b/testing/features/step_definitions/footer_steps.rb
@@ -1,7 +1,12 @@
 # frozen_string_literal: true
 
-Given("a Footer component is on the page") do
-  @component = DesignSystem::Footer.new
+Given("the Default Footer component is on the page") do
+  @component = Footer::Default.new
+  @component.load
+end
+
+Given('the Minimal Footer component is on the page') do
+  @component = Footer::Minimal.new
   @component.load
 end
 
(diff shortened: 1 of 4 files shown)


testing/features/step_definitions/footer_steps.rb
```diff
@@ -1,16 +1,21 @@
 # frozen_string_literal: true
 
-Given("a Footer component is on the page") do
-  @component = DesignSystem::Footer.new
+Given("the Default Footer component is on the page") do
+  @component = Footer::Default.new
```
luke-hill, line 4: lets write this as tap like the others to save space.

testing/features/support/components/footer/minimal.rb
```diff
@@ -1,15 +1,10 @@
...
-    set_url "/iframe.html?id=components-footer--default-story&viewMode=story"
-
-    sections :category_titles, ".cads-grid-col-md-3 > h2" do
-      elements :links, :xpath, "./following-sibling::ul[1]/li/a"
-    end
+module Footer
+  class Minimal < ::Base
+    set_url "/iframe.html?id=components-footer--minimal&viewMode=story"
 
     section :initial_form, ".cads-footer" do
```
luke-hill, line 7: lets remove this wrapping here. And instead just have the components as top level ones. NB: You'll need to edit the below method too.

testing/features/support/components/footer/default.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module Footer
+  class Default < Minimal
+    set_url "/iframe.html?id=components-footer--default-story&viewMode=story"
+
+    sections :category_titles, ".cads-grid-col-md-3 > h2" do
+      elements :links, :xpath, "./following-sibling::ul[1]/li/a"
+    end
+
+    element :website_feedback, ".cads-footer__feedback a"
```
luke-hill, line 11: we need a validate method below this one. `has_category_titles?(wait: 5)` should suffice.

luke-hill (changes requested):
Couple of minor things to change, but other than that seems good.

Can you test it locally on an iPhone first as well, just to validate we won't merge this and it goes pete-tong

testing/features/support/components/footer/minimal.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Footer
+  class Minimal < ::Base
+    set_url "/iframe.html?id=components-footer--minimal&viewMode=story"
+
+    element :logo, ".cads-logo"
+    element :copyright, ".cads-footer__company-info div p:nth-of-type(1)"
+    element :company_info, ".cads-footer__company-info div p:nth-of-type(2)"
+
+    def validate_initial_state!
```
luke-hill, line 11: the standard validation here should be used so we're not loading based off a validated assertion.

luke-hill: There's a standard validation on `::Base` - So you can just remove this method.

luke-hill: @celonicolosi We can remove this validation and just use the base one (And remove the footer element above)

luke-hill (approved):

PR #575, opened 2020-11-27
Feature: DS component search and it's variant [NP-1184]

celonicolosi:
Automation for the search components and the variant search component with value.

- create a folder search in support/components;
- separate each variant of search in a different file, standard.rb and with_value.rb;
- write all the scenarios
- develop the step_definition


testing/features/support/components/search/standard.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def validate_initial_state!
+      has_cads_search?(wait: 5)
+    end
+
+    def clear_field
+      cads_search.search_field.value.length.times do
+        cads_search.search_field.send_keys [:backspace]
+      end
+    end
```
celonicolosi, line 20: I tried:
```
@component.cads_search.search_field.send_keys([:command, 'a'], :backspace)
```
But this way was not going to work for windows since it is `[:control, 'a']`

testing/features/search.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Search components
+
+  The Search component allows users to search for the desired term and consists
+  of a search field and a button. The variant 'Search with value' has a
```
luke-hill, line 4: lets omit this description here, as this isn't parsed in the html report proper. And amend the cukes below

testing/features/search.feature
```diff
@@ -0,0 +1,26 @@
...
+  pre-defined value in the search field.
+
+  Rule: Standard Search
+    Background:
+      Given a Standard Search component is on the page
+
+    Scenario: Search component is on the page without any value
```
luke-hill, line 11: User is able to search

testing/features/search.feature
```diff
@@ -0,0 +1,26 @@
...
+      Then the search field is clear
+      And I can search for "Anything"
+
+  Rule: Search with value
```
luke-hill, line 15: Value*

testing/features/search.feature
```diff
@@ -0,0 +1,26 @@
...
+    Background:
+      Given a Search With Value component is on the page
+
+    Scenario: Search component is on the page with a pre-defined value
+      Then I can search for the pre-defined value "Current search term"
+
+    Scenario: It is possible to change the pre-defined value
```
luke-hill, line 22: This scenario is excess, because we're describing native events. Instead lets absorb this into the above.

testing/features/search.feature
```diff
@@ -0,0 +1,26 @@
...
+    Scenario: Search component is on the page without any value
+      Then the search field is clear
+      And I can search for "Anything"
+
+  Rule: Search with value
+    Background:
+      Given a Search With Value component is on the page
+
+    Scenario: Search component is on the page with a pre-defined value
+      Then I can search for the pre-defined value "Current search term"
```
luke-hill, line 20: How about this

```
Then the search field has a pre-defined value
And I can search for "Anything"
```

testing/features/step_definitions/search_steps.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+Given("a Standard Search component is on the page") do
+  @component = Search::Standard.new
```
luke-hill, line 4: use tap load here to save space

testing/features/step_definitions/search_steps.rb
```diff
@@ -0,0 +1,39 @@
...
+  @component.load
+end
+
+Given("a Search With Value component is on the page") do
+  @component = Search::WithValue.new
+  @component.load
+end
+
+Then("the search field is clear") do
+  expect(@component.cads_search.search_field.value).to eq("")
```
luke-hill, line 14: `.to be_empty`

testing/features/support/components/search/standard.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Search
+  class Standard < ::Base
+    set_url "/iframe.html?id=components-search--search&viewMode=story"
+
+    section :cads_search, ".cads-search" do
```
luke-hill, line 7: Shouldn't need this wrapper.

testing/features/support/components/search/standard.rb
```diff
@@ -0,0 +1,22 @@
...
+      element :search_field, ".cads-search__input"
+      element :search_button, ".cads-search__button"
+    end
+
+    def validate_initial_state!
+      has_cads_search?(wait: 5)
```
luke-hill, line 13: alter this to wait for the button

testing/features/support/components/search/standard.rb
luke-hill: I've no issue with this, but it's a bit clunky / heavy. A quicker one would be a native double click. This highlights all.

So something like
```
search_field.double_click
search_field.send_keys(:delete)
```

testing/features/step_definitions/search_steps.rb
```diff
@@ -0,0 +1,39 @@
...
+  expect(@component.cads_search).to have_search_button
+end
+
+Then("I can search for the pre-defined value {string}") do |pre_defined_value|
+  expect(@component.cads_search.search_field.value).to eq(pre_defined_value)
+
+  expect(@component.cads_search).to have_search_button
+end
+
+Then("the search field is filled with {string}") do |pre_defined_value|
```
luke-hill, line 31: so based on my above review and the changes here. This then becomes

`expect(field).not_to be_empty`

luke-hill (changes requested):
Remember we're trying to describe behaviour. Some of the actions you've described are very low-level / native. We'd never test these as they vary from OS to OS and it's not in our remit if something breaks here, the vendor would fix it.

testing/features/support/components/search/standard.rb
celonicolosi: The problem with double_click is that the pre-defined value has more than one word, so the double_click just selects the last one.
Screenshot 2020-11-27 at 18 12 05
Screenshot 2020-11-27 at 18 12 17

testing/features/search.feature
luke-hill: A Search "value" is a technical term we want to try avoid.

testing/features/search.feature
```diff
@@ -0,0 +1,21 @@
...
+  The Search component allows users to search for the desired term and consists
+  of a search field and a button.
+
+  Rule: Standard Search
+    Background:
+      Given a Standard Search component is on the page
+
+    Scenario: Search component is on the page without any value
+      Then the search field is clear
+      And User is able to search for "Anything"
```
luke-hill, line 12: the user*

testing/features/search.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Rule: Search with Value
+    Background:
+      Given a Search With Value component is on the page
+
+    Scenario: Search component is on the page with a pre-defined value
+      Then the search field has a pre-defined value
```
luke-hill, line 19: lets call this and the above a pre-defined `term`

testing/features/step_definitions/search_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  expect(@component).to have_search_button
+end
+
+Then("the search field has a pre-defined value") do
+  expect(@component.search_field.value).not_to be_empty
+
+  expect(@component).to have_search_button
+
+  @component.clear_field
```
luke-hill, line 28: This shouldn't live here,it should be part of a method called `search_for` inside the component

```
@component.search_for(term)
```

With the implementation then wiping the field

testing/features/support/components/search/standard.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Search
+  class Standard < ::Base
+    set_url "/iframe.html?id=components-search--search&viewMode=story"
+
+    element :search_field, ".cads-search__input"
+    element :search_button, ".cads-search__button"
+
+    def validate_initial_state!
```
luke-hill, line 10: remove this validation as we're asserting on it.

testing/features/support/components/search/standard.rb
```diff
@@ -0,0 +1,20 @@
...
+      has_search_button?(wait: 5)
+    end
+
+    def clear_field
+      search_field.value.length.times do
```
luke-hill, line 15: just do element.double_click then element.send_keys(:delete) it's more optimal.

luke-hill (changes requested):
Few minor tweaks still outstanding.

luke-hill:
Probably worth (As this is a new spec), getting one of @davidrapson / @davidsauntson to review this too.

celonicolosi:
I created the clear_field method, and below are all my tries and results, I'm out of ideas since firefox didn't work on my last solution of use **ctrl + a** and **command + a** together. I'm accepting suggestions.

```
def clear_field
      #double_click: just select the last word, so this solution does not work in any browser
      search_field.double_click
      search_field.send_keys [:backspace]

      #triple_click: does not exists
      search_field.triple_click
      search_field.send_keys [:backspace]

      #Ctrl + a and/or Command + a:
        # - Ctrl + a works for all browsers on Windows except Firefox
        # - Command +a works for all browsers on Mac except Firefox
        #-  both combined works for all browsers on Mac and Windowss except Firefox
      search_field.send_keys [:control, "a"]
      search_field.send_keys [:command, "a"]
      search_field.send_keys [:backspace]

      # the previous solution works on all browsers and all OSs
      search_field.value.length.times do
        search_field.send_keys [:backspace]
      end
    end
```

davidrapson:
FWIW if the clear field stuff is proving a problem then I don't think this is worth worrying about here. This is a plain search box, we're not adding any custom behaviour. Is this not testing browser native behaviour?

If you've got it working then 👍

celonicolosi:
> FWIW if the clear field stuff is proving a problem then I don't think this is worth worrying about here. This is a plain search box, we're not adding any custom behaviour. Is this not testing browser native behaviour?
> 
> If you've got it working then 👍

Don't worry David, I pair with Luke this afternoon, and I think we found a good solution for this clear field stuff and the better part is... work on **mobile phones** too

luke-hill (approved):

PR #590, opened 2020-11-30
bugfix: Corrected path for grid test artifacts [NP-1273]

simonwgill:
The change from grid to other was late in the previous PR, and the update to the Jenkinsfile was missed.

- Updated artifact search path to use the correct other path when running grid tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,7 +15,7 @@ def collect_task_errors(tasks)
 end
 
 def base_cucumber_path
-  "artifacts/#{ENV['BROWSER']}/#{ENV.fetch('BROWSERSTACK', 'false') != 'false' ? ENV['BROWSERSTACK_CONFIGURATION_OPTIONS'] : 'other'}"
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -15,7 +15,7 @@ def collect_task_errors(tasks)
 end
 
 def base_cucumber_path
-  "artifacts/#{ENV['BROWSER']}/#{ENV['BROWSERSTACK'] ? ENV['BROWSERSTACK_CONFIGURATION_OPTIONS'] : 'other'}"
+  "artifacts/#{ENV['BROWSER']}/#{ENV.fetch('BROWSERSTACK', 'false') != 'false' ? ENV['BROWSERSTACK_CONFIGURATION_OPTIONS'] : 'other'}"
```
luke-hill, line 18: if you just write this as `ENV["BROWSERSTACK"] == "true"` it uses the same check we use in the main site. It's slightly more optimal as it won't do a double eval.

luke-hill (approved):

PR #594, opened 2020-12-01
Experimental: Rails engine

davidrapson:
As demoed earlier today this is a 🌱 experimental 🌱 attempt at distributing the haml partials as a Rails engine using view components. There are two main parts:

## `engine/`

Contains the Rails engine code. Consisting of component versions of about 2/3 of the existing haml partials along with a version of the locale files. Also includes specs for most of the components.

Screenshot 2020-12-01 at 12 16 20

## `demo/`

A stripped back demo Rails app which loads both the engine and the npm package as local dependencies. This contains a basic home controller and view that you can use to experiment with the components.

Screenshot 2020-12-01 at 12 17 40

If you are running this locally you may also need to run `npm run build` at the root of the project to make sure the javascript is built as it would be if it were published to npm.

## Installing in an existing rails app

You can add the following to your Gemfile to try this out in an existing Rails app

```
gem "design_system", github: "citizensadvice/design-system", branch: "experimental/engine", glob: "engine/*.gemspec"
```

See https://github.com/citizensadvice/public-website/pull/601/files for reference


davidsauntson (commented):
Probably this is a conversation to have with more people and not on GitHub.

But I'd like to take a step back and ask us all to consider the long term plan for the design system and where it's headed.  My concerns mostly lie around whether this would put people off using the design system - my hunch is that local CA offices (who we would like to be users of the design system) won't have the knowledge or be able to get the resources to build rails apps and I don't think it's something we should be forcing onto users.  

Additionally, Casebook will want to be able to consume some elements of the design system (eventually) in their react components and forcing them to scrape out the HTML and only use the CSS would be a shame and a missed opportunity for CA, I think.

I know we already have ruby in our haml, and are using ruby to provide translations but I don't think either of those decisions were made whilst considering the long term use case of the design system.

Additionally additionally, CA doesn't require their front end recruits to know anything about ruby, rails or even HAML and it's a pretty steep learning curve if you're not prepared for it.  Certainly I will feel much less confident making changes to the design system.  Happy to learn etc but it is something to bear in mind from a long term resourcing perspective.

Ultimately I think much of this will be decided when/if(?) the design system team is inaugurated.

davidrapson:
Yep, that makes sense. As a companion to this work I'm planning on putting together a bit of a proposal as to where the separation would lie with this particular model.

The problem this experiment explores is specifically the use case of integrating with Rails apps (which is what both we do on new platform, smart energy, and on Casebook). But would 100% not replace the use-case of using this stuff directly, nor impose the restriction of using Rails. It's essentially a way of publishing the current haml + locale files in a more Rails-appropriate way (currently it's a bit of a fudge) and it way that can give us more confidence as they can be unit tested.

But this could likely just be _an_ implementation not _the only_ implementation. For example I could also see is publishing React equivalents.

If anything this has the potential to make it _easier_ to document HTML-only equivalents too much like https://design-system.service.gov.uk/components/date-input/ provides HTML snippets by default along with a language specific version (nunjucks in that case, rails components in ours) as a more domain-specific option.

PR #595, opened 2020-12-01
Breadcrumb bugfixes

davidrapson:
## Remove redundant title attribute (NP-1277)

The title attribute contains the same text as the link itself

## Make individual breadcrumb inline-block (NP-1186)

Fixes additional space after link

**Before**:

image

**After**:

image

Fun fact: This is such a small difference that Backstop doesn't care about it. I updated the snapshots in this PR anyway.


baseonmars (approved):
good stuff.

PR #597, opened 2020-12-01
Documentation: Design system guidance Np-1306

minniessuyuhuang:
To help the user of the Design System know how to use the foundations and components. 
- Added the guidelines for icons, breadcrumbs, callouts, and targeted contents.


styleguide/components/breadcrumbs/breadcrumbs.stories.mdx
```diff
@@ -11,11 +11,21 @@ import templateNoCollapse from './_no-collapse.html.haml';
...
 
-## Default (collapse)
+## When to use
 
-By default they will collapse to the parent page on mobile. You can explicity set this behaviour by passing in
+Use breadcrumbs when a page is located in a hierarchy of at least two levels. 
+
+## When not to use
+
+Do not use breadcrumbs on a website with a flat hierarchy or to show the journey through a transaction. If a page is taking users through a multistep process, use the "Step-by-step sidebar navigation" (or consider if a progress indicator is needed) instead.
```
davidrapson, line 22: Step-by-steps don't exist in the design system, so is there enough context here to know what this means? Can we link to a figma file that helps people understand?

styleguide/components/breadcrumbs/breadcrumbs.stories.mdx
```diff
@@ -11,11 +11,21 @@ import templateNoCollapse from './_no-collapse.html.haml';
...
+
+## How it works
+
+Breadcrumbs should be placed on the top left of each page and below "Global navigation". They should start with a link to the home page and end with the current page title. Page titles of parent pages should be linked; the current should not have a link.
```
davidrapson, line 26: If you are using the provided template in Rails "the current should not have a link" this will happen for you.

styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -8,9 +8,11 @@ import { translate } from '../../../scripts/story-helpers';
 
 # Callouts
 
-There are several different variants of the callout component:
+There are 4 main callout styles in our Design system:
```
davidrapson, line 11: Nit: Should this be Design System or design system (capitalisation)?

styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -23,7 +25,9 @@ import templateStandard from './_standard.html.haml';
   </Story>
 </Preview>
 
-## Example
+## Example callout
+
+This style is when an example needs to be displayed. Also, use it for quotes.
```
davidrapson, line 30: "Also, use it for quotes". Interesting, do you have an example of this? I'd expect quotes to be their own design pattern but I might just be missing the context here.

styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -49,7 +55,9 @@ import templateImportant from './_important.html.haml';
   </Story>
 </Preview>
 
-## Adviser
+## Adviser callout
+
+Content that is for advisers. This will only show on the logged-in public site and AdviserNet.
```
davidrapson, line 60: "This will only show on the logged-in public site and AdviserNet". Perhaps we could reframe this to talk about intent (i.e. if another product is looking to use this component).

Something like: "Content that is for advisers. These should not be shown to the public, and you should only show this when an adviser is logged in"

styleguide/components/callouts/callouts.stories.mdx
```diff
@@ -62,6 +70,26 @@ import templateAdviser from './_adviser.html.haml';
...
+
+Before using a callout, make sure you really need it. Fewer callouts have more of an impact on a piece of content. Too many callouts can confuse and distract users from other important content on a page. 
+
+## Usage
+
+Callouts should come with a title that explains the subject of the information when possible. This will give the users an idea of what the information is about.
+
+### Markup
+
+The title should be used as `h3` no matter where they are on the page. If a callout is the first thing on the page under a `h1`, still put it in the backend as a `h3` and the CMS will make sure it displays correctly.
```
davidrapson, line 90: The design system probably shouldn't have too much knowledge of Contentful here. I actually think this is a problem with the content model if we're asking people to knowingly put in incorrect headings. Ideally the content model should have a dedicated title field, and the website could work out what the best heading level is based on context.

For now, I'd suggest rather than putting this guidance here we maybe add it as a help field in Contentful?

styleguide/components/targeted-content/targeted-content.stories.mdx
```diff
@@ -17,6 +17,25 @@ import templateAnchor from './_anchor.html.haml';
...
+
+Because targeted contents hide content, they should always have a clear label/title that helps users understand what is in the section. You can use `h3` within them to break up the content.
+
+## When to use 
+
+When content is only relevant to some particular users visiting the page.
+
+## When not to use
+
+They shouldn’t be used to make a page seem shorter, for example: using them to hide large chunks of content.
```
davidrapson, line 32: 💯

styleguide/foundations/icon-font.stories.mdx
```diff
@@ -72,3 +87,36 @@ Do not forget to prefix the given class name with `cads-icon_`.
...
+### Considerations before creating a new icon
+
+Before trying to create a new icon, consider if:
+- The new icon serves a need(s). If in doubt, you probably don’t need it
+- The new icon would be used multiple times throughout the products
+- The new icon has a clear meaning
+
+Propose the new icons to the design community and discuss details and requirements. This will help us to have a transparent process for creating new assets, maintaining a minimal and high quality icon set in our Design system. 
+
+### Making an icon
```
davidrapson, line 108: Interesting. Some really great principles in here but/and I think some of these steps might be unnecessary in the figma stage as we could handle it in code. I have a couple of clarifying questions first though, will ask on slack.

davidrapson (changes requested):
💯 This wonderful, and really well written. A few queries and clarifications, and I have some follow up questions about icons but this is fantastic so far.

styleguide/components/breadcrumbs/breadcrumbs.stories.mdx
minniessuyuhuang: Good point, will link to figma

styleguide/foundations/icon-font.stories.mdx
davidrapson: fWe chatted through this the notes all make sense now. Suggested adding a visual example of the pixel grid requirements to show what to look out for, as well as a good and bad example of SVG output.

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: This was from [the guidance that Zarlashta](https://docs.google.com/document/d/1xBPM_hrPuOyGoeiWYDTEXcNSvQjWGlup08tAIXG5f0w/edit#heading=h.b5hdw4glkauw), a content strategist, wrote a while ago. There was no example image but I [found this on the public site](https://www.citizensadvice.org.uk/debt-and-money/if-you-cant-pay-your-bills-because-of-coronavirus/#h-overdrafts).

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: Love this kind of detail, hyphen or dash? :grimacing: 
Let's have a rule and stick to it. Should we use Sentence Case for any specific names, such as Design System, Related Content, and Targeted Content?

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: Sounds good, will update. Thanks!

styleguide/components/callouts/callouts.stories.mdx
davidrapson: Yeah I'd use the title case version like Design System for describing the system as a whole (just capitalising Design doesn't make sense), sounds sensible to do that for components too/

styleguide/components/callouts/callouts.stories.mdx
davidrapson: Ah interesting. In which case we probably don't want this guidance next to the example callout, as that component will always include the "Example" tag, which doesn't make sense for a quote. Perhaps suggest using the standard variant, or we need a separate design pattern for quotes.

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: Actually, I'll just update "Design System" for now. As this is just a draft and will require a content designer (probably Adam) to review, I think he is the best person to look at these details.

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: Good point. I don't know if we've considered another design for quotes, @soniaturcotte ? Let me ask Zarlashta this question in her doc and see what she thinks.

styleguide/components/callouts/callouts.stories.mdx
minniessuyuhuang: Not sure about this one, as we originally included content designers as the users of the Design System. What are your thoughts @soniaturcotte ?

davidrapson: Ah yeah, I'm mostly splitting hairs about where this guidance should _live_. It feels like the wrong level, but that's mostly because the way we solve this is kinda funky via some javascript after the fact rather than solving it within the content model. I reckon we could leave this as is though as it reflects the current state of things.

minniessuyuhuang: Sounds good to me

styleguide/foundations/icon-font.stories.mdx
minniessuyuhuang: I've updated this section based on the suggestions above. I think the images for the pixel section work well; however, I'm not too sure about the SVG ones. They feel a bit overkill to me, what do you think?

minniessuyuhuang:
I've addressed most of the comments. Still trying to find out the answer to "Example callout for quotes" as Zarlashata didn't know too much about it…

davidsauntson (approved):
✏️

minniessuyuhuang:
Can anyone help me with the conflict situation please? :pray:

davidrapson (approved):
👍 Latest changes look fab!

PR #600, opened 2020-12-02
refactor of env variables to just use local or master to determine APP_ENVIROMENT

mankind:
The PR is seeking to address failing builds in Jenkins.  These failures were caused by some environment variables not being passed in some context and from issues related with how docker was being used.

- Refactored code in the following files
  - AUTOMATED_BROWSER_SUPPORT.md
  - docker-compose.yml
  - testing/.env.example
  -  testing/browserstack_ci.env
  -  testing/features/support/helpers/env_variables.rb 
  - testing/grid_ci.env


simonwgill:
Where was the original problem you were trying to solve?

luke-hill:
Try use the templates @mankind as that'll help set out the issue better. It's a bit mangled / incorrectly named atm.

testing/features/support/helpers/env_variables.rb
```diff
@@ -55,7 +55,11 @@ def headless?
     end
 
     def base_url
-      ENV.fetch("APP_URL", "https://citizensadvice.github.io/design-system")
+      if ENV["APP_ENVIRONMENT"] == "local"
```
luke-hill, line 58: can we refactor this and include a secondary method.

So lets have
`def environment` that takes the input, does a fallback to sym - Make the fallback the master website.
then `base_url` takes the symbol and does the if/else

luke-hill (changes requested):
In `grid_ci.env` we need to remove browser

testing/features/support/helpers/env_variables.rb
mankind: changes applied and  pushed @luke-hill

luke-hill (approved):
LGTM

PR #601, opened 2020-12-02
fix: breadcrumb alignment [NP-1188]

baseonmars:
I've changed the `nav` element of Breadcrumbs act as a site level component. The design calls for a border that spans the entire page.

Screenshot 2020-12-02 at 15 08 23
Screenshot 2020-12-02 at 15 08 48


davidrapson (changes requested):
👍 Yes, but might want to check how these behave in the context of the search breadcrumbs within public website.

Currently they use the same styles via the "no-collapse" variant. But I have the feeling we'd be better off decoupling this and inlining the search breadcrumb styles into public-website as they behave fairly differently.

baseonmars:
Hey David, that's a good point. What's the change requested here - make an additional PR in public-website?

davidrapson:
I reckon so yeah. Decouple the styles so they don't clobber each other. Search breadcrumbs feel special case enough that it's not worth fighting against this component.

baseonmars:
I'll see if I can come up with a better solution than copy/pasting styles from one repo to the other. I think it's a bit of a fail to not be able to re-use styles, they're mostly the same, it's just the containers that are different.

davidrapson:
True. It's also not as bad as I thought

image

So maybe we remove constraints and borders from the base component and have a `cads-breadcrumbs-container` that does the layout work?

baseonmars:
That was my initial approach but the bem linter complained at me for trying to turn a border off using a selector.

davidsauntson (commented):
On the search breadcrumbs - it was specifically requested that they use the same component, which was the whole point of adding the no-collapse variant. 

The only difference in the breadcrumbs in search is that they have no bottom border, the no collapse is achieved through the variant.

davidrapson:
Cool, yep. I think the base component is just doing too much at the moment. So as long as we can make sure the borders and constraints only apply when used in the global context then I think we're good.

baseonmars:
@davidrapson This change is now non-breaking so should be ok to merge pending another code review.

davidrapson (approved):
👍 Nice! Tested this against public website and all looks good.

PR #603, opened 2020-12-03
 feature: Optimise Test Code - Use `ca_testing` [NP-1413]

luke-hill:
This PR introduces a new gem to the stack `ca_testing` where a lot of our custom testing code will be housed.

For now this gem can only create a local driver, but it has been rigorously tested and uses the same logic that is present here.

NB: I also optimised the CI run to speed up, by only ever running ONE of Grid / Browserstack. Because if we're going to run browserstack, there is 0 value from running grid.


celonicolosi (approved):
LGTM! 👍

PR #604, opened 2020-12-03
Adjust targeted content border style

davidrapson:
Mentioned in: https://citizensadvice.atlassian.net/browse/NP-1034

Designs specify this uses 1px link border. Current behaviour was inheriting wider focus styles.

**Before:**

Screenshot 2020-12-03 at 11 38 11

**After:**

Screenshot 2020-12-03 at 11 33 38


scss/6-components/_targeted-content.scss
```diff
@@ -10,16 +10,15 @@
     display: block;
     color: $cads-language__brand-primary;
     background: $cads-palette__white;
-    border: solid $cads-border-width-small $cads-language__brand-primary;
+    border: 1px solid $cads-language__link-colour;
```
davidsauntson, line 13: is there a reason that you aren't using `$cads-border-width-small` here anymore?

scss/6-components/_targeted-content.scss
```diff
@@ -50,16 +50,21 @@
       width: 100%;
       padding: $cads-spacing-3 $cads-spacing-4;
       text-align: left;
+      border: 1px solid $cads-language__link-colour;
 
       @include cads-media-breakpoint-down(sm) {
         align-items: baseline;
       }
 
-      @include cads-interactive-element();
```
davidsauntson, line 58: I get that this mixin doesn't do exactly what you want it to, but the focus styles in this mixin are the same as the ones you've added.  Just wondering if there is a way we can tie the two together using extend or another mixin perhaps?

Am a bit wary of having independent instances of these styles in mixins - it gives the impression we can just change the mixin to update interactive styles everywhere, which won't be the case here.

Maybe this isn't a real problem, what do you think?

davidsauntson (commented):
👍  - couple of small questions about reusing variables / mixins where we can

scss/6-components/_targeted-content.scss
davidrapson: Small is 2px, but the designs specify 1px. But maybe we need a variable for that.

scss/6-components/_targeted-content.scss
davidrapson: Yeah I did wonder about that, the targeted content felt like a reasonable balance as it already has to do a bunch of "resetting" because of the way that toggle works. I do agree though, maybe we need a lower level mixin for the focus styles. I'll have a look.

scss/6-components/_targeted-content.scss
davidsauntson: It's 1px - I just checked 😄 
https://github.com/citizensadvice/design-system/blob/35256977409c5e1998757702809561805f96ade7/scss/1-settings/_borders.scss#L2

davidrapson: 😆 Whoops!

davidsauntson (approved):
:dart: lush

PR #611, opened 2020-12-04
fix: use the grid to align footer [NP-1318]

baseonmars:
Use the grid! This takes away the extra 1rem horizontal padding and just uses grids (as requested by design). footer content is now aligned with content across the board.

Screenshot 2020-12-04 at 15 17 57
 
Screenshot 2020-12-04 at 15 18 09


davidrapson (approved):
📏

PR #613, opened 2020-12-04
Add the tag not_mobile for Scenario

celonicolosi:
This PR just add the tag @not_mobile in the Scenario: Users can report a problem about the specific page they are on.
A permanent fix is going to be done soon, but in order to not block any developers builds this is the fastest solution for now.

### DIFF
--- a/testing/features/footer.feature
+++ b/testing/features/footer.feature
@@ -15,6 +15,7 @@ Feature: Footer component
       And a copyright notice is present
       And a company info is present
 
+    @not_mobile
     Scenario: Users can report a problem about the specific page they are on
       When I report a problem with this page
       Then I am presented with a form to report the issue about the page I am on


baseonmars (approved):
Any idea why this fails?

luke-hill:
Have we got a ticket to triage this? If we're not fixing it now?

luke-hill (approved):

celonicolosi:
> Any idea why this fails?

I think the problem is how Appium is handling the following step:

Then("I am presented with a form to report the issue about the page I am on") do
  switch_to_newly_opened_window!(new_page: true)

  expect(page.current_url).to eq("https://www.research.net/r/J8PLH2H")
end

celonicolosi:
> Have we got a ticket to triage this? If we're not fixing it now?

I'm creating it, I just wanted to unblock it as soon as possible since is Friday end of the day.

PR #614, opened 2020-12-04
Migrate notice banner docs

davidrapson:
- Migrate notice banner to new docs
- Fix minor issue with page review component that caused storybook navigation to collapse


baseonmars (approved):
LGTM

PR #620, opened 2020-12-07
Scope stories glob

davidrapson:
Scope `stories` glob to be the `styleguide` directory rather than searching the whole project. 

Not hugely noticeable on master but makes a big difference in the context of https://github.com/citizensadvice/design-system/pull/594 where there are far more files.


baseonmars:
Smart. I like it. 👍

baseonmars (approved):

PR #635, opened 2020-12-10
feature: external links [NP-1382]

baseonmars:
Screenshot 2020-12-10 at 14 26 44

### DIFF
--- a/scss/5-objects/_prose.scss
+++ b/scss/5-objects/_prose.scss
@@ -51,4 +51,9 @@
   .cads-list-no-bullet {
     @extend %cads-list-no-bullet;
   }
+
+  a[href*="//"]:not([href*="citizensadvice.org.uk"])::after
+  {
+    @extend %cads-hyperlink--external-icon;
+  }
 }
--- a/styleguide/foundations/typography/typography.stories.mdx
+++ b/styleguide/foundations/typography/typography.stories.mdx
@@ -172,7 +172,7 @@ Read our guidance about using <a href ="?path=/docs/components-buttons--primary"
 
 ### External Links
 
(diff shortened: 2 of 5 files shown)


styleguide/foundations/typography/typography.stories.mdx
```diff
@@ -170,6 +170,30 @@ export const linkTextHtml = `<p class="my-paragraph"><a href="">cads-hyperlink</
 
 Read our guidance about using <a href ="?path=/docs/components-buttons--primary">buttons as links</a>.
 
+### External Links
+
+When you link to a page external to your website add the `.cads-hyperlink--external` class to display an icon after the link. This helps to signals to a user that they are leaving the current website.
```
baseonmars, line 175: This reason seems to be disputed internally.

scss/5-objects/_prose.scss
```diff
@@ -51,4 +51,9 @@
   .cads-list-no-bullet {
     @extend %cads-list-no-bullet;
   }
+
+  a[href*="//"]:not([href*="citizensadvice.org.uk"])::after
```
davidrapson, line 55: Gnarly, but I guess this makes sense

davidrapson (approved):
👍  I imagine the rationale for the icon will need updating as you mentioned. I think it's more "to maintain parity with existing pages" as these icons are ambiguous. No doubt @soniaturcotte or @minniessuyuhuang could frame that better.

baseonmars:
> 👍 I imagine the rationale for the icon will need updating as you mentioned. I think it's more "to maintain parity with existing pages" as these icons are generally speaking ambiguous. No doubt @soniaturcotte or @minniessuyuhuang could frame that better.

I'll remove it. We should be able to stand behind our decisions rather than give a weak reason.

PR #638, opened 2020-12-11
chore: match against http and https for external links (tighter match…

baseonmars:
… than previous

Match against `https://` or `http://` at start of href, rather than just `//`. This will avoid showing external link icon for urls that just contain a double slash.


davidrapson (approved):
👍 Makes sense

PR #640, opened 2020-12-11
Design system guidance [Np-1306]

minniessuyuhuang:
Add new design system guidance and update:
- Added guidance for colours, buttons, error summary, input, and radio.
- Updated the breadcrumbs guidance


soniaturcotte:
A few notes: 

## Colours: 
This is a bit confusing now, with the tables and then the swatches below. Particularly because the names in both places are different. A few suggestions: 

- It would be good to see the colours in context in the table view. 
- If adding a preview of the colour isn't feasible in the table, then move the swatches to the top. 
- In the swatches/implementation view, some of the text labels are overlapping. 
- Name the colours the same thing in both places 

## Buttons 
Can we get the actual implementation of these tertiary button and "Next" "Previous" instead of screenshots?

## Radios 

- The heading to text ratio is a bit heavy, which makes it difficult to read. We should edit the text so it's clearer, possibly removing the headings.
- Inline varient should show an inline example 
- Remove "Varient:" from before each example, it's not necessary 
- Add the same link to error summary for the error example (as in Inputs) 
-

davidrapson:
@soniaturcotte Bit of clarification on the colours. The intent is to merge the existing swatches into the tables but that needs a  a small code update so I suggested @minniessuyuhuang add the tables in the structure that's needed an they we'd work together to get the swatches in there and have the hex values come from the actual sass files. So consider this placeholder for now 👍

minniessuyuhuang:
> A few notes:
> 
> ## Colours:
> This is a bit confusing now, with the tables and then the swatches below. Particularly because the names in both places are different. A few suggestions:
> 
> * It would be good to see the colours in context in the table view.
> * If adding a preview of the colour isn't feasible in the table, then move the swatches to the top.
> * In the swatches/implementation view, some of the text labels are overlapping.
> * Name the colours the same thing in both places

I've moved the colour swatches to the top. Regarding the rest of the suggestion, they will require FEs to help.

> 
> ## Buttons
> Can we get the actual implementation of these tertiary button and "Next" "Previous" instead of screenshots?

@davidrapson Can we have this example in Storybook?

> 
> ## Radios
> * The heading to text ratio is a bit heavy, which makes it difficult to read. We should edit the text so it's clearer, possibly removing the headings.
Updated
> * Inline varient should show an inline example

@davidrapson can we have an example of it in Storybook please?

> * Remove "Varient:" from before each example, it's not necessary

Done

> * Add the same link to error summary for the error example (as in Inputs)

Done

minniessuyuhuang:
Updated the guidance based on Sonia's comments. Some of them will require the FEs help (see my comment above).

davidrapson:
@soniaturcotte @minniessuyuhuang Updated how the swatches are generated so we can do this in the code now:

```
<Description>
  {colourLanguageTable([
    ['Button', 'primary-button-colour'],
    ['Button hover', 'primary-button-hover-colour'],
  ])}
</Description>
```

Which lets us show the table like this:

colour-table

All updated in this PR now.

minniessuyuhuang:
Thanks @davidrapson ! I've also updated the 'Inline' radio example based on @soniaturcotte comment. There is one comment left to do which is about the button example. @davidrapson do you think you will have the capacity for this task in this or next sprint? If not, I'm suggesting that we should leave it for now. We can update this when we start building the decision tree.

soniaturcotte (approved):

PR #650, opened 2020-12-14
fix: make nav visited links white

davidsauntson:
Fix to 4.0.0-alpha.1 to make visited links in the global navigation white.


davidrapson (approved):
👍

PR #652, opened 2020-12-14
docs: contact details [NP-1028]

baseonmars:
Updated contact details documentation to use new format. 
Switched `strong` tag out for `b` - this matches the markup Contentful produces.


davidrapson (approved):
👍 Nice, also very happy that there is only one file left in the old component.stories.js. Nearly there!

PR #657, opened 2020-12-15
bugfix: Fix Breadcrumb tests for iPhone [NP=1253]

celonicolosi:
Some breadcrumbs scenarios were falling on the iPhone because breadcrumb behaves differently on responsive screens.
This PR fixes those problems by expecting different results for mobile_phones.

### DIFF
--- a/testing/features/breadcrumbs.feature
+++ b/testing/features/breadcrumbs.feature
@@ -10,7 +10,7 @@ Feature: Breadcrumbs component
     Scenario: Breadcrumbs only link up till the last one
       Then a series of breadcrumbs are present
       And the initial breadcrumbs are all links
-      But the final breadcrumb isn't a link
+      But the final breadcrumb isn't a link unless on mobile
 
   Rule: Site Wide Breadcrumb
     Background:
--- a/testing/features/breadcrumbs.feature
+++ b/testing/features/breadcrumbs.feature
@@ -1,4 +1,3 @@
-@not_mobile @NP-1253
 Feature: Breadcrumbs component
 
   The Breadcrumbs component allows users to see a navigable path to their
--- a/testing/features/breadcrumbs.feature
+++ b/testing/features/breadcrumbs.feature
@@ -7,10 +7,15 @@ Feature: Breadcrumbs component
     Background:
       Given a Default Breadcrumbs component is on the page
 
-    Scenario: Breadcrumbs only link up till the last one
+    @not_mobile
+    Scenario: On large devices breadcrumbs only link up till the last one
       Then a series of breadcrumbs are present
       And the initial breadcrumbs are all links
-      But the final breadcrumb isn't a link unless on mobile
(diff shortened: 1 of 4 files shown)


testing/features/step_definitions/breadcrumbs_steps.rb
```diff
@@ -3,6 +3,7 @@
 Given("a Default Breadcrumbs component is on the page") do
   @component = Breadcrumbs::Default.new
   @component.load
+  @default = true
```
luke-hill, line 6: This shouldn't be needed. We can either check the class, or do something else. But I'm fairly certain just the mobile_phone? check should be fine.

luke-hill (commented):
Could we add a tag - `@small_screen` to the bottom scenario. And have it do 1 of 2 things inside our hooks.

1) If mobile, nothing
2) If desktop, resize to small res?

testing/features/step_definitions/breadcrumbs_steps.rb
celonicolosi: The problem here is not the small_screen of mobile_phone. The problem is in the step "But the final breadcrumb isn't a link", that is not true for breadcrumb_default:
Screenshot 2020-12-16 at 17 38 21
So maybe we need to change the step to be like: But the final breadcrumb isn't a link unless on default? What do you think @luke-hill

luke-hill:
If the specs i.e. the physical specification is different on different variants, then that's fine. Reflect that in the tests. But yeh we should still test all variants where they are functionally possible (So instead of mobile only, make ti small screen only)

testing/features/breadcrumbs.feature
```diff
@@ -8,6 +7,15 @@ Feature: Breadcrumbs component
     Background:
       Given a Default Breadcrumbs component is on the page
 
+    Scenario: Breadcrumbs only link up till the last one
+      Then a series of breadcrumbs are present
+      And the initial breadcrumbs are all links
+      But the final breadcrumb isn't a link unless on mobile
```
luke-hill, line 13: lets remove this "unless on mobile" and just not assert. The previous cuke will validate all breadcrumbs are links.

testing/features/step_definitions/breadcrumbs_steps.rb
```diff
@@ -10,14 +10,31 @@
   @component.load
 end
 
+Given("a Site Wide Breadcrumbs component is on the page") do
+  @component = Breadcrumbs::SiteWide.new
```
luke-hill, line 14: can we tap load this and the other, to save space.

luke-hill (approved):
@davidrapson Are you ok with this too. I think this distinction is quite good (As we start using the concept of small screens, not just it's a mobile)

davidrapson:
Yep, definitely prefer thinking in terms of breakpoints, not devices (unless it's genuinely device specific)

PR #658, opened 2020-12-15
hotfix: Fix image caching on master branch

simonwgill:
Ugly fix to handle the container naming issue. When design-system builds container images on the master branch, it uses a image tag like dev_ruby_master. This tag doesn't exist in the repository, so should be rebuilt using the caching of the primary image (e.g. ruby). Because it wasn't getting removed, when a big change happened, each of the parallel tests had a chance of picking up the pre-change image from the last run - or rebuilding the post change image.

This adds some steps to ensure that after every run of the tests, these local container tags get removed. The layers and their hashes still exist, so the benefits of caching are still there, they just won't be used if they are supposed to have changed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -298,15 +298,18 @@ def withForcedDockerUpdate(images = [], local_images = [], Closure body) {
       sh "docker pull ${it}"
       sh "docker image rm ${it}_tmp"
     }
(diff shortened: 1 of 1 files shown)


baseonmars (approved):
Makes sense. this might fail the first time it's merged right - it cleans up after a test run so the first time this runs we could pick up the old image.

PR #659, opened 2020-12-15
Refactor: Rework Docker Image and Update Ruby [NP-1072]

luke-hill:
- Reworked Docker Image to be Ruby based
- Update Design System (& Associated Ruby tests), to `2.7`
- Update suite-wide default docker to `2.7.2`

Review Points:
- It's likely that (As I've copied from Public-Website), there may be 1 or 2 redundant / slightly sub-optimal things
- I'd prefer to try keep things that are "big changes" i.e. reworking huge portions; to a break-out ticket later on. As the purpose here was to update ruby which was proving tricky in the old format.


luke-hill:
EDIT: During playing with this. I found that each time you do a docker-compose build. It isn't caching the npm package install (It does cache the gem installs). Maybe we need to verbosely copy the node_modules as a separate `COPY` step as a breakout ticket? Or experiment in some other way

luke-hill:
I'm happy to do that, but do you want it in this PR or a follow-up PR. Just thinking of time that's all?

simonwgill:
> EDIT: During playing with this. I found that each time you do a docker-compose build. It isn't caching the npm package install (It does cache the gem installs). Maybe we need to verbosely copy the node_modules as a separate `COPY` step as a breakout ticket? Or experiment in some other way

@luke-hill from what I could see earlier this appears to be an issue with chown copying making the version in the image appear different from the source at all times - forcing everything afterwards to be invalidated.

simonwgill:
> I'm happy to do that, but do you want it in this PR or a follow-up PR. Just thinking of time that's all? I'm fairly certain I'll need to alter other bits / packages.

In this one, if you can. No point filling the ECR repository with bloated builds.

luke-hill:
Ok think that's all sorted now. Much smaller PR to review.

I can add some comments in certain places if needed?

docker/Dockerfile.ca-styleguide
```diff
@@ -1,23 +1,34 @@
...
-ENV RUBY_VERSION=2.6
-ENV LANG=C.UTF-8 \
+ENV NODE_MAJOR_VERSION=14 \
+    BUNDLER_VERSION=2.1.4 \
+    LANG=C.UTF-8 \
     BUNDLE_JOBS=6 \
     BUNDLE_RETRY=3
 
-RUN apk --no-cache --update add ruby~$RUBY_VERSION ruby-dev~$RUBY_VERSION ruby-json ruby-etc gcc libc-dev libxslt-dev libxml2-dev zlib-dev build-base ruby-bigdecimal git && gem install bundler -v $BUNDLER_VERSION
+RUN apk add bash curl gcc libc-dev libxslt-dev libxml2-dev zlib-dev build-base inotify-tools && \
```
simonwgill, line 9: Can you check what is still needed from this list of added things? I don't think any of them are needed now we're not installing ruby from source.

Also, this command should be using --no-cache.

docker/Dockerfile.ca-styleguide
```diff
@@ -1,23 +1,34 @@
...
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
```
simonwgill, line 10: What are these two lines for? Do we need to copy them over to other containers to keep up with best practice?

docker/Dockerfile.ca-styleguide
```diff
@@ -1,23 +1,34 @@
...
+    truncate -s 0 /var/log/*log
 
+RUN apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/main/ nodejs npm
```
simonwgill, line 13: Can you group up all the installs into a single RUN line and join them with &&? (pull the gem install up into it as well)

simonwgill (dismissed):

docker/Dockerfile.ca-styleguide
luke-hill: It was something mentioned online about doing the equivalent of an `apt-get clean` I also think we have/had it in public-website (the clean). Happy to remove if you want me to?

simonwgill: No, that's fine. I just wasn't sure what it was for.

luke-hill:
Ok hopefully now this is RtM again!

docker/Dockerfile.ca-styleguide
```diff
@@ -1,23 +1,37 @@
...
+    # Default repo is (3.12), which whilst latest only contains Node 12
+    # This new repo allows us to install Node 14 LTS
+    apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/main/ nodejs~$NODE_MAJOR_VERSION npm~$NODE_MAJOR_VERSION
+
+RUN addgroup --gid 1000 ruby && \
+    adduser --home /home/ruby --uid 1000 --ingroup ruby --disabled-password ruby
 
 RUN mkdir -p /app && chown -R 1000:1000 /app
-COPY --chown=node:node package*.json Gemfile* /app/
+COPY --chown=1000:1000 package*.json Gemfile* /app/
```
davidrapson, line 23: Out of interest why "1000" for the user id / group?

simonwgill (approved):
I've added a ticket to go back and check out the packages to make sure they aren't being added unnecessarily https://citizensadvice.atlassian.net/browse/NP-1487

davidrapson (commented):
👍 Final change for this look pretty minimal in the end which is nice. Flipping the primary image to Ruby makes sense to me but will leave approval to @simonwgill for the Docker specifics.

docker/Dockerfile.ca-styleguide
luke-hill: Not a clue, it works 🤷‍♂️

simonwgill: Technically it could be ruby:ruby, since the user/group is getting created with those ids, but kubernetes gets very bitchy about using user names in Dockerfiles instead of UIDs. So, to be consistent, it's just easier to use numbers throughout and 1000 seems to be the standard non-root user.

Also, using a non-root user for processes is a really good idea to prevent potential container escape issues.

PR #665, opened 2020-12-16
Add placeholders for targeted content translations

davidrapson:
Add placeholders for targeted content translations. By default Rails injects translation missing markup, which when used in attributes results in the following:

Screenshot 2020-12-16 at 13 34 27

Adding placeholders to cy.yml for now, have requested translations but imagine they will come in the new year so suggest merging this as is.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -168,6 +168,8 @@ GEM
       unicode-display_width (>= 1.4.0, < 2.0)
     rubocop-ast (0.8.0)
       parser (>= 2.7.1.5)
+    rubocop-rake (0.5.1)
+      rubocop
     ruby-progressbar (1.10.1)
     ruby2_keywords (0.0.2)
     rubyzip (2.3.0)
@@ -215,6 +217,7 @@ DEPENDENCIES
   rake
   retriable (~> 3.1)
   rspec
+  rubocop-rake
   selenium-webdriver (= 4.0.0.alpha6)
   site_prism (~> 3.7)
   webdrivers
--- a/Rakefile
+++ b/Rakefile
@@ -27,32 +27,25 @@ namespace :design_system do
     ) || raise
   end
 
+  desc "Creating report folders"
   task :report_folders do
     puts "Creating folder structure for this test run"
-    [
-      "testing/#{base_cucumber_path}/html_pages",
-      "testing/#{base_cucumber_path}/logs",
-      "testing/#{base_cucumber_path}/reports",
-      "testing/#{base_cucumber_path}/screenshots"
-    ].each { |dir| FileUtils.mkdir_p(dir) }
+    dirs = %w[html_pages logs reports screenshots]
+    dirs.each do |dir|
+      FileUtils.mkdir_p("testing/#{base_cucumber_path}/#{dir}}")
+    end
   end
 end
 
 desc "Run all checks in the local context"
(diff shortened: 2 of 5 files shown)


baseonmars (approved):
Looking good!

Given our use case do you think it would be worth setting up fallbacks in the rails app?

alternatively - we should have a script that compares the two files and reports any missing keys as part of CI.

davidrapson:
> alternatively - we should have a script that compares the two files and reports any missing keys as part of CI.

I'd be up for this. Normally you can set `raise_on_missing_translations` in development to catch this, which would be possible if we went [full engine](https://github.com/citizensadvice/design-system/pull/594), but couldn't get that working in our script context so a CI check would be a good safeguard.

davidrapson:
I've updated this PR with a basic lint script which compares the keys in both locale files and should error if anything is missing

Rakefile
```diff
@@ -18,6 +19,19 @@ def base_cucumber_path
   "artifacts/#{ENV['BROWSER']}/#{ENV['BROWSERSTACK'] == 'true' ? ENV['BROWSERSTACK_CONFIGURATION_OPTIONS'] : 'other'}"
 end
 
+def compare_yaml_hash(yaml_a, yaml_b, context = [])
+  yaml_a.each do |key, value|
```
luke-hill, line 23: If i'm understanding here, you're wanting to validate the nested structure of the yaml, but only validate the keys right?

davidrapson: Yep, want to validate that all the keys match. But not the values obviously as they are translated. Not looking for a perfect check, only a little safeguard. Simpler ways to do this welcome.

davidrapson:
Hmm this _does_ fail as expected but `collect_task_errors` seems to swallow the actual error message in CI

Rakefile
luke-hill: Actually it seems quite good. Although I saw it's a SO answer that you've based it off. Only thing I got confused by was the name. It's definitely a **very good** way of using clever recursion. I would have suggested `hashdiff` but that computes value difference too using diff-lcs as an algorithm (Which you obviously won't want), so this is a good solution.

If it was me I'd call it something like `ensure_identical_structure(yml_a, yml_b, context = [])` but that's definitely just a personal opinion. Not something I'm fussed on too much

Gemfile.lock
```diff
@@ -215,6 +217,7 @@ DEPENDENCIES
   rake
   retriable (~> 3.1)
   rspec
+  rubocop-rake
```
luke-hill, line 220: This is unlikely to run by default, it'll probably need configuring in the ca style repo proper. Or patched in to our config in this repo inside the `.rubocop.yml` configuration file

NB: Disclaimer (I've never used this extension, so I could be wrong!)

Rakefile
davidrapson: [looks around shiftily] definitely did~n't~ copy paste this guvner.

Gemfile.lock
davidrapson: There's a `require: rubocop-rake` in the .rubocop.yml if you missed it. Caught a couple little things and as I was adding more to the rakefile it felt helpful.

davidrapson:
Closing in favour of https://github.com/citizensadvice/design-system/pull/681

PR #667, opened 2020-12-16
feature: global nav design review [NP-1026]

davidsauntson:
Brings the 'more' drop down in global navigation in line with the designs:
 - on `sm`, the drop down is full width
 - on `md` and `lg`, the drop down is aligned to the right of the viewport
 - spacing and padding corrected as per designs


scss/2-tools/_interactive-elements.scss
```diff
@@ -15,3 +15,40 @@
     @include cads-interactive-focus();
   }
 }
+
+@mixin cads-nav-link-focus() {
+  background-color: $cads-palette__white;
```
baseonmars, line 20: can these be swapped out for a `$cads-language` variable to make theming a bit easier? Perhaps out of scope.

scss/2-tools/_interactive-elements.scss
```diff
@@ -15,3 +15,40 @@
...
+  color: $cads-language__brand-primary;
+  border: $cads-border-width-medium solid $cads-palette__heritage-yellow;
+  outline: none;
+}
+
+@mixin cads-nav-link() {
+  font-size: 1rem;
```
baseonmars, line 27: Some opportunity to use more vars here and on line-height?

scss/6-components/_navigation.scss
```diff
@@ -80,25 +55,28 @@
...
       display: block;
       content: '';
       border-top: $cads-border-width-small solid $cads-palette__white;
       position: relative;
-      left: $cads-spacing-4;
-      width: calc(100% - #{2 * $cads-spacing-4});
+      // take focus border into account
+      left: cads-rem-to-px($cads-spacing-4) + $cads-border-width-medium;
+      width: calc(
+        100% - #{2 * $cads-spacing-4} - #{2 * $cads-border-width-medium}
```
baseonmars, line 67: TIL

src/ts/greedy-nav/Config.ts
```diff
@@ -69,7 +69,7 @@ const defaultConfig: Config = {
   navDropdownBreakpointLabel: 'menu',
   breakPoint: 0,
   throttleDelay: 50,
-  offsetPixels: 0,
+  offsetPixels: -2,
```
baseonmars, line 72: Why does it need this?

baseonmars (commented):
Gonna check in browser - but a few minor comments for you to chew on.

baseonmars (approved):
checked in browser :+1:

scss/2-tools/_interactive-elements.scss
davidsauntson: `$cads-language__brand-primary-background`

src/ts/greedy-nav/Config.ts
davidsauntson: added comment in config setting

PR #668, opened 2020-12-16
Adviser targeted content variant [NP-1456]

davidrapson:
New adviser specific variant of targeted content. Needed a bit of trial an error to make sure that the badge didn't break heading / button semantics, and that it continued to work with reference buttons so there's a little refactoring going on to help keep the implementation simple.

Example showing embedded table:

image

Screenshot showing this running in public-website whilst logged in:

image


davidrapson:
If the general approach looks good then I'm happy to pull out the callout / badge refactoring into a separate PR.

baseonmars (approved):

PR #670, opened 2020-12-17
bugfix: Skip Links [NP-1396]

luke-hill:
This is now fixed same as in public-website

We use option+tab to skip through them on safari. They are still not enabled for mobile as this concept is irrelevant there


celonicolosi (approved):
LGTM! 👍

PR #672, opened 2020-12-17
Feature/np 1461 header links

davidsauntson:
Extend the header haml to allow a `current_site` bool, that renders a span instead of a link in the header links section.


scss/6-components/_navigation_greedy-nav.scss
```diff
@@ -51,6 +52,10 @@
       }
     }
 
+    span {
```
davidsauntson, line 55: a conscious choice so as not to have styling classes in the GreedyNav typescript

scss/6-components/_search.scss
```diff
@@ -75,3 +75,18 @@
     }
   }
 }
+
+/* stylelint-disable plugin/selector-bem-pattern */
+.cads-header__search-row {
```
davidrapson, line 80: If possible we should avoid touching other component classes, could we add a `cads-header__search-form` wrapper so we could write:

```
.cads-header__search-form {
    margin-left: auto;

    @include cads-media-breakpoint-only(md) {
      width: 70%;
    }

    @include cads-media-breakpoint-only(lg) {
      width: 55%;
    }
}
```

Avoids the need for the `stylelint-disable` then

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **New**
 
 - 📋 Footer Feedback: add a class to allow selection of footer feedback link
+- 🔗 Header: Links in the header now accept a `current_value` hash property to render a span instead of a link
```
davidrapson: Should `current_value` be `current_site`?

davidrapson (approved):
👍 Looks good, with a docs change and a question about BEM styles for the search form

scss/6-components/_search.scss
davidsauntson: well that's a good idea 👍

davidrapson:
🚀

PR #673, opened 2020-12-17
fix: padding in more dropdown

davidsauntson:
- removes padding added by @minniessuyuhuang around the 'More' dropdown `ul`, as requested by @soniaturcotte 
- corrects top and bottom padding on list items in the more dropdown


davidrapson (approved):
👍 Big stuff

PR #675, opened 2020-12-18
Extract badge styles from callout component

davidrapson:
Extracted from https://github.com/citizensadvice/design-system/pull/668

**New badge styles**

Extracts out shared badge/label styles into a `cads-badge` class. Went with badge instead of label to avoid confusion with form labels.

Badges being this mini-component:

| Basic badge | Adviser badge |
| ------- | ------- |
| image | image |

We can then reuse this for callouts and adviser targeted content. Has the bonus of simplifying the callout styles.

**Callout variants**

Whilst refactoring the callouts styles I updated the modifiers to use consistent BEM conventions: e.g. cads-callout--adviser.


davidsauntson (commented):
Nice - there is a reduction in space between the new label and the callout title now.  Is this in line with the orig designs?

davidsauntson (approved):

davidrapson:
> there is a reduction in space between the new label and the callout title now. Is this in line with the orig designs?

Good eye, looks like I was overzealous with my refactoring. Just fixed that up.

PR #696, opened 2021-01-04
Support callouts within callouts

davidrapson:
image

- Nested callout styles, sit flush right when nested
- Add nested callouts example and docs
- Add visual regression tests
- Update changelog
- Remove a redundant style declaration

These were largely covered already, with the main change to sit flush right when nested and to tidy up margins.


scss/6-components/_callout.scss
```diff
@@ -17,10 +17,6 @@
     @extend %cads-h3;
   }
 
-  @include cads-media-breakpoint-up(md) {
-    padding: $cads-spacing-4;
```
davidrapson, line 21: Removed this as it's the same as the default padding.

baseonmars (approved):
👍

PR #702, opened 2021-01-06
feature: Automation - Review language specific tests [NP-1328]

luke-hill:
I have gone through the entire automated tests repo. I have added tests for each component that is language specific, amended existing ones, and flagged ones not fully working.

### DIFF
--- a/haml/_footer.html.haml
+++ b/haml/_footer.html.haml
@@ -29,5 +29,5 @@
         .cads-footer__logo
           %a.cads-logo{ href: homepage_url, title: t("cads.shared.logo_title") }
         .cads-footer__meta
-          %p.cads-footer__meta-text= t("cads.shared.copyright")
-          %p.cads-footer__meta-text= t("cads.shared.legal_summary")
+          %p.cads-footer__meta-text{ data: { test_id: "copyright" }}= t("cads.shared.copyright")
+          %p.cads-footer__meta-text{ data: { test_id: "legal summary" }}= t("cads.shared.legal_summary")
--- a/haml/_footer.html.haml
+++ b/haml/_footer.html.haml
@@ -29,5 +29,5 @@
         .cads-footer__logo
           %a.cads-logo{ href: homepage_url, title: t("cads.shared.logo_title") }
         .cads-footer__meta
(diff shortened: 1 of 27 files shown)


haml/_footer.html.haml
```diff
@@ -29,5 +29,5 @@
         .cads-footer__logo
           %a.cads-logo{ href: homepage_url, title: t("cads.shared.logo_title") }
         .cads-footer__meta
-          %p.cads-footer__meta-text= t("cads.shared.copyright")
-          %p.cads-footer__meta-text= t("cads.shared.legal_summary")
+          %p.cads-footer__meta-text{ data: { test_id: "copyright" }}= t("cads.shared.copyright")
+          %p.cads-footer__meta-text{ data: { test_id: "legal summary" }}= t("cads.shared.legal_summary")
```
davidrapson, line 33: Could we use `data-testid` rather than `data-test-id`. That way we're consistent with the conventions used in https://testing-library.com/docs/dom-testing-library/api-queries#bytestid for our DOM tests. And mind making the values hyphenated, e.g "legal-summary", rather than including spaces?

davidrapson (changes requested):
Minor change

I have more general question though. I don't _really_ understand why we use I18n directly in our tests. Doesn't this make the check circular? Like if we accidentally include some duff text in the cy.yml this test would still pass, right? I might be missing some important conceptual difference though.

haml/_footer.html.haml
luke-hill: Will update the code. And yes to your question, it does make it semi-circular. However this in practice is good with very simple static stuff like this, because eventually down the line we should be getting to me reviewing all tickets going into production (if not in code, just testing them). That way I could see if the logic was poor.

As for whether this would trap a poor label, the answer is no it wouldn't. But having 2 sets of yml files wouldn't either, because we would just update both of them simultaneously. That kind of logic is a bit untestable.

davidrapson: Interesting. I don't entirely follow the notion of having a manual gate keeper vs relying on static tests but I think I get what you mean. Nothing blocking though as we seem to be using this I18n pattern across the board anyway.

davidrapson (approved):

PR #712, opened 2021-01-08
feature: Targeted content [NP-1185]

luke-hill:
Adds in the spec for Targeted Content.

NB: Cannot validate the "scrolling" behaviour, as it doesn't freeze or mutate the DOM representation of the element, but we validate the auto-open or not.

NB2: Minor issue in styleguide code, have amended and fixed (Correspondingly, the tests affected are tagged as future release)


luke-hill:
@davidrapson there's a bug both on the en and cy locales, where some of the labels come through as missing. But they're present in the locale files. I had a little look but didn't get very far. Should I create a breakout for these, or is there a quick fix I'm not seeing.

baseonmars (approved):

PR #734, opened 2021-01-13
fix: apply focus state to active pagination page

davidrapson:
https://citizensadvice.atlassian.net/browse/NP-1490

> Missing focus state on active page on pagination when tabbing.

Turns out this was exactly what it sounds like. Active pages were missing a link focus state.

Screenshot 2021-01-13 at 11 08 05


davidsauntson (approved):
👍

PR #738, opened 2021-01-15
Fix/np 1504 header accessibility review

davidsauntson:
The accessibility bug was caused by the `aria-hidden` attribute remaining `true` when the more dropdown is first opened.  The code contained duplicated code for opening and closing the menu which was slightly different in different circumstances.

This PR:
 - creates `openDropDown()` and `closeDropDown()` methods
 - refactors existing code to use these methods where appropriate
 - adds test to support new methods
 - changes tab opening listener to `keyup` to simplify tab listener

### DIFF
--- a/src/ts/greedy-nav/GreedyNav.test.ts
+++ b/src/ts/greedy-nav/GreedyNav.test.ts
@@ -286,7 +286,7 @@ describe('Greedy Nav', () => {
     });
 
     it('toggles the menu open', () => {
-      const event = new dom.window.FocusEvent('focus');
+      const event = new dom.window.KeyboardEvent('keyup', { key: 'Tab' });
       nav.navDropdownToggle!.dispatchEvent(event);
 
       expect(nav.navDropdown!.className).toContain('show');
--- a/src/ts/greedy-nav/GreedyNav.test.ts
+++ b/src/ts/greedy-nav/GreedyNav.test.ts
@@ -1,6 +1,8 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 /* eslint-disable import/no-extraneous-dependencies */
 import '@testing-library/jest-dom';
+import { fireEvent, createEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { JSDOM } from 'jsdom';
 import path from 'path';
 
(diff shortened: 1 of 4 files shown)


src/ts/greedy-nav/GreedyNav.test.ts
```diff
@@ -286,7 +287,7 @@ describe('Greedy Nav', () => {
     });
 
     it('toggles the menu open', () => {
-      const event = new dom.window.FocusEvent('focus');
+      const event = new dom.window.KeyboardEvent('keyup', { key: 'Tab' });
       nav.navDropdownToggle!.dispatchEvent(event);
```
baseonmars, line 291: Does testing libraries `fireEvent` work here? I feel like I wrote this in a weird way - not sure why.

src/ts/greedy-nav/GreedyNav.test.ts
```diff
@@ -312,4 +313,58 @@ describe('Greedy Nav', () => {
...
+      nav.init();
+    });
+
+    afterEach(() => {
+      document.body.innerHTML = '';
+    });
+
+    describe('openDropDown', () => {
+      it('opens the dropdown menu', () => {
+        nav.openDropDown(nav.mainNavWrapper);
```
baseonmars, line 340: Any reason not to trigger this the same way the user does (via click or keyboard press)?

src/ts/greedy-nav/GreedyNav.test.ts
```diff
@@ -312,4 +313,58 @@ describe('Greedy Nav', () => {
...
+      it('sets aria-hidden to false on the drop down', () => {
+        nav.openDropDown(nav.mainNavWrapper);
+
+        expect(nav.navDropdown!).toHaveAttribute('aria-hidden', 'false');
+      });
+    });
+
+    describe('closeDropDown', () => {
+      beforeEach(() => {
+        nav.openDropDown(nav.mainNavWrapper);
```
baseonmars, line 354: Same as above

baseonmars (approved):
Much cleaner. great improvement. I left some comments re the tests but nothing blocking.

src/ts/greedy-nav/GreedyNav.test.ts
davidsauntson: yup - all changed

PR #753, opened 2021-01-20
Set max-width on images inside prose areas [NP-1610]

davidrapson:
Ensures that an inline images have a flexible max-width to allow them to scale down on small screens. Also sets display block on images to avoid double margins.

Whilst updating this I noticed that the local image used in the same page results in a 404 when storybook is deployed as it doesn't know how to reference it. Replacing it with a placeholder.com URL


davidsauntson (approved):
👍 Does this need a vrt?

davidrapson:
Probably, noticed we don't have one for the sample pages. I can add one to this PR.

PR #757, opened 2021-01-21
[Experiment] parallel test stages

davidrapson:
Just checking if this works to run unit and lint stages in parallel. Consider this throwaway.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,23 +93,26 @@ def pipeline() {
     }
   }
 
-  stage('Lint') {
-    withDockerSandbox([ images['ca-styleguide'], images['ruby'] ]) {
-      withEnv(['PRODUCTION=true', 'NODE_ENV=test']) {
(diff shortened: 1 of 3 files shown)


davidrapson:
Nope…or rather the unit test step runs, but the lint fails because it needs selenium browsers (??)

davidrapson:
Parallel passed the second time, but also going to try removing the extra image from the lint step as I can's see why it would be needed.

luke-hill:
I've got some tickets for these I can try help with. They're in the backlog. The issue here is mixing declarative and scripted pipelines. We've migrated in this repo to a pure scripted pipeline.

davidrapson:
Gotcha, this was a bit of a "what I can try in 1/2 hour" test as build times are pretty painful at the mo, was instructive at least as I can see a few optimisations we can make even outside of parallelising (the fact we run `npm build` three times jumped out). But will close this for now and we can revisit.

PR #758, opened 2021-01-21
Fix: print styles [NP-1555]

davidsauntson:
Updates print styles for components covered by the sample pages. In particular:

 - updates to scss for print styles
 - add VRT scenarios for sample pages (mostly to ensure general typography is covered)
 - add VRT scenarios for sample pages with media set to `print` (puppeteer has a function for this)
 
It's worth noting that for some elements (esp tables), the actual printed version looks quite different from the `media print` styles you see in the VRT images.

@davidrapson some fun table pics for ya
**Print emulated long tables**
image

**Actual printed long tables**
image


scss/_print.scss
```diff
@@ -4,18 +4,54 @@
   .cads-error-summary,
   .cads-navigation,
   .cads-success-message,
-  .cads-related-content {
+  .cads-related-content,
+  .cads-skip-links {
     display: none !important;
   }
 
   // Components that adapt for printing
```
davidrapson, line 12: Personal preference is to have a `@media print` block in the same file as the main component styles. But don't feel strongly about that, so this works too.

testing/visual-regression/backstop-common.js
```diff
@@ -32,6 +32,21 @@ module.exports = function backstopCommon(baseUrl) {
     ];
   }
 
+  function buildSamplePageScenarios(labelPrefix, componentId) {
+    return [
+      {
+        label: `${labelPrefix} (Print)`,
+        url: storyUrlFor(componentId),
+        onReadyScript: 'onReadyEmulatePrint.js',
```
davidrapson, line 40: We might want to override the viewport here for a single value that approximates an A4 page. Currently this is generating small screen versions of print styles which is a bit confusing.

davidrapson (approved):
👍 I reckon we want to tweak the viewport setting for the print test but this looks great. Approving though as that's not blocking.

Out of interest what's the major difference with tables between emulated and real printing?

scss/_print.scss
davidsauntson: Yeah I think what you're suggesting is a better approach

davidsauntson: going for a specific print query at the bottom of the file rather than dotted throughout

PR #763, opened 2021-01-22
test: increase thresholds for footer and header

davidsauntson:
I'm not having any luck replicating the vrt failures locally, despite trying:
 - npm clean install of main and vrt package.jsons
 - deleting docker containers and rebuilding from images locally and running the vrts with those
 - reverting the first PR merge after the last green build overnight
 - deleting all the reference images and rerunning the tests in CI with those

Not really sure why they are failing - perhaps a different version of puppeteer in being used in CI?  But npm ci would have flagged that I think.

Anyway, upping the threshold on the failing tests which relate to the svg in the header and footer.  Any other ideas welcome.


baseonmars (approved):
this seems reasonable until we have time to dig into this.

PR #768, opened 2021-01-25
No need to run build as a prerequisite  to eslint or jest tasks

davidrapson:
Small build improvement I spotted whilst doing this experiment: https://github.com/citizensadvice/design-system/pull/757

Remove `build` dependency from `npm:lint` and `npm:jest` step. It shouldn't be needed and results in us doing unnecessary builds.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -73,13 +73,13 @@ namespace :npm do
   end
 
   desc "Run jest tests in node"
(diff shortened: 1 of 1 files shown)


luke-hill (approved):
If this passes that's a great improvement.

PR #769, opened 2021-01-25
Fix/np 1555 print styles 2

davidsauntson:
Update print styles to:
 - print `href` after `a` tags
 - exclude adviser callout from print styles


scss/5-objects/_prose.scss
```diff
@@ -57,6 +57,20 @@
...
+}
+
+@media print {
+  a[href]::after {
+    content: (' (' attr(href) ')');
+    display: inline;
+  }
+
+  a[href^='/']::after {
+    content: (' (https://www.citizensadvice.org.uk' attr(href) ')');
```
davidrapson, line 73: Not sure I follow this line? Will this prepend `https://www.citizensadvice.org.uk` to any links starting with `/` i.e. internal?

davidrapson (approved):
👍 Small query about the link text behaviour, but looks good otherwise.

scss/5-objects/_prose.scss
davidsauntson: That's right - currently we just render relative links, so when they are printed they aren't much use.  This just adds that to the start of any internal links when a page is printed.  Shall I pop a comment in?

davidrapson: Might be worth a quick comment to that effect, ta 👍

PR #770, opened 2021-01-25
feat: header sign out link can be a form [NP-1604]

baseonmars:
Adds support for rendering the sign out link as a form.

* Add a `cads-linkbutton__regular` class as link buttons are usually bold
* Add `form` boolean parameter to header links
* Add styles for mobile dropdown
* Fix styling so header links displayed in dropdown stretch full length of the dropdown (previously had 1rem left padding which differs from other navigation links)
* Update docs to include link button styling.

Screenshot 2021-01-25 at 14 35 40
should look the same as before

Screenshot 2021-01-25 at 15 31 31

Mobile dropdown styling

### DIFF
--- a/haml/_header.html.haml
+++ b/haml/_header.html.haml
@@ -29,7 +29,11 @@
 
           - if sign_in_out
             %li.cads-header__account-link
-              %a.cads-header__hyperlink{href: sign_in_out["url"], "data-testid": "account-link"}= sign_in_out["title"]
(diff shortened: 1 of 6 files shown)


davidsauntson (approved):
LGTM - the nav links shouldn't be right-aligned in the drop down, but that isn't part of this PR so I'll make sure it's captured in JIRA.

PR #771, opened 2021-01-25
feature: Disable Mobile testing for Design-System [NP-1591]

luke-hill:
Temporary measure for 3-4 weeks whilst we ...

- Sign our new BS contract
- Migrate all users over to new "potting" system
- Work out how to get manual runs working again
- Fix up VAULT access for Dual Access API keys

Then we can re-enable the testing on iPhones


celonicolosi (approved):
LGTM 👍 
Visual Regression is falling but I don't think it is related to this simple change.

PR #775, opened 2021-01-26
test: update BackstopJS and re-reference VRTs

baseonmars:
VRTs were failing in CI but passing locally. CI is using old reference images.

Fixed by upgrading to backstopjs 5.1, which uses puppeteer 5 instead of 2.

I've re-referenced all images as there are subtle but consistent differences between SVG rendering.


davidrapson (approved):
👍 LGTM

PR #778, opened 2021-01-27
Fix: close menu on last item blur [NP-1624]

davidsauntson:
Restores the close on blur functionality for the last item in the greedy nav dropdown. 

 - updates header link copying to include a `js-cads` class later used to identify the last item in the copied header links (handles both `a` and `button` elements for Sign out
 - calculates the last item in the dropdown based on whether the copied links are visible or not 
 - update test fixtures to include the copied links


baseonmars (approved):
LGTM

PR #781, opened 2021-01-27
Fix: h4s in callouts [NP-1630]

davidsauntson:
Update heading styles in callout so they are aligned with the design/content decision documented here:
https://citizensadvice.atlassian.net/wiki/spaces/NP/pages/2627371014/Visual+appearance+of+headings+in+callouts


minniessuyuhuang:
Thanks David! I've had a look, the headings in the build are not aligned with the diagram attached: 

image

image

The first heading in the nested adviser callout is `h3` instead of `h4`? 

Also, the other heading in the adviser callout is `h4` but it doesn't have the h4 styling, is that intended?

davidsauntson:
> Thanks David! I've had a look, the headings in the build are not aligned with the diagram attached:
> 
> The first heading in the nested adviser callout is `h3` instead of `h4`?
> 
> Also, the other heading in the adviser callout is `h4` but it doesn't have the h4 styling, is that intended?

Thanks Minnie - I've updated by PR comment to hopefully make it a bit clearer.  This change was to make sure that any `h4` inside a callout gets the correct visual style (we discussed briefly and agreed that the `h4` should look like `h3`s - if I got the wrong end of the stick let me know!) I will add a specific example of the title being demoted to an`h4` too.

minniessuyuhuang:
> > Thanks David! I've had a look, the headings in the build are not aligned with the diagram attached:
> > The first heading in the nested adviser callout is `h3` instead of `h4`?
> > Also, the other heading in the adviser callout is `h4` but it doesn't have the h4 styling, is that intended?
> 
> Thanks Minnie - I've updated by PR comment to hopefully make it a bit clearer. This change was to make sure that any `h4` inside a callout gets the correct visual style (we discussed briefly and agreed that the `h4` should look like `h3`s - if I got the wrong end of the stick let me know!) I will add a specific example of the title being demoted to an`h4` too.

Ah I see, sorry! Yes, I'm fine with styling h4 the same as h3 (just for the headings in the callouts?) as long as the content designers are fine with it (since you can't tell the content hierarchy by the visuals).

davidsauntson:
@minniessuyuhuang OK - turns out I may have just made the requirement for h4s to look like h3 up - not sure which recess of my mind that came from 💫 Now this change just makes sure if the callout title is an h4, it is displayed correctly.

davidrapson (approved):
👍 Latest changes look good

PR #782, opened 2021-01-28
fix: link button focus styles

baseonmars:
This fixes a visual issue with link buttons:

Screenshot 2021-01-28 at 17 41 54

Screenshot 2021-01-28 at 17 33 03


davidrapson (approved):
I would like every fix to come with big 👇 👉  pointing arrows from now on 😄

PR #783, opened 2021-01-29
Feature/np 1589 heading levels targeted content

davidsauntson:
Allows clients of design-system to specify the heading level of the title element in targeted content components.  Defaults to a heading level of 2 if none provided.  Coerces to a integer and clamps between 1 and 6.


davidrapson (approved):
👍 Clever use of `haml_tag` here!

PR #799, opened 2021-02-02
feat: responsive table without Javascript [NP-1430]

baseonmars:
* Deprecate table javascript
* Insert headers into table cells using ruby rather than javascript
* Update VRTs (it seems like it didn't work before)

Screenshot 2021-02-02 at 15 26 38

Screenshot 2021-02-02 at 15 26 45

Screenshot 2021-02-02 at 14 35 13
_deprecation warning in action_

Screenshot 2021-02-02 at 15 20 42
_vrts being corrected_


davidrapson (approved):
Did this not even need any CSS changes in the end, only how the markup is generated? Nice!

Really like the deprecated pattern, we should do the same for https://github.com/citizensadvice/design-system/blob/master/src/js/callout.js (which is never initialised in public website and obsoleted by the heading renderers, but good to deprecate it kindly too)

styleguide/components/tables/tables.stories.js
```diff
@@ -1,10 +1,8 @@
 import docs from './tables-docs.mdx';
 import template from './_table.html.haml';
 import { translate } from '../../story-helpers';
-import initTables from '../../../src/js/tables';
 
 function decorateWithBehaviour(Story) {
```
davidrapson, line 5: This whole `decorateWithBehaviour` can go now too

src/js/tables.js
```diff
@@ -1,35 +1,11 @@
...
-          console.warn(`Error ${e} initialising table ${table}`);
-        }
-      }
-    }
-  } catch (e) {
-    console.warn(`Could not initialise tables. ${e}`);
-  }
+  // eslint-disable-next-line no-console
+  console.info(
+    'deprecation warning: initTables is no longer required and will be removed in future versions.'
```
davidsauntson, line 7: <3

davidsauntson (approved):
Lush.

Super nit pick, if you fancy fixing whilst you're here 😄  The table heading (caption?) is the same colour as the rest of the table in figma: https://www.figma.com/file/J9qfbS9n1pUM2dpRQnTXUD/Design-system-2.0?node-id=9%3A182

PR #807, opened 2021-02-03
fix: breadcrumb layout when collapsed [NP-1579]

baseonmars:
* Adjust the layout of breadcrumbs when collapsed so text is aligned with itself rather than chevron
* Add breadcrumb docs featuring many long titles
* Change the layout of no-collapse mobile variant to match up with the collapse variant (agreed with @minniessuyuhuang)

Screenshot 2021-02-03 at 09 34 35
Screenshot 2021-02-03 at 09 37 02
Screenshot 2021-02-03 at 09 34 18


davidsauntson (approved):
🍞 🥯 🥖

baseonmars:
> 🍞 🥯 🥖

spent most of my time picking which emoji to use in the changeling.

PR #808, opened 2021-02-03
Fix/targeted content margins

davidsauntson:
Adds targeted content components to the list of things that give subsequent headings the correct margin.


baseonmars (approved):
Space to breath. so nice :+1:

scss/2-tools/_typography.scss
```diff
@@ -28,7 +28,8 @@
   .cads-prose img + &,
   .cads-prose details + &,
```
davidrapson, line 29: Can we remove `.cads-prose details` at the same time (that's the _old_ targeted content)

davidrapson (approved):
👍 with a little tidy up request

scss/2-tools/_typography.scss
davidsauntson: nice catch!

PR #831, opened 2021-02-11
Fix: NP-1699 scroll to top of targeted content on close

davidsauntson:
Targeted contents do not alter the window's scroll position, which means when you close a long targeted content via the 'Close' button at the bottom, lots of content is skipped.  The epi site scrolls the user to the top of the targeted content on close, this PR replicates that behaviour.

### DIFF
--- a/src/js/targeted-content.js
+++ b/src/js/targeted-content.js
@@ -136,7 +136,6 @@ function initTargetedContentFor(el) {
       const newScrollY = window.pageYOffset + elTop - FAKE_MARGIN;
       const supportsSmoothScroll =
         'scrollBehavior' in document.documentElement.style;
-      console.log(supportsSmoothScroll);
       if (supportsSmoothScroll) {
         window.scrollTo({ left: 0, top: newScrollY, behavior: 'smooth' });
       } else {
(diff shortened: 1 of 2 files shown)


src/js/targeted-content.js
```diff
@@ -128,6 +128,14 @@ function initTargetedContentFor(el) {
     if (matchEl) {
       setState(matchEl, 'closed');
     }
+
+    const elTop = el.getBoundingClientRect().top;
```
davidrapson, line 132: Does something like `el.scrollIntoView({ behavior: "smooth", block: "center" })` work instead of needing to calculate a specific position?

davidsauntson: Probably - was going for something that doesn't need another polyfill.

davidrapson: Cool, as far as I know that shouldn't need a polyfill. I think it's the `smooth` behaviour that isn't always available, but that's true with `scrollTo` as well I think? Looks like calling `scrollIntoView` on the button / heading of targeted content might do the job. Crack on if I've misunderstood though.

davidsauntson: Just went for boring old feature detection.  I think the niceness of smooth scrolling and the extra bit of space between the top of the targeted content is worth it.

baseonmars (approved):

PR #856, opened 2021-02-19
chore: changelog parser and update tool

baseonmars:
Tool for parsing and updating the design system CHANGELOG.

usage is: 

```javascript
const fs = require('fs');
const { release } = require('./scripts/changelog');

const changelog = release('v1.1.1');
fs.writeFileSync('CHANGELOG.md', changelog);
```

This isn't tied into the release script yet - the release script has so many checks that it makes testing complicated. Will move away from checking in generated code in future releases which will make testing a whole lot easier.


scripts/changelog.js
```diff
@@ -0,0 +1,151 @@
...
+      i += 1;
+    }
+  }
+
+  return items.length ? `${items}\n` : '';
+};
+
+const findSubHeadings = (tokens) => {
+  const headings = new Set();
+  const regex = /\*\*([a-zA-Z ]+)\*\*/;
```
davidsauntson, line 77: I noticed that sometimes we use:
`**New**`
and sometimes:
`**New:**`

davidsauntson (approved):
❤️ 

Question around the conventions this has enshrined in code, mostly:
 - not having a heading for unreleased changes (we all do this anyway so fine)
 - deciding on `New:` and `New` (I do not append  `:`, others do)

Is it worth noting these in the contributing.md file for convenience?  Not a blocker just something to consider for the follow on tasks.

scripts/changelog.js
baseonmars: and also sometimes `**New**:` - I'm thinking a lint tool for the changelog might be a better option - thoughts?

scripts/changelog.js
```diff
@@ -0,0 +1,151 @@
...
+
+const assert = require('assert');
+const semver = require('semver');
+const fs = require('fs');
+const path = require('path');
+const marked = require('marked');
+const moment = require('moment');
+
+// TODO stronger check for version heading?
+const getPreviousVersion = (tokens) =>
```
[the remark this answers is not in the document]
baseonmars: At the time of execution package.json will contain the next version as that is what we'll use to trigger the release process.

scripts/changelog.js
davidsauntson: Yeah good shout.  If we are deciding here I prefer the colon-less version 🍰

baseonmars: yeah, I think so too. Could also change it to a heading instead of some bold text - but no need to change too much. I'll look into how easy it is to write lint rules and/or prettier extension.

PR #867, opened 2021-02-22
fix: scroll targeted content to center of viewport

davidsauntson:
Fixes issue where scrolling viewport to align with top of targeted content on close resulted in an undismissed OISC warning covering the targeted content.

Just using `scrollIntoView` instead for simplicity and robustness (can't really be sure what other consumers of the design system will be rendering at the top of the viewport).


davidrapson (approved):
👍 Looks good to me!

PR #868, opened 2021-02-22
Add Ca testing patches - Remove duplicate code

luke-hill:
Copy of PR 981 on public-website.

Add patches that exist in gem framework and remove duplicate code.


testing/features/support/core_ext/capybara/node/element.rb
```diff
@@ -1,31 +1,23 @@
...
-    module Node
-      module Element
-        def text(type = nil, normalize_ws: ::Capybara.default_normalize_ws)
-          super(type, normalize_ws: normalize_ws)
-        end
-
-        def color
-          native.css_value("color")
-        end
+module Capybara
```
davidrapson, line 3: More out of interest, but how come these are going into this shared gem, I can only see `vertical_position` being used in our code and it's particular to the OISC warning so wondering if that one helper couldn't just be inlined.

(This is probably my javascript background showing but having lots of core extensions to things feels very unusual to me. I know this is more common in Ruby land)

luke-hill: These ones aren't being moved yet. but they will be.

If we aren't using the other ones we can delete them (This code hasn't been very well maintained, so you could be right in that only one is used and in-line in one place). I'll need to double check, but we could possibly delete a load of them as well

For other examples, we're using them in both code-bases equally, so having them shared removes the need to duplicate. It will also help further if other repos like Casebook want to use them (Or maybe a 3rd one from us?)

davidrapson: Ah gotcha! I was totally misunderstanding what this was doing, most of this is just whitespace 🙈

luke-hill: Mini tip if you need it: 

Add the suffix here

https://github.com/citizensadvice/design-system/pull/868/files?diff=unified&w=1

davidrapson (approved):

PR #879, opened 2021-02-23
bugfix: Browserstack Allocation Issue [NP-1752]

luke-hill:
Find out / fix browserstack dynamic allocation issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -153,9 +153,9 @@ def pipeline() {
     }
   }
 
-  if (!isRelease) {
-    stage('Regression Tests') {
-      parallel define_regression_tests()
+  if (isRelease) {
+    stage('Browserstack Tests') {
+      parallel define_browserstack_tests()
     }
   } else {
     stage('Grid Tests') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -266,15 +266,15 @@ def define_grid_tests() {
...
+def define_browserstack_tests() {
+  browserstack_tests = [:]
+  browserstack_tests.failFast = false
 
   configurationTypes.each {
     def (config, browser) = it
     def stepName = "${browser} on ${config}"
-    isMobile = (browser == "ios" || browser == "android")
-    regression_tests[stepName] = {
+    def isMobile = (browser == "ios" || browser == "android")
```
luke-hill, line 276: This was the crux of the issue. For some reason having a global variable here, didn't percolate through and down the correct value of isMobile (Possibly due to default values being set). But using a local variable here works fine?? :shrug:

simonwgill (approved):
Yeah, LGTM. I swear I approved this earlier today...

PR #882, opened 2021-02-23
Feature: navigation behavioural tests [NP-1751]

davidsauntson:
Baby's first cucumber.  Please go wild!

Wanted to add some behavioural tests for Greedy Nav as:
 - there aren't any
 - I want to practice
 - we are about to refactor it and the current tests rely on testing internals
 - the refactor involves moving to `IntersectionObserver` which isn't really testable in a JSDOM unit test, so behavioural tests seem appropriate


testing/features/navigation.feature
```diff
@@ -0,0 +1,52 @@
...
+        Then the dropdown menu is hidden
+
+    Scenario: tabbing onto the dropdown toggle opens the dropdown menu
+        Given the navigation does not fit on the screen
+        And the dropdown toggle is present
+        When I tab onto the More button 
+        Then the Close button is present
+        And the dropdown menu is present
+
+    @failing
```
davidsauntson, line 35: This currently works on live because of the presence of header links

testing/features/support/components/navigation/default.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Navigation
+  class Default < ::Base
+    set_url "/iframe.html?id=components-navigation--default-story&viewMode=story"
+
+    element :navigation, ".js-cads-greedy-nav"
```
davidsauntson, line 7: I've tried to go for a more user based approach (eg find buttons by text), which means this class is basically empty.  Is that...ok?

testing/features/support/driver.rb
```diff
@@ -25,7 +25,7 @@ def setup_capybara
     Capybara.configure do |config|
       config.run_server = false
       config.default_driver = :selenium
-      config.default_max_wait_time = 0
+      config.default_max_wait_time = 1
```
davidsauntson, line 28: Not sure if we would want this to be 0?  I'm aware I've only used capybara selectors so not sure if there is a SitePrism thing I should be using?  Or if this should just be set for the nav tests (the delay is required to wait for the More and Close buttons to be present).

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+When("I click the More button") do 
+  click_button("More")
+end
+
+When("I click the Close button") do
+  click_button("Close")
+end
+
+When("I tab onto the More button") do
+  5.times do
```
davidsauntson, line 57: feels a bit brittle

testing/features/navigation.feature
luke-hill: Rightio. Is this something we have a ticket for. We could add it to this as an `@` tag like `@NP-1244`

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+
+Given("the navigation is on the page") do
+  @component = Navigation::Default.new.tap(&:load)
+end
+
+Given("the navigation is on the page with header links") do 
+  @component = Navigation::WithHeader.new.tap(&:load)
+end
+
+Given("the navigation does not fit on the screen") do
```
luke-hill, line 11: We have a tag which will do this for you. I believe it is something like `@requires_small_screen` or something similar.

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+
+Given("the navigation is on the page with header links") do 
+  @component = Navigation::WithHeader.new.tap(&:load)
+end
+
+Given("the navigation does not fit on the screen") do 
+  page.driver.browser.manage.window.resize_to(320, 568)
+end
+
+Given("the dropdown menu is already open") do
```
luke-hill, line 15: Based on the way you have written this cuke (Gherkin step). This is really good. As it's more imperative. I'll put some suggestions to keep your other ones like this.

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+  
+Then("the dropdown toggle is not present") do 
+  expect(@component).to have_no_button("More")
+end
+
+Then("the dropdown toggle is present") do 
+  expect(@component).to have_button("More")
+end
+
+Then("the dropdown menu is hidden") do
```
luke-hill, line 28: For this one. Do we want to validate that it has been set to invisible? Or do we not care? (Just noticed you've switched from not present to hidden)

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+
+Then("the Close button is present") do
+  expect(@component).to have_button("Close")
+end
+
+Then("the More button is present") do
+  expect(@component).to have_button("More")
+end
+
+Then("the header links should be included in the dropdown menu") do
```
luke-hill, line 44: `the dropdown menu has header links`

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+  expect(@component).to have_button("Sign out")
```
luke-hill, line 45: For the signature here you can chain things. So you can do something like

`expect(@component.dropdown_menu).to have_header_links`

testing/features/step_definitions/navigation_steps.rb
luke-hill: We have a tab helper. It's in the header tests (Which caters to safari which is a bit odd).

It might not work directly out of the box. But if you can try "hook" into that. Maybe by providing an additional keyword arg or something - I'll let you run wild.

testing/features/support/components/navigation/default.rb
luke-hill: So if you need to directly use text. I'm not 100% for it, but if enough people decide to then I guess I'll have to cede.

2 things to note.
1) Do these text options change by language? If so should we be doing it?
2) We can still create POM elements using capybara locator args.

Essentially the declaration from item 3 onwards (The hashable args). Is a capybara locator strategy
`element` is the keyword method definition
`:symbol`, is the reference name
`PascalCaseClass` is the class (Only used in Section/Sections)
`{hash: args}` are then the arguments to locate. So here you could do something like `"button", text: "Hey there!"`

testing/features/support/driver.rb
luke-hill: So this is a setting we should be very careful with editing. Ideally we would use in-line waiters. Any declaration or invocation of a CNE (Capybara::Node::Element), either in-line with capybara or using SitePrism, can pass it.

i.e.

```
page.find("button", wait: 5) # Wait up to 5secs for button
@component.navigation(wait: 3) # Wait up to 3secs for navigation SitePrism element
element :navigation, ".js-cads-greedy-nav", wait: 3 # Wait up to 3 secs to "set" the SitePrism reference
```

If you find you need to set an implicit waiting strategy, and you're using something that isn't Mojave Safari, I'd be very intrigued why it's needed.

luke-hill (commented):
I've tried to not focus on style / grammar for now. As that again is a subjective view.

We tend to have a few more "Behavioural" style definitions where possible. I appreciate in the DS it's crap, because we don't really have proper user stories, and it's more of an integration environment. I can maybe try get a feel for how you want to tweak things or pair on how I come to certain vocab style decisions?

testing/features/step_definitions/navigation_steps.rb
davidsauntson: Yeah I guess it should be `not present` for consistency 👍

testing/features/support/components/navigation/default.rb
davidsauntson: Agreed to move to using the SitePrism elements

testing/features/support/driver.rb
davidsauntson: Agreed to use inline waiting where required

testing/features/step_definitions/navigation_steps.rb
davidsauntson: Will potentially look at refactoring as it's tightly coupled to testing the skip links currently.

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,70 @@
...
+
+Given("the navigation is on the page") do
+  @component = Navigation::Standard.new.tap(&:load)
+end
+
+Given("the navigation is on the page with header links") do
+  @component = Navigation::WithHeader.new.tap(&:load)
+end
+
+Given("the navigation does not fit on the screen") do
```
luke-hill, line 11: Everywhere this and the other resize logic is used. If you just use a tag of `@small_screen` it will work for both mobiles and non mobiles, and saves you a load of steps.

luke-hill (commented):
Looking really good. Couple of minor tidy-ups but approach seems fine

testing/features/step_definitions/navigation_steps.rb
davidsauntson: @luke-hill Not sure about implementing this.  It litters the gherkin with `@small_screen` to remove a single step definition.  The `Given` clauses will obviously need to stay in the gherkin as it is part of the scenario.  Using  the `@small_screen` tag with `Background:` instead of `Scenario:` produces an obscure error: 

```
When using the Selenium Logger that is set to pipe to a file, the Net::HTTP adapter (default),
can return unencoded binary (confusingly called ASCII-8BIT in Ruby).
Consequently we set the logger to binmode so it doesn't try to encode the data - this would always
cause errors for non-ASCII characters, whatever the parent encoding is. An example of this is ©.
```

testing/features/support/components/navigation/standard.rb
```diff
@@ -0,0 +1,36 @@
...
+module Navigation
+  class Standard < ::Base
+    set_url "/iframe.html?id=components-navigation--default-story&viewMode=story"
+
+    element :navigation, ".js-cads-greedy-nav"
+    element :more_button, "button", { text: "More", wait: 2 }
+    element :close_button, "button", { text: "Close", wait: 2 }
+    element :last_link, "a", text: "More from us"
+
+    def open_dropdown_menu
```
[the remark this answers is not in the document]
davidsauntson: OK, but this is the opposite of what we discussed yesterday.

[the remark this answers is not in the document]
davidsauntson: Would be useful to get some rationale / guidance on what you expect in a step definition file and what you expect to be in the page object model and, importantly, why.

testing/features/step_definitions/navigation_steps.rb
luke-hill: Fair enough. i think this is one we can table to fix later.

Ordinarily I'd tell you to tag the rule - Which then saves you a whole heap of lines - However you can't do this (Yet).

The only issue with this is this logic will fail on mobiles. So if you check in `hooks.rb` I'm happy if you want to copy the conditional there, and I can make a tech debt ticket to refactor this feature later on once we've done the upgrades required.

davidsauntson: I don't know what this means.  What logic? Resizing the screen?  Are you saying I need to tag these with `@not_mobile`?

luke-hill: Ah sorry. No not quite.

So if you see here: https://github.com/citizensadvice/design-system/blob/master/testing/features/support/hooks.rb#L12-L13 - You'll see the logic that would be desirable.

But for now, as a short-term fix. Just whacking a `unless mobile_phone?` at the end of your resizing logic should be fine. I'll tidy everything up down the line once the requisite updates are done.

davidsauntson: Great, thanks.

luke-hill (approved):

testing/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+Given("the navigation is on the page") do
+  @component = Navigation::Standard.new.tap(&:load)
```
pedantic-git, line 4: General question more for @luke-hill here - how come this is the conventional way to call `#load` on these classes? Is there scope for making it either one of these three things:

1. Happen automatically as part of `.new` (I'm guessing there are good reasons not to do this one!)
2. Happen if a given parameter is passed to `.new` (e.g. `Navigation::Standard.new(load: true)`)
3. Or at least have the `#load` method return `self` so that you can call it as `.new.load` instead of `.new.tap(&:load)`?

pedantic-git:
This was a really good read! Sorry I didn't get to it before it merged but have a stamp of approval anyway!

testing/features/step_definitions/navigation_steps.rb
luke-hill: So.

`#load` as a method will do about 4 things under the hood: https://github.com/site-prism/site_prism/blob/main/lib/site_prism/page.rb#L39-L57

When you instantiate a new page class, it just returns an OO object. So you can't do anything on instantiation (Yet, see below).
When instantiating a page object. We don't like users passing in much (As a few things need to be passed in).
The load method "could" return `self`, but then that conflicts with the specification, which says it returns a yielded object (This is to support Capybara simple / string objects).
By default most of the time load will return `true`. This is to stipulate that it has passed the requisite load validation tests.

However, that all being said, a long standing branch of mine: https://github.com/site-prism/site_prism/tree/feature/auto_run_validations would allow load validations to be automatically ran when calling `#new` - But I've not got time / priority to work on this. This is something that will likely land in 4.x at some point.

pedantic-git: @luke-hill This is very clear. Thank you for the explanation!

PR #883, opened 2021-02-23
fix: nav highlight colour [NP-1650]

baseonmars:
Screenshot 2021-02-23 at 15 04 02

_The right kind of yellow._


davidsauntson (approved):
🟡

PR #902, opened 2021-02-26
Feature/NP-1601 Review v4 build of DS

celonicolosi:
- all scenarios reviewed
- all falling / future releases / NP-xxxx tags removed
- some scenarios fixed
- one scenario removed (Scenario: Content is not auto-opened if previously interacted with)

### DIFF
--- a/testing/features/support/components/header/standard.rb
+++ b/testing/features/support/components/header/standard.rb
@@ -8,10 +8,10 @@ module Header
     include Helpers::Javascript
 
     element :change_language, "a", text: "Cymraeg"
-    element :login, ".cads-header__account-link"
+    element :login, "[data-testid='account-link']"
     element :logo, ".cads-logo"
     element :search_field, "[type='search']"
-    element :search_button, ".cads-search__button"
+    element :search_button, "[data-testid='search-button']"
     element :open_search_pane, "[data-testid='expand-button']"
 
     def search_for(search_term)
--- a/testing/features/support/components/header/standard.rb
+++ b/testing/features/support/components/header/standard.rb
@@ -8,7 +8,7 @@ module Header
     include Helpers::Javascript
 
     element :change_language, "a", text: "Cymraeg"
-    element :login, "[data-testid='account-link']"
+    element :login, ".cads-header__account-link"
     element :logo, ".cads-logo"
     element :search_field, "[type='search']"
     element :search_button, "form > button"
--- a/testing/features/support/components/header/standard.rb
+++ b/testing/features/support/components/header/standard.rb
@@ -11,7 +11,7 @@ module Header
     element :login, ".cads-header__account-link"
     element :logo, ".cads-logo"
     element :search_field, "[type='search']"
-    element :search_button, "form > button"
+    element :search_button, ".cads-search__button"
     element :open_search_pane, "[data-testid='expand-button']"
 
     def search_for(search_term)
--- a/testing/features/support/components/search/standard.rb
+++ b/testing/features/support/components/search/standard.rb
@@ -18,8 +18,7 @@ module Search
     private
 
     def clear_field
-      search_field.send_keys(%i[control end]) if firefox?
-      search_field.send_keys(:end)
+      search_field.send_keys(%i[end])
 
       if mobile_phone?
         current_text_length.times { search_field.send_keys(:backspace) }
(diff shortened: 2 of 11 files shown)


testing/features/support/components/header/standard.rb
```diff
@@ -8,10 +8,10 @@ class Standard < ::Base
     include Helpers::Javascript
 
     element :change_language, "a", text: "Cymraeg"
-    element :login, "[data-testid='account-link']"
```
luke-hill, line 11: We probably don't want to undo data id hooking if possible. Also the data id probably still exists.

testing/features/support/components/search/standard.rb
```diff
@@ -18,6 +18,7 @@ def search_for(term)
     private
 
     def clear_field
+      search_field.send_keys(%i[control end]) if firefox?
```
luke-hill, line 21: Is this something you were debugging? This shouldn't be required on firefox>?

luke-hill (changes requested):
Couple of things to quickly triage.

testing/features/support/components/search/standard.rb
celonicolosi: The tests were falling for Firefox, then I realized that just send_keys(:end) was not sending the caret navigation to the end of the text, resulting in the following error:

Screenshot 2021-02-26 at 13 14 57

```
RSpec::Expectations::ExpectationNotMetError: 
expected: "Anything"
     got: "AnythingCurrent search term" 
```

testing/features/support/components/header/standard.rb
celonicolosi: **Sign in / Sign out element**
```
<li class="cads-header__account-link">
    <form action="/sign-out/out" class="cads-header__account-form" method="POST">
        <button class="cads-linkbutton__regular">Sign out</button>
    </form>
</li>
```

Expecting for login (element :login, "[data-testid='account-link']")
`expect(@component).to have_login`

Generate the following error:
```
RSpec::Expectations::ExpectationNotMetError: expected #<Header::Standard:0x00007fa2161373f0 @loaded=true, @load_error=nil> to have login
...
/Users/marcelonicolosi/Projects/design-system/testing/features/step_definitions/header_steps.rb:12:in `"a login/logout link is present"'
/Users/marcelonicolosi/Projects/design-system/testing/features/header.feature:26:in `a login link is present'
```

testing/features/support/components/search/standard.rb
luke-hill: Right, oki. This shouldn't be the case. See if you can find out what's causing it. I'll dig into it on Monday too if needs be.

This should work for all browsers. If it's not something is likely wrong with our code.

testing/features/support/components/header/standard.rb
luke-hill: Looking at the haml it's likely we just need to add the test id into both branches of the if statement.

testing/features/support/components/search/standard.rb
celonicolosi: In the end, I didn't find out why `search_field.send_keys(:end)` wasn't working properly on Firefox.
I just find out that the array of symbols format works fine for Firefox `search_field.send_keys(%i[end])`, unlike the default `search_field.send_keys(:end)`. The array of symbol works fine on chrome too.

testing/features/support/components/header/standard.rb
celonicolosi: I didn't fully get it! Add the data-testid again on the front-end?

luke-hill: Yep, I'll give you a call this morning if you need help

testing/features/support/components/search/standard.rb
```diff
@@ -18,7 +18,7 @@ def search_for(term)
     private
 
     def clear_field
-      search_field.send_keys(:end)
+      search_field.send_keys(%i[end])
```
luke-hill, line 21: Either we both have or I have seen this error before.

I think going forward we just need to be mindful to always use arrays for this method. I'll raise a minor issue in ruby

luke-hill (changes requested):
Make sure to push up the haml changes too.

luke-hill (approved):

luke-hill:
@baseonmars Just for sanity can you once over these changes (We edit haml here slightly).

davidrapson (approved):
👍 HAML changes look good to me

PR #912, opened 2021-03-01
bugfix: i18n weirdness [NP-1759]

luke-hill:
MInor gem bump to keep both repos in sync fixing a weird i18n bug


KPobeeNorris (approved):
LGTM!

PR #926, opened 2021-03-04
feature: Missing language specific fixes/tests [NP-1480]

luke-hill:
Adds in some final language specific tests
Makes a slight optimisation on a haml render


testing/features/header.feature
```diff
@@ -55,3 +55,18 @@ Feature: Header component
...
         | footer     |
+
+  Rule: A Standard Header (Mobile View)
+    Background:
+      Given a Standard Header component is on the page
+
+    @small_screen @failing @NP-1480
+    Scenario: English Header has be expanded to show the full search bar
+      Then I am able to expand the search bar
+      And I am able to collapse the search bar
```
luke-hill, line 66: This line is hard-coded on both tests.

luke-hill:
This fails backstop for 4 tests that show in some situations the position of the Search Bar is slightly different. I guess we'll need to confirm this change is ok (This change is likely caused by the non-rendering of a blank section now, causing undue padding)

haml/_header.html.haml
```diff
@@ -26,14 +26,12 @@
                   = link["title"]
               - else
                 %a.cads-header__hyperlink{ href: link["url"] }= link["title"]
-
-          - if sign_in_out
```
baseonmars, line 30: This should be reverted as not all users of the design-system will supply a sign_in_out link. As just one example, not all micro-sites need it.

luke-hill:
For this PR what would we like to do now?

Approve and Merge or attempt to fix up the failing labels and then merge?

Thoughts @davidsauntson / @baseonmars

davidsauntson:
@luke-hill what is failing?  The tickets for the `@failing` scenarios in this PR are NP-1480, which is also what this PR is fixing, so I'm a bit confused, sorry!

luke-hill:
Ah sorry. Yes that's poorly explained.

This ticket was to remedy several things. Primarily around labelling / missing tests.

In doing so (And re-adding all the missing tests e.t.c.) I've found one of the newly added tests doesn't work. I.e. one of the bits mentioned in the original ticket didn't get fixed (Or is invalid). So essentially it's a circular reference 😄 

So the two scenarios here that are failing I "believe" might have some missing logic somewhere. The issue was around labelling.

davidsauntson:
OK - if the failing ones are marked as such, happy for you to merge if there is a clear Jira ticket for each failing feature.

luke-hill:
Can we approve this then tackle the remaining issues in a follow-up ticket?

davidsauntson (approved):

PR #946, opened 2021-03-10
Fix: tab helper NP-1765

davidsauntson:
Refactors a method inside the header support file into a helper file, to allow a common way of sending tabs that caters for safari weirdness.  Will unblock one of the failing feature tests for navigation.

### DIFF
--- a/testing/features/support/components/header/standard.rb
+++ b/testing/features/support/components/header/standard.rb
@@ -20,8 +20,7 @@ module Header
     end
 
     def tab_to(desired_area)
-      tab_count = tab_quantity_for_skip_link(desired_area)
-      send_tabs(tab_count)
+      send_tabs(tab_quantity_for_skip_link(desired_area))
     end
   end
 end
--- a/testing/features/support/components/header/standard.rb
+++ b/testing/features/support/components/header/standard.rb
@@ -20,28 +20,8 @@ module Header
     end
 
     def tab_to(desired_area)
-      tab_quantity_for_skip_link(desired_area).times do
-        # This next line is a hack until capybara cut a new version containing
-        # https://github.com/teamcapybara/capybara/pull/2422
-        # ETA Early December based on previous releases
-        #
-        # LH - Nov 2020
-        fake_active_element = page.find("body")
-        if safari?
-          fake_active_element.send_keys(%i[alt tab])
-          # This next line is ANOTHER hack that is needed because Safari is crap
-          # and broken. The modifier key here (OPTION, but called :alt), is not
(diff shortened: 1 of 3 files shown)


testing/features/support/helpers/page.rb
```diff
@@ -31,6 +31,23 @@ def switch_to_newly_opened_window!(new_page: false)
...
 
+    def send_tabs(count)
+      count.times do
+        if safari?
+          send_keys(%i[alt tab])
+          # In Safari the modifier key (OPTION, but called :alt), is not
+          # released after being used in the alt+tab call. This means that
+          # subsequent calls do things like alt+click which downloads a link.
+          #
+          # See: https://bugs.webkit.org/show_bug.cgi?id=219948
```
luke-hill, line 42: Can we keep the date tracker on this (It's a style myself / Quinn trying to keep to) So LH - Dec 2020

testing/features/support/components/header/standard.rb
```diff
@@ -20,28 +20,8 @@ def search_for(search_term)
...
-          # See: https://bugs.webkit.org/show_bug.cgi?id=219948
-          # LH - Dec 2020
-          fake_active_element.send_keys(:alt)
-        else
-          fake_active_element.send_keys(:tab)
-        end
-        sleep 0.1
-      end
+      tab_count = tab_quantity_for_skip_link(desired_area)
+      send_tabs(tab_count)
```
luke-hill, line 24: This can be just written as `send_tabs(tab_quantity_for_skip_link(desired_area))`

luke-hill (commented):
Awesome stuff. This definitely helps things.

Couple of minor stylistic things to tweak, then good to merge.

testing/features/support/helpers/page.rb
davidsauntson: oops sorry - a copy pasta error :-)

testing/features/support/components/header/standard.rb
luke-hill: So because the return value is directly used, it's actually slightly more optimal.

We don't yet use it (But I want to soon), there's a framework called rubocop-performance which links to something called fast-ruby.

If there was say 4/5 chains I'd agree with you. But for < 3 I think doing it like this is good.

luke-hill (approved):

testing/features/support/components/header/standard.rb
davidrapson: Given this is in a browser test I'd be incredibly surprised if method calls had any noticeable affect on performance compared to the browser automation work the test is already doing. So +1 for whatever feels most readable.

PR #948, opened 2021-03-11
feature: Upgrade cucumber (And associated changes)

luke-hill:
This does 3 large things (Even though the diff is small).

* Upgrade cucumber to v5
  * This only includes 1 bump to gherkin, but then along the 5.x series there is a lot of bumps to gherkin (So wanted to do these separately)
  * This also starts removing / prohibiting certain things (So these have been removed)

* Standardise error logging
  * Copy structure from Public-Website

* Begin to remove guards against certain scenarios passing on certain situations
  * We don't care if stuff fails on iOS11 as it's ancient, so we can remove that guard (We had previously amended / fixed up all tests that were tagged as such)


testing/features/support/helpers/page.rb
```diff
@@ -19,12 +19,6 @@ def save_full_page_screenshot(scenario)
       attach(path, "image/png")
     end
 
-    def save_full_page_html(scenario)
```
davidrapson, line 22: If this is going, should `html_file_path` and `html_pages` from the archive list go as well?

luke-hill: Yes. I'll double check that.

celonicolosi (approved):

PR #950, opened 2021-03-12
feat: option to remove the close button for targeted content [NP-757]

baseonmars:
This is coded in the positive, with a default set to true so existing behaviour is maintained.

Set local `"targeted_content" => { "close_button" => false }` to omit the close button. This is useful when targeted content is as part of a UI that does not require the close button - specifically this applies to the Expandable Sidebar that is part of the public advice website.


davidsauntson (approved):
👍 nice!

PR #964, opened 2021-03-16
WIP Fix/header button translation [NP-1884]

davidsauntson:
Fixes a bug in the mobile header where the title of the search expand / collapse button was not translated in the javascript by:
 - creating translations for open and close states
 - storing these in `data-` attributes on the button
 - updating the js to use these attributes when updating the button title

Waiting for the Welsh translation of 'close search' but ready for review.

### DIFF
--- a/haml/_header.html.haml
+++ b/haml/_header.html.haml
@@ -3,6 +3,13 @@
   sign_in_out ||= {}
   main_content_anchor_link ||= "#cads-main-content"
   homepage_url ||= "/"
+  button_attr = {
+    class: %w[cads-header__search-reveal js-cads-search-reveal cads-icon_search],
+    title: t("cads.header.open_search"),
+    "data-testid": "expand-button",
+    "aria-expanded": "false",
+    data: { descriptive_label_show: t("cads.header.open_search"), descriptive_label_hide: t("cads.header.close_search") }
+  }
 
 %header.cads-header
   .cads-grid-container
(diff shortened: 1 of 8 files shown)


haml/_header.html.haml
```diff
@@ -3,6 +3,12 @@
   sign_in_out ||= {}
   main_content_anchor_link ||= "#cads-main-content"
   homepage_url ||= "/"
+  button_attr = {
+    class: %w[cads-header__search-reveal js-cads-search-reveal cads-icon_search],
+    title: t("cads.header.open_search"),
```
luke-hill, line 8: is this value needed now? Given your JS switches it below?

testing/features/header.feature
```diff
@@ -60,12 +60,12 @@ Feature: Header component
...
       Given a Standard Header component is on the page
 
-    @small_screen @failing @NP-1480
+    @small_screen @failing @NP-1885
     Scenario: English Header has be expanded to show the full search bar
       Then I am able to expand the search bar
       And I am able to collapse the search bar
 
-    @small_screen @failing @NP-1480
+    @small_screen @failing @NP-1885
```
luke-hill, line 68: we can change this tag to `@small_screen @future_release @v4.1.0+`

luke-hill (approved):
Seems legit. Couple of minor points but non blocking / could be wrong.

haml/_header.html.haml
davidsauntson: It is needed - for the initial render

testing/features/header.feature
davidsauntson: It's going to be in the next release though so I'll change it to `4.0.4`

davidrapson (approved):
👍 But you've got a lint error from `i18n-tasks`, asking you to run `normalize`.

PR #971, opened 2021-03-18
fix: chrome resizing bug by increasing offset [NP-1603]

baseonmars:
Adjust the offset (I'm not sure if it's correct but this is what's needed for chrome to stop screwing up).

The CSS change is required to prevent the dropdown toggle from hanging over the edge of the page. Changes to GreedyNav are mostly house keeping - I believe further changes are in the works so did not go crazy here.

Documentation for polyfill added as well as a Mock for unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+**Bugfixes**
+
+- 🧭 GreedyNav: Fix greedy nav behaviour when resizing
+
 ## <sub>v4.0.4</sub>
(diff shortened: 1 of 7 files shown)


davidsauntson (approved):
Nice.  This fixes the quick screen resize issue but sometimes the More button is a few pixels too close to the last menu item.  This is fine on all the viewports we tested 👍 

At some point we should replace it with `IntersectionObserver` or smth.

PR #973, opened 2021-03-19
feature: Add new log enhancements (Copy from PW) [NP-1802]

luke-hill:
Same as sister PR.

- Refactor logger creation
- Allow passing log info to stdout
- Standardise log levels


luke-hill:
This should just need rubber stamping.

baseonmars (approved):

PR #983, opened 2021-03-23
feat: remove outline from links and link buttons [NP-1876]

baseonmars:
After some discussion with @minniessuyuhuang we decide to hide the `outline` from links and link buttons. We already style links with a yellow background and black bottom border which more than meets accessibility needs. The outline caused multiple layout issues which are costly to fix and at times inhibit legibility.

After testing the only change that I've found is the site logo requiring a transparent bottom border to prevent a layout shift when `active`. This reduces the logo size slightly which explains the visual regression test updates.

I've linked this against `public-website` and tested various layouts and components (most notably On This Site and Expandable Sidebar both of which are fine, great even).


davidsauntson (approved):
<3

PR #987, opened 2021-03-24
Experimental: Github Actions

davidrapson:
This is an experimental pull request to explore using github actions. This is not intended to _commit_ to anything, but to better understand if this is a good fit for this project.

Going to use this pull request to show my working for each step in the process. If this has legs this will form part of a proposal and we can apply these changes fresh with some more documentation and clean things up.

My general goals with this are to get a pipeline which:

- Runs significantly faster than the current Jenkins pipeline for the same steps (5-7 minutes for a pull request build ideally)
- Has minimal or no dependencies on custom images in `docker/`, runs using common actions steps


davidrapson:
OK, so far so good. We can run the node tests pretty simply as expected.

image

davidrapson:
Can we run ruby lint in a separate job? Yes we can. And fast!

image

image

davidrapson:
image

davidrapson:
Nice to see the checks tab populated, bonus.

image

Also noticed that it runs _two_ builds. One for the commit and one for the PR.

image

Probably good, but I wonder if we can limit that later on?

davidrapson:
OK, the docs build step runs too. Unsurprisingly this takes a little while the storybook build is slow, but still decent enough.

image

davidrapson:
Also just noting here that this is not reporting "billable minutes" as it would for private repos, as public build minutes are free. We knew that going in, but good to confirm.

davidrapson:
Hey pa11y runs

image

davidrapson:
Going to try running selenium tests too. The ubuntu actions agents have a recent version of chrome pre-installed so I wonder if we can just test against that?

davidrapson:
Hey! It ran!

image

Not running against grid as that’s for running against whole flock of testing nodes and we only really want one in CI, but it turns out we can just run it using the chrome image that’s pre installed on actions runners. Nice!

davidrapson:
Got artefacts uploading. We probably only want to run this on failure, and watch out for the retention period as I've set it pretty short but otherwise all good.

image

davidrapson:
Wondering about running server-dependent tests in parallel.

It looks like https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#sharing-data-between-jobs might be the way to go combined with https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#creating-dependent-jobs

davidrapson:
OK, so this is basically working now.

image

This is running all the key steps we need in a pull request. A little clean up to do and then on to the harder stuff: running in browserstack and reporting to Slack. That needs secrets configuring first though.

testing/features/support/helpers/env_variables.rb
```diff
@@ -60,7 +60,7 @@ def headless?
 
     def base_url
       if environment == :local
-        "http://ca-styleguide:6006"
+        ENV.fetch("TESTING_BASE_URL", "http://ca-styleguide:6006")
```
luke-hill, line 63: This change can be permanent - We never did it here. Have no probs with us just having `ENV.fetch("TESTING_BASE_URL")` as a param with a default like you've done here.

luke-hill (commented):
Quick observation (I know this is a spike).

Happy for some of the artifacts not to be stored for long / uploaded for long. But definitely would want to try retain some of the files. Specifically `automation_logs.log` for a bit longer.

The webdriver / report files are definitely less important (And the larger ones).

👍

testing/features/support/helpers/env_variables.rb
davidrapson: Cool, that was my plan if we go with this approach. But wanted to avoid needing to touch any other build config in this spike.

.github/workflows/ci-workflow.yml
```diff
@@ -0,0 +1,158 @@
...
+      - run: ruby -run -e httpd -- docs -p 6006 &
+      - run: ./bin/wait-for localhost:6006 -- bundle exec rake design_system:all
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: grid-artifacts
+          path: testing/artifacts
+          retention-days: 7
+
+  browserstack:
```
davidrapson, line 133: This doesn't actually _run_ browserstack tests yet, but illustrates how we'd use a matrix to do that. I think in reality we might want to run this as a separate workflow file which listens to the `page_build` event as that's what we run these tests against. Running against every master build doesn't tell us anything, only when the pages build changes.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#page_build

.github/workflows/ci-workflow.yml
```diff
@@ -0,0 +1,158 @@
...
+          node-version: ${{ env.NODE_VERSION }}
+      # Cache and install npm dependencies using https://github.com/bahmutov/npm-install
+      - uses: bahmutov/npm-install@v1
+      - uses: actions/download-artifact@v2
+        with:
+          name: docs-build
+      # Explicitly install testing dependencies
+      -  run: npm ci
+        working-directory: ./testing
+      - run: ruby -run -e httpd -- docs -p 6006 &
```
davidrapson, line 77: This is not the most elegant, but realised we can use the built in ruby to start a simple server in the background. This can be consistent for each steps as it doesn't rely on having all our dependencies installed for every job.

davidrapson:
Initial browserstack test ran OK

image

Need to fake the `DOCKER_TAG` for the time-being, but otherwise promising.

davidrapson:
Closing this as the experiment did what it needed, starting to pull out mergeable changes from this PR.

PR #1001, opened 2021-03-29
fix: polyfill ResizeObserver

baseonmars:
We need to polyfill ResizeObserver so that storybook works with IE11 and older versions of Safari.

I've moved the location of the polyfills to preview.js (loaded from polyfill.js) as I noticed manager.js is not loaded when viewing components in a separate window.

Screenshot 2021-03-29 at 11 21 21


luke-hill:
I think the approval probably should come from the FE'ers. 30k LOC changed makes me worry. But I'm guessing that's Javascript for you :)

Once we've re-merged this, I can re-test this on all the relevant ones. Might I suggest as part of this ticket, we also tag the nav feature, so we don't get failures for now.

The best way to do this would be at the top of `navigation.feature` add the following to line 1.

`@future_release @v4.0.5+`

baseonmars:
@luke-hill doh - I've only gone and installed node 15 haven't I. good spot.

luke-hill:
If you stub the nav feature with the tag I suggested I can approve this. But probably still might need some better 👀

luke-hill (approved):

baseonmars:
> If you stub the nav feature with the tag I suggested I can approve this. But probably still might need some better 👀

Great - that's done. You'll need to test against localhost as the changes won't be published to GitHub Pages until the next release is cut.

PR #1015, opened 2021-03-31
Initial GitHub actions workflow

davidrapson:
## Context

https://citizensadvice.atlassian.net/browse/NP-1786

## Changes

This pull request introduces the minimal changes requires to be able to start using Github Actions. It focuses solely on what's required for the main build, so no BrowserStack, but in  https://github.com/citizensadvice/design-system/pull/987 you can see the full suite of BrowserStack tests running as proof of concept.

There are couple of additional steps to work out in before we can replace the full CI pipeline:

- A replacement for `DOCKER_TAG` being used to group BrowserStack runs
- An approach to Slack notifications on failures

On that basis, the intent of this PR is that it should cover everything needed for the default build run and can be run **alongside** the existing Jenkins pipeline. When we've seen this running successfully for a week or so, and have resolved the final details with BrowserStack tests and Slack notifications we can begin decommissioning the Jenkins pipeline.

## Verification

You can see the original exploratory PR here with full verification steps and comments: https://github.com/citizensadvice/design-system/pull/987

## Workflow

- Runs node and ruby test stages in parallel as separate jobs
- Runs a build stage which prepares the storybook docs and "seeds" dependent jobs with a copy of the built docs
- Pa11y, Backstop, and Selenium tests are then ran in parallel against a static server running the built docs

image

As a result build times should be a little over **~6 minutes**. Most of the time is spent in the Storybook build and in Backstop tests both of which we can stand to optimise, but that's outside of the scope of this PR.

## Things to note

- Does not require us to build any of our custom docker images.
- This requires no scripts from `bin/` with the exception of `wait-for`

We may want to keep some version of the Docker image for basic local development but they are not required for this workflow so we have the option to completely remove this tooling if we want.


davidsauntson (approved):
🎉 

This is a great improvement!  I very much enjoy the speed and messaging improvements in particular.  Getting feedback on all the stages regardless of which ones have failed is very useful and since this is quick/free it makes sense to keep rather than 'failing quickly' like we do in Jenkins.

Looks like it ready to start running alongside the current Jenkins pipeline to me.  

Some random thoughts:
 - would be good to somehow/where serve the backstop reports without having to download and serve them locally.  Wonder if we can pop them in an S3 bucket or smth?
 - I think it's worth starting a dedicated build channel in Slack for the design system - when people start contributing to it they will want to see it without the other content-platform related build messaging
 - How does the current Slack messaging happen?  I'm guessing this just by virtue of having the repo linked up in Slack.  The failure message isn't great as you mentioned.  I think the current build failure message from Jenkins is fine, if we can replicate that?  Saw there was a GitHub Action Slack package that we could use but not necessary for this PR.

davidrapson:
> would be good to somehow/where serve the backstop reports without having to download and serve them locally. Wonder if we can pop them in an S3 bucket or smth?

Yeah, we certainly could. Serving locally isn't too bad, but you're right it'd be nice to have them somewhere. If we could set up an s3 bucket for hosted artefacts we could potentially use that for a lightweight "preview" app at the same time by publishing the docs build there too. Something to come back to.

> I think it's worth starting a dedicated build channel in Slack for the design system - when people start contributing to it they will want to see it without the other content-platform related build messaging

Thinking the same thing. I don't think it's helpful having them in the content platform builds channel, public-website failing is a very different level of severity from the design system build failing. May aim to keep it the same to begin with, but totally up for splitting things out.

> How does the current Slack messaging happen? I'm guessing this just by virtue of having the repo linked up in Slack. The failure message isn't great as you mentioned. I think the current build failure message from Jenkins is fine, if we can replicate that? Saw there was a GitHub Action Slack package that we could use but not necessary for this PR.

The current Slack messaging happens within the Jenkins pipeline, so yeah we'd have to use a similar action / custom script in the workflow if we wanted a similar thing.

davidsauntson:
>> The current Slack messaging happens within the Jenkins pipeline, so yeah we'd have to use a similar action / custom script in the workflow if we wanted a similar thing.

Ah sorry I was wildy vague - I meant the current messaging from Actions rather than the Jenkins pipeline (eg https://acab.slack.com/archives/GVAETDJUD/p1617266965127700).  But yeah, sounds good.

davidrapson:
Oh yeah, that's from whatever github integration we have within Slack.

PR #1020, opened 2021-04-01
BrowserStack github actions worfklow

davidrapson:
Companion to https://github.com/citizensadvice/design-system/pull/1015 with a focus solely on running BrowserStack tests when github pages is updated.


.github/workflows/browserstack-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+            browserstack_access_key: 'PRO_BROWSERSTACK_ACCESS_KEY'
+          - config: 'Windows_7_latest'
+            browser: 'firefox'
+            browserstack_username: 'PRO_BROWSERSTACK_USERNAME'
+            browserstack_access_key: 'PRO_BROWSERSTACK_ACCESS_KEY'
+          - config: 'OSX_Mojave_latest'
+            browser: 'safari'
+            browserstack_username: 'PRO_BROWSERSTACK_USERNAME'
+            browserstack_access_key: 'PRO_BROWSERSTACK_ACCESS_KEY'
+          - config: 'iPhone11_13'
```
davidsauntson, line 31: Would be nice to put the two mobile tests into a separate job called 'mobile-browsers'

davidsauntson (approved):
Nice!  Looks good to me 👍 

Checked in Browserstack and the build and session names look correct.  I can't see any mobile runs in the Browserstack UI, but can see they are passing/failing correctly in GitHub and the reports.

.github/workflows/browserstack-workflow.yml
davidrapson: I did start with it this way and it’s quite a lot more boilerplate, but let me push a change and we can compare the two.

.github/workflows/browserstack-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+on: [push]
+
+jobs:
+  browserstack:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - config: 'Windows_10_latest'
```
luke-hill, line 15: can we use version numbers here. Their implementation of these things isn't great.

.github/workflows/browserstack-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+            browserstack_username: 'MOBILE_BROWSERSTACK_USERNAME'
+            browserstack_access_key: 'MOBILE_BROWSERSTACK_ACCESS_KEY'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - run: bundle exec rake design_system:all
+        env:
+          BROWSERSTACK: true
```
luke-hill, line 46: can't these values be read from the env file we provide?

testing/features/support/drivers/browserstack/base.rb
```diff
@@ -65,11 +65,15 @@ def browserstack_hub_url
...
+        "Design System - #{identifier} - #{base_url}"
       end
 
       def session_name
-        PayloadValuesGenerator.session_name
+        "SHA: #{sha} - CALLING_PROJECT: design-system"
+      end
+
+      def identifier
+        browserstack_build_name.present? ? browserstack_build_name : sha
```
luke-hill, line 76: if you remove the present call here, and alter the original method to a `[]` graceful failure. Then it fixes any potential NME's

This then can just become `browserstack_build_name || sha`

Even better would be (Given we contain the sha in the session_name), would be to make this `browserstack_build_name || "Local Machine run - Month #{Time.now.month} - Ignore results!"`

luke-hill (changes requested):
Great stuff, couple of places it could error locally but other than that it's grand.

.github/workflows/browserstack-workflow.yml
davidrapson: Not really, this is the documented way of doing this for an action; to pass things in via this env property. If we go with this long term than that env file would go anyway.

.github/workflows/browserstack-workflow.yml
davidrapson: I'd quite like to keep up with evergreen browser versions if we can. This seems to be working here so are you happy if we trial it using latest and if we hit any issues we can lock the versions?

PR #1022, opened 2021-04-01
Fix colour docs

davidrapson:
Noticed whilst browsing the design system docs that the colour documentation was showing an empty page 💥.

This was down to a non-existing colour variable being used in the docs and an optimistic helper method that threw an error. Storybook can still be built as this error happens at runtime.

Also added a visual regression test for the colour docs. Feels a little odd to add visual regression tests for written documentation rather than component samples, but as this is a partially generated page it feels like the simplest way to tell that is has been built correctly.


davidsauntson (approved):
🍸

PR #1032, opened 2021-04-06
Remove cross-domain test from the footer

davidrapson:
We had a test for the footer link which opens up a new tab and switches to it.

It'd be nice to avoid needing to test cross domain. As it happens the test doesn't tell us much as this component is designed to accept any link, not specifically the one being tested.

Deleting this scenario also lets us remove some complex helper methods.

### DIFF
--- a/testing/features/footer.feature
+++ b/testing/features/footer.feature
@@ -23,11 +23,6 @@ Feature: Footer component
       And a copyright notice is present
       And a company info is present
 
-    @not_mobile
-    Scenario: Users can report a problem about the specific page they are on
-      When I report a problem with this page
-      Then I am presented with a form to report the issue about the page I am on
-
   Rule: Minimal Footer
     Background:
       Given the Minimal Footer component is on the page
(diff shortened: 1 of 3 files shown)


davidsauntson (commented):
Agree that all the design-system does is render the link so makes sense to remove this from here.  

However, we should be adding it into the public website as we do want to make sure the use can navigate to a form to submit feedback on the page.

Might be worth checking with @celonicolosi  / @luke-hill  to see if they have other plans for the helper methods before you remove them - they may be expecting them to be there when they write other behavioural tests.

davidrapson:
Yeah good point, I think we already have an equivalent test in public website. I think losing the helper methods is _OK_ as we shouldn't really be testing cross-domain for components. But will definitely double check first 👍

davidsauntson (approved):
Cool - approving then!

davidrapson:
Confirmed with Marcelo and we're all good. Going to merge this.

PR #1034, opened 2021-04-06
Adjust dependabot rules

davidrapson:
- Makes npm schedule weekly to match with ruby dependencies
- Fixes an issue with the testing rule, package.json is at root of testing/ directory, not testing/visual-regression


davidsauntson (approved):
👍

PR #1051, opened 2021-04-12
feature: DS Automation - Notice Banner [NP-1881]

celonicolosi:
Simple automation for Notice Banner in Design System.

### DIFF
--- a/testing/features/support/components/notice_banner/notice_banner.rb
+++ b/testing/features/support/components/notice_banner/notice_banner.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-module DesignSystem
-  class NoticeBanner < ::Base
+module NoticeBanner
+  class Example < ::Base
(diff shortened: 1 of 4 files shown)


testing/features/support/components/notice_banner/notice_banner.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module NoticeBanner
+  class Example < ::Base
+    set_url "/iframe.html?id=components-notice-banner--example&args=&viewMode=story"
+
+    element :notice_banner_title, ".cads-notice-banner__title"
```
davidsauntson, line 7: Would be good to remove the styling class and use something like:
`element :notice_banner_title, "span", { text: "Notice banner title" }` instead

celonicolosi: Done!

davidsauntson (approved):
👍 nice!

PR #1052, opened 2021-04-12
feature: DS Automation Pagination [NP-1882]

celonicolosi:
A couple of scenarios for automate the Pagination component.

### DIFF
--- a/testing/features/step_definitions/pagination.rb
+++ b/testing/features/step_definitions/pagination.rb
@@ -5,7 +5,7 @@ Given("an Example Pagination component is on the page") do
 end
 
 Given("a Paging Info Pagination component is on the page") do
-  pending # Write code here that turns the phrase above into concrete actions
+  @component = Pagination::PagingInfo.new.tap(&:load)
 end
 
 Then("the current page is highlighted") do
@@ -30,6 +30,6 @@ Then("there are buttons to skip to the next and previous page") do
   pending # Write code here that turns the phrase above into concrete actions
 end
 
-Then("the number of pages is displayed") do
-  pending # Write code here that turns the phrase above into concrete actions
+Then("the number of results is displayed") do
+  expect(@component).to have_number_of_results
 end
--- a/testing/features/step_definitions/pagination.rb
+++ b/testing/features/step_definitions/pagination.rb
@@ -1,33 +1,35 @@
-Given('an Example Pagination component is on the page') do
+# frozen_string_literal: true
+
+Given("an Example Pagination component is on the page") do
   pending # Write code here that turns the phrase above into concrete actions
 end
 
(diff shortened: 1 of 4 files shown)


testing/features/step_definitions/pagination.rb
```diff
@@ -0,0 +1,41 @@
...
+  @component = Pagination::Example.new.tap(&:load)
+end
+
+Given("a Paging Info Pagination component is on the page") do
+  @component = Pagination::PagingInfo.new.tap(&:load)
+end
+
+Then("the current page is highlighted") do
+  if chrome? || mobile_phone?
+    expect(@component.current_page.native.css_value("background-color")).to eq("rgba(1, 129, 118, 1)")
```
davidsauntson, line 13: We shouldn't text colours like this in feature tests - the visual regression tests should be sufficient.

testing/features/support/components/pagination/example.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Pagination
+  class Example < ::Base
+    set_url "/iframe.html?id=components-pagination--example&args=&viewMode=story"
+
+    element :current_page, ".cads-paging__current a"
```
davidsauntson, line 7: A better selector for this might be `[aria-current="page"]` as you would then be:
 - not using a styling class in a test (this makes the test fragile)
 - also testing accessibility 

If you change the selector to this, then you can just do `expect(@component).to have_current_page` in the step definitions.

testing/features/support/components/pagination/paging_info.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Pagination
+  class PagingInfo < ::Base
+    set_url "/iframe.html?id=components-pagination--paging-info&args=&viewMode=story"
+
+    element :number_of_results, ".cads-paging-info"
```
davidsauntson, line 7: You can avoid using styling classes in these selectors by using the element and text option, like this:
```
element :number_of_results, "h2", { text: "123456 results" }
```

davidsauntson (changes requested):
👍🏻 great!  

Some small suggested changes around how we find elements to test (more user centric testing) and not testing for specific colours.

testing/features/step_definitions/pagination.rb
celonicolosi: Agree with you, I don't like to assert this either, but the steps say:
`Then the current page is highlighted`
This was the way I found to assert if the current_page was *highlighted*.

I'm going to change the step to something like `Then current page displayed` and then `expect(@component).to have_current_page` in the step-definition.

davidsauntson (approved):
👍🏻 lovely!

PR #1054, opened 2021-04-15
fix/feat: selected radio buttons [NP-1889]

baseonmars:
Update the styling of radio buttons to match the Design System - basically a grey border and dot.

I've also added the functionality to render a radio button pre-checked, this allows me to easily add tests for this state to the VRT suite.

In response to @davidsauntson's comment I've adjusted the sizes of the central dot, and also reduced the small size from 25px to 24px. Aria guidelines are not corrected in this PR.


davidsauntson (approved):
This fixes the issue in the ticket but I have noticed the following - I can put this in a new ticket to keep this PR concise (I was sort of expecting to do it anyway as part of the advice only search filtering):
 - we should be following the WAI-ARIA guidelines for radio buttons here: https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-15
 - the inner circle on selected buttons does not match the design (too small and not centered)

PR #1076, opened 2021-04-23
Minimal Rails engine

davidrapson:
## Background

https://citizensadvice.atlassian.net/browse/NP-1862

## Goals

The goal of this pull request is to:

- Add the skeleton structure of a Rails engine.
- Include a couple of experimental components.
- Include rspec tests and component previews for all experimental components.

## Non-goals

- Not published anywhere. The aim is for this pull request to be mergeable and installable as an *experimental* feature, but not included in any releases.
- No integration with storybook. I'm not generating any example files for use in storybook here, that will come later.
- I've not added any code to try and install the node module when installing the gem. I think this is the right call whilst this is in experimental mode, but we should consider doing this before we do a stable release.

## Structure

This pull request introduces two new directories:

- `engine/` - This is the source for the Rails engine code, and is what gets required by apps.
- `demo/` - A companion Rails demo app used by the engine tests and as a way of viewing component previews.

## Initial components

I've selected three components as good candidates for this initial iteration.

- ~`PageReviewComponent` - Selected because it's by far the smallest component we have and is a good test of date formatting and localisation within the engine.~ Removed this one as I reckon the date localisation needs a bit more thought as in the current version we accept a pre-formatted date string.
- `PaginationComponent` - Selected because we already have a production ready view component version within public-website. This ports this over with the existing test suite and previews.
- `AssetHyperlinkComponent` - Selected because it's also small but is used within Contentful renderer functions in public-website which is the main way the existing components are used.

## How to review this pull request

There's a lot of code in this pull request but most of it Rails boilerplate for the `demo/` app. I suggest:

- Focussing on the `engine/` code
- Try running the rspec tests within `engine/`
- Try running the demo Rails app and checking the component previews
- Checkout https://github.com/citizensadvice/public-website/pull/1278 and try running public-website using this code

There are instructions in `engine/README.md` with details on how to get set up locally.


demo/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "demo",
+  "private": true,
+  "version": "0.1.0",
+  "dependencies": {
+    "@citizensadvice/design-system": "file:..",
```
davidrapson, line 6: This loads the design system package from the local project so the demo app always runs against the latest code.

engine/design_system.gemspec
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+lib = File.expand_path("lib", __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require "design_system/version"
+
+Gem::Specification.new do |spec|
+  spec.name = "design_system"
```
davidrapson, line 8: @pedantic-git I'm a little unsure about best practices for naming this.

At least for Rails engines the gem name and the engine namespace need to match. Using plain `design_system` here is nice as it means components can be called with `DesignSystem::PaginationComponent` which I like as an API.

I do wonder what happens if we published this more widely though. I assume best practice would involve prefixing this somehow. I have an aesthetic distaste for the `ca_` prefix we often use and that would end up with `CaDesignSystem::PaginationComponent` which…ugh. Alternatively `citizens_advice_design_system` would result in `CitizensAdviceDesignSystem::PaginationComponent` or even `CitizensAdvice::DesignSystem::PaginationComponent` which is accurate but…verbose.

Am I overthinking things? If we went for the prefixed name is there a way for apps to alias to a shorter name?

engine/design_system.gemspec
```diff
@@ -0,0 +1,33 @@
...
+    # spec.metadata["allowed_push_host"] = "https://rubygems.org"
+  else
+    raise "RubyGems 2.0 or newer is required to protect against " \
+      "public gem pushes."
+  end
+
+  spec.files = Dir["{app,config,lib}/**/*", "README.md"]
+
+  spec.add_runtime_dependency     "haml", [">= 5.1.0", "< 6.0"]
+  spec.add_runtime_dependency     "rails", [">= 6.0.0", "< 7.0"]
```
davidrapson, line 27: I'm requiring the whole of rails here, Primer goes for the approach of requiring just the parts of Rails used which might be worthwhile? https://github.com/primer/view_components/blob/main/primer_view_components.gemspec#L30-L31

engine/design_system.gemspec
davidrapson: Ended up looking around at other examples and went with `citizens_advice_components`, which makes the namespace `CitizensAdviceComponents::PreviewComponent`.

That extra `Component` at the end is the convention of view components. So a little verbose, but descriptive and includes the organisation name as a prefix. Feels OK.

engine/app/components/citizens_advice_components/asset_hyperlink_component.html.haml
```diff
@@ -0,0 +1,5 @@
+%a{ href: href }
+  = description
+  %span.cads-asset-type<
+    %span.cads-icon_file
+    \ #{number_to_human_size(size)}
```
davidrapson, line 5: Because we're in a Rails context here we can use the `number_to_human_size` helper rather than the limited `" #{(asset_hyperlink['size'] / 1_000_000.0).round(1)} MB"` we are currently using.

pedantic-git (approved):
So I wrote half a dozen comments and deleted every one because every time I decided your way was better than the way I had in my head.

As a first pass, this is perfect. I've checked it out and played with it too and it's really good.

In the next release you probably want to create an install generator that installs the npm package and any config files (e.g. I can imagine the header component adding a file to config/locales/*.en.yml). You can crib from ActionText's install generator: https://github.com/rails/rails/blob/main/actiontext/lib/generators/action_text/install/install_generator.rb

But that's one for another day. Merge away!

davidsauntson:
👏🏻 This is really great!

davidsauntson (approved):

davidrapson:
@davidsauntson Mind giving this one more review. Not much changed aside from:

- Backed out the `PageReviewComponent`. I realised the current version accepts a pre-formatted but this version I had accepting a Date object. I think this is the right interface, but warrants a bit more review first.
- Tidied up the set up documentation, particularly around the demo app

davidsauntson (approved):
Noice 🕺🏻

PR #1092, opened 2021-04-29
Generate examples tasks from component previews

davidrapson:
## Background

As of https://github.com/citizensadvice/design-system/pull/1076 we've introduced an experimental rails engine based on view components as a future replacement for our standalone haml partials.

Currently we use a custom webpack loader to let us use our haml partials from storybook. This _works_, but requires us to recreate parts of rails on an ad-hoc basis. With the new engine we have a full copy of rails available to us so we'd like to use that somehow to generate examples for the docs.

## Goals

- Provide a way of generating plain HTML examples from component previews.
- Use at least one generated example to replace an existing example within the docs.
- The backstop tests should pass with little or no modification.

### Adds a task for generating examples

Adds a rake task that generates plain html examples from component previews which can then be used in storybook.

```
./bin/generate-examples
```

I've swapped out the pagination example—which was already plain html—with one generated by this new task. If the backstop tests pass then I've managed to swap out a visually like-for-like example.

I've also added an example documentation section showing how the component is used from within rails. It's felt OK hand-writing the usage docs for the existing partials, so have followed that approach here too rather than trying to generate anything.

### Adds a watcher to rebuild examples on change

Adds a version of the start command which also runs a watcher which rebuilds the examples when previews are updated; which in turn triggers a storybook update giving us live-reload behaviour.

```
npm run start:watch
```

We might eventually want to make this the default for `npm start`, but for now I've kept it as a separate command.

I went with using npm run tasks to do this rather than ruby scripts as all storybook related scripts are already using node.

## How to test these changes

- Run storybook using `npm run start:watch`
- Go to `http://localhost:6006/?path=/story/components-pagination--example`
- Try editing the parameters for `with_fifth_page` in `pagination_component_preview.rb` and confirm the example live reloads (tip: try reducing `num_pages` to < 10)
- Add a new component preview example and run `./bin/generate-examples` manually and confirm you get the expected output

## Other options considered

I did look at https://github.com/jonspalmer/view_component_storybook which looks like a great if you want really tight integration with Storybook. For our purposes we want _less_ integration, with a longer term goal of building our own docs using the component library itself. So for that reason I went for a more lo-fi approach.

💭 Thoughts very welcome as I'm sure there's plenty of different ways we could go about this.


styleguide/components/pagination/pagination-docs.mdx
```diff
@@ -23,3 +21,25 @@ Supplementary component used alongside pagination to show how many results have
...
+
+```rb
+render(CitizensAdviceComponents::PaginationComponent.new(
+  current_params: request.query_parameters,
+  num_pages: 100,
+  current_page: 100
+))
+```
+
+The component assumes that you are using `page` as the query otherwise it will respect any other passed in from `current_params`.
```
davidrapson, line 37: I realised whilst writing this that we are force the query parameter used for pagination. Maybe that should be configurable?

pedantic-git: Might be worth looking at https://github.com/kaminari/kaminari for best practice? If I recall correctly, it provides default names for these params but allows them to be overridden if people _really_ need it.

demo/lib/tasks/generate_examples.rake
```diff
@@ -0,0 +1,40 @@
...
+        puts "Skipping #{component_key}"
+      else
+        @session.get "/rails/view_components/#{preview.preview_name}/#{example}"
+        doc = Nokogiri::HTML.parse(@session.response.body)
+
+        component_html = doc.css("#content").inner_html
+
+        output_path = Rails.application.root.join("../styleguide/examples/#{component_key}.html")
+
+        dirname = File.dirname(output_path)
```
pedantic-git, line 31: Slightly weird to call dirname on a string you've just created - couldn't you construct it in two parts:

```ruby
dirname = Rails.application.root.join("../styleguide/examples")
output_path = dirname + "#{component_key}.html"
```

demo/lib/tasks/generate_examples.rake
```diff
@@ -0,0 +1,40 @@
...
+        FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
```
pedantic-git, line 32: The `unless` here is unnecessary since `mkdir_p` succeeds even if the directory exists.

styleguide/components/pagination/pagination-docs.mdx
pedantic-git: I'm not sure I understand the English explanation above. Does `respect` mean "pass through"?

styleguide/components/pagination/pagination.stories.js
```diff
@@ -1,5 +1,5 @@
 import docs from './pagination-docs.mdx';
-import templateExample from './_example.html';
+import templateExample from '../../examples/pagination/with_fifth_page.html';
```
pedantic-git, line 2: What happens if this page hasn't been generated yet? Are we going to force a run of the new rake task during the main build?

If the answer to this is "the files are committed to git", how are we going to ensure people remember to regenerate them if they change the component.

pedantic-git (approved):
This is great! Nothing major but a few minor comments below.

demo/lib/tasks/generate_examples.rake
pedantic-git: If you want to leave it that way round, your existing `output_path` is already a `Pathname` object so you can write: `dirname = output_path.dirname`.

styleguide/components/pagination/pagination-docs.mdx
davidrapson: That's correct, pass through is much clearer.

styleguide/components/pagination/pagination.stories.js
davidrapson: Running the task during the main build would be a bit too late, so largely relying on the files being committed to git. But what we likely want to do as a future iteration is have a watcher that runs alongside the storybook dev server which regenerates them then. But open to other suggestions. This is kinda what we do now with the current webpack loader approach, although it's a bit flaky.

pedantic-git: This sounds like a good approach.

davidrapson: 👍  Ended up adding a watcher in this PR, have updated the description to cover that

package.json
```diff
@@ -68,6 +69,7 @@
     "babel-eslint": "^10.1.0",
     "babel-loader": "^8.1.0",
     "chalk": "^4.1.1",
+    "concurrently": "^6.0.2",
```
davidsauntson, line 72: We have `npm-run-all` in the deps already - this can run tasks in parallel (see the `npm build` task).

styleguide/components/pagination/pagination.stories.js
davidsauntson: Can we have the rake task in the CI process so we know the docs will always be correct on release, regardless of whether someone has remembered to run the task / edited haml files with the watcher active?

davidsauntson (approved):
🎺  This is great!

Question about CI and the rake task but nothing blocking.

package.json
davidrapson: Good catch!

davidrapson:
> Can we have the rake task in the CI process so we know the docs will always be correct on release, regardless of whether someone has remembered to run the task / edited haml files with the watcher active?

The issue with the CI process is that you'd then have to commit the files again during CI which is a touch circular. But thinking about it we could totally run this during the current release process when we build the docs.

davidrapson:
> The issue with the CI process is that you'd then have to commit the files again during CI which is a touch circular. But thinking about it we could totally run this during the current release process when we build the docs.

Committed and then reverted an updated to do this. Needs a little more thought to work with the current CI scripts. I reckon we come back to this once we've got a few more components migrated over.

PR #1093, opened 2021-04-30
Replace custom check-size script with size-limit

davidrapson:
Scratching a bit of an itch. We previously provided a custom check-size script which recorded the size of our lib.css file over time and failed the build if the size increased. This approach has a few limitations:

- It only checks raw size, not compressed size
- It checks percentage increase rather than a threshold. This allows slowly expanding the size of our assets as long as it only increases by 1% per time, with no limits set

Proposing switching to https://github.com/ai/size-limit which uses configurable limits for each asset type. It also supports bundling our javascript modules using a clean webpack build to check combined output size.


davidsauntson (approved):
We should formulate a plan for what we're going to do when we reach this limit - otherwise we'll just increase the limit 😄  I'll have a think and we can disucss on Thurs.

davidrapson:
> We should formulate a plan for what we're going to do when we reach this limit - otherwise we'll just increase the limit

Indeed! FWIW I'm not totally against choosing to increase the limit if we hit it. Mostly looking to avoid that choice happening gradually for us. Will hold of merging until we've chatted 👍

davidsauntson:
I'm totally happy for you to merge it!

PR #1102, opened 2021-05-04
Fix: element not of type [NP-1999]

davidsauntson:
Ensures the sign-out button is styled a button rather than a link.

`npm linked` with public website:
image
image

Will require the following overrides in public website advisernet colours scss file:

```scss
$cads-language__brand-primary-contrast: $cads-palette__white;
$cads-language__primary-button-colour: $cads-palette__blue-adviser;
$cads-language__tertiary-button-dark-hover-colour: $cads-palette__blue-adviser-dark;
$cads-header__sign-out-background-colour: $cads-language__brand-primary;
$cads-header__sign-out-colour: $cads-language__brand-primary-contrast;
$cads-header__sign-out-border-colour: $cads-language__brand-primary-contrast;
```


scss/4-elements/_buttons.scss
```diff
@@ -125,3 +125,16 @@ $button-shadow-size: 4px;
     transform: none;
   }
 }
+
+%cads-button__tertiary--dark {
```
davidsauntson, line 129: Went for a placeholder here as the button is only styled to be dark inside the greedy nav - so this isn't needed as a class ATM.  Can always change later if required.

davidrapson: Sounds good; I'd even be tempted to inline this into `cads-header__sign-out` for now.

davidrapson (approved):
👍 Looks good to me

When we come to convert this to a view component I reckon we can stream-line the template side of this then; so this looks good for now.

PR #1104, opened 2021-05-04
fix: callouts as sections [NP-2017]

davidsauntson:
Renders callout as `section` elements as recommended by DAC.

If a `title` attribute is provided, it is rendered as an `aria-label` attribute on the `section` element.  Note that:
 - this is already present in the `notice` hash created by the public website renderer
 - only Adviser callouts have a meaningful title so we will need to be careful when we merge this into public website, as some non-adviser callouts will generate `notice["title"]` like `example-callout-milli-9e8r93`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 **Bugfixes**
 
 - 👤 Header: The sign out button now has the visual appearance of a button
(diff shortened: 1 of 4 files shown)


davidrapson (approved):
👍 Nice

PR #1105, opened 2021-05-04
[Experimental] Convert callout to view component

davidrapson:
As we are updating the callout component in https://github.com/citizensadvice/design-system/pull/1104 I thought it worth having a quick stab at converting this to a view component. Considering this experimental as it's probably one case where we want to thrash out the API for this component a bit more.

- Add component definition and template, use block for content
- Add rspec scenarios; include new title and render check.
- Generate example files

## Design questions

- How do we want to validate the allowed types? We can use active model validations, but these can cause runtime errors. Maybe something like https://github.com/primer/view_components/blob/main/app/lib/primer/fetch_or_fallback_helper.rb to provide a default value if an invalid type is passed
- Should `cads-badge` be its own component? Technically the only valid modifier for this is `adviser`

### DIFF
--- a/engine/app/components/citizens_advice_components/callout_component.rb
+++ b/engine/app/components/citizens_advice_components/callout_component.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class CalloutComponent < ViewComponent::Base
+    attr_reader :type, :title
+
+    def initialize(type: nil, title: nil)
+      super
+      @type = type&.to_sym || :standard
+      @title = title
+    end
+
+    def show_badge?
+      type != :standard
+    end
+
+    def render?
+      content.present?
+    end
+  end
+end
--- a/engine/app/components/citizens_advice_components/callout_component.rb
+++ b/engine/app/components/citizens_advice_components/callout_component.rb
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
 module CitizensAdviceComponents
-  class CalloutComponent < ViewComponent::Base
+  class CalloutComponent < BaseComponent
     attr_reader :type, :title
 
     def initialize(type: nil, title: nil)
       super
-      @type = type&.to_sym || :standard
+      @type = fetch_or_fallback(
+        allowed_values: %i[standard example important adviser],
+        given_value: type,
+        fallback: :standard
+      )
       @title = title
     end
 
--- a/engine/app/components/citizens_advice_components/asset_hyperlink_component.rb
+++ b/engine/app/components/citizens_advice_components/asset_hyperlink_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module CitizensAdviceComponents
-  class AssetHyperlinkComponent < ViewComponent::Base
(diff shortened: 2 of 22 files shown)


engine/app/components/citizens_advice_components/callout_component.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class CalloutComponent < ViewComponent::Base
+    attr_reader :type, :title
+
+    def initialize(type: nil, title: nil)
+      super
+      @type = type&.to_sym || :standard
```
davidsauntson, line 9: Why are we converting the type to a symbol here?

davidsauntson:
In the buttons example you did ages ago, you validated button types using a `%w[]` array with the allowed types in and raising an error - this seemed like a reasonable approach.

engine/app/components/citizens_advice_components/callout_component.rb
pedantic-git: Casting it, rather than converting it, right? Users are expected to pass a Symbol but if they pass a String it'll get cast back into a Symbol.

With this kind of thing, where the value gets rendered in the final HTML, you could have a holy war about whether String or Symbol is better practice because they both have merit but it really doesn't matter as long as you pick one and stick to it, since String is implemented the same ways as Symbol these days anyway (two identical strings will always occupy the same place in memory).

pedantic-git (approved):
I like it. Tick.

pedantic-git:
> In the buttons example you did ages ago, you validated button types using a `%w[]` array with the allowed types in and raising an error if the passed type was not present - this seemed like a reasonable approach.
> 
> I don't think it's worth making cads-badge its own component ATM - but follow your heart, David.

If we're using symbols you can do the same with a `%i[]` array just FYI. Might be a good call if arbitrary values are banned.

engine/app/components/citizens_advice_components/callout_component.rb
davidrapson: I did have this a plain string, which was a little simpler. But convinced myself a symbol was more descriptive; good to know they are basically the same anyway.

davidrapson:
> In the buttons example you did ages ago, you validated button types using a %w[] array with the allowed types in and raising an error if the passed type was not present - this seemed like a reasonable approach.

Posting the discussing @davidsauntson and I had during our front-end catch up. 

The `ActiveModel` validations I was using in the original experiment are a little heavy and importantly treat validation errors as a fatal runtime error. Particularly for our use case where we often pass attributes in from contentful it feels safer to have a fallback value. Better to render a different colour callout than break the app.

We're going to use something along the lines of https://github.com/primer/view_components/blob/main/app/lib/primer/fetch_or_fallback_helper.rb

This takes the approach of using a fallback value in production environments, and throwing a validation error in non-production environments which seems like a reasonable balance 👍

engine/app/components/citizens_advice_components/callout_component.rb
davidsauntson: Gotcha - thanks!

davidsauntson (approved):
✨

PR #1107, opened 2021-05-05
Enable network logs same as PW

luke-hill:
Ronseal PR. Same as PW mirror PR.


davidrapson:
Where are we sending the network logs?

luke-hill:
These are the network logs available by hitting f12 and clicking network. They're being sent as an additional silent capture (So they'd be available via REST, and via the BS interface).

davidrapson:
Just seen the public website equivalent. If this is for debugging a specific device type could we limit the network logging to just that?

luke-hill:
It's for a few different device combinations. Few older versions of chrome. iOS e.t.c.

davidrapson (approved):
👍 Although it would be great to set a date to remove this, or add a comment explaining what we are using them for so we don't forget to disable this afterwards.

PR #1108, opened 2021-05-05
chore: remove resize polyfill from storybook

davidsauntson:
In https://github.com/citizensadvice/design-system/pull/1001, we introduced polyfill for storybook to allow older safari and IE browserstack tests to pass.

This has now introduced a typescript conflict between `resize-observer-polyfill` and `typescript` which is preventing build.

Testing removing the polyfill completely - this may cause the browserstack tests to fail.

### DIFF
--- a/.storybook/polyfill.js
+++ b/.storybook/polyfill.js
@@ -1,7 +1,1 @@
 require('url-polyfill');
-
-const ResizeObserverPolyfill = require('resize-observer-polyfill').default;
-
-if (!window.ResizeObserver) {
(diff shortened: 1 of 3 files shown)


davidrapson (approved):
👍 But worth forcing the browserstack tests in this PR by tweaking the Jenkinsfile?

davidsauntson:
> 👍 But worth forcing the browserstack tests in this PR by tweaking the Jenkinsfile?

https://github.com/citizensadvice/design-system/pull/1109 🎉

PR #1111, opened 2021-05-06
Release/v4.0.6 alpha.1

davidsauntson:


luke-hill:
Before merging this in, could you do a quick rebase with master (Just gets a security fix in too)

PR #1122, opened 2021-05-10
Chore/update npm deps

davidsauntson:
- update npm deps
- run new prettier rules
- npm audit fix
- add postcss (now required for major version bump of cssnano


davidrapson (approved):
👍

PR #1123, opened 2021-05-10
Remove Component suffix

davidrapson:
Looking at other examples such as github's own Primer components, and govuk-components wich distrubute view components, neither of those examples include the Component suffix.

Given we already have Components in the module namespace it makes sense to follow suit and drop the Component suffix for a cleaner interface.


davidsauntson (approved):
👍 nice!

PR #1127, opened 2021-05-11
Targeted content: Fix title and border colours [NP-1944]

davidrapson:
We were previously:

- Incorrectly setting the targeted content title to be the brand colour when it should be the link colour.
- Similarly, overriding the adviser variant to use the adviser brand colour for borders when it should stay as the link colour.

Fixes NP-1944

Need to confirm the second point with @minniessuyuhuang, but I believe this is correct based on the figma files.


davidsauntson (approved):
🟢  &rarr; 🟢

PR #1130, opened 2021-05-11
bugfix/refactor: Browserstack IE Driver issues [NP-1809]

luke-hill:
Partially remedies IE issues (same as public-website)

### DIFF
--- a/testing/features/support/drivers/browserstack/base.rb
+++ b/testing/features/support/drivers/browserstack/base.rb
@@ -26,20 +26,19 @@ module Drivers
       end
 
       def standard_capabilities
-        standard_browserstack_capabilities.merge(standard_browser_capabilities)
+        standard_browserstack_capabilities.merge(browser_version)
       end
 
       def specific_browser_capabilities
         {}
       end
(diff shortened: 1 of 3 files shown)


davidrapson (approved):
👍 Although a couple of thoughts:

- IE isn't in the testing matrix
- If we _did_ add it, it wouldn't work very well at the moment as we'd be testing against storybook which does not have good IE support if any (our components do, but our docs don't)

Does that change anything here?

luke-hill:
We can't actually add it to the testing matrix ... yet.

We have been on the backlog for a while, they've got issues their end to fix. They've fixed one of them, but in doing so have regressed other areas. My gut feel is that it'll be a while before we see anything out of them.

The latter point is valid. I would say at the moment let's deal with that once IE is working, because I envisage that will be a long time away. I don't really know what the best solution would be (Maybe by then we stop supporting IE 🤷 )

The purpose of this ticket is to remedy / remove the hacks that were in place previously (One of these also affects all of our tests on browserstack in a tiny way).

PR #1141, opened 2021-05-17
Targeted content component

davidrapson:
Draft structure for a targeted content component. In starting this I realised we probably need to make the table component first in order to make all the expected previews available. But raising this for some initial 👀  to thrash through the API design for this component.

Ended up following my ❤️  and extracting a `Badge` component as it helped to solve some duplication between the locale files.

**TODO:**

- [x] Rspec tests
- [x] Add all example variants
- [x] Document component API


davidsauntson (commented):
Practically perfect in everyway!

Interface looks good, nothing wrong that I can see 👍🏻

testing/features/components/targeted_content.feature
```diff
@@ -3,78 +3,29 @@ Feature: Targeted Content component
...
-
-    Scenario: Content can be closed after being expanded
-      When I expand the targeted content
-      And I close the targeted content
-      Then I can no longer see additional information
-
-    Scenario: Content can be be auto-opened
-      When I jump to the targeted content
-      Then I can see additional information
+  Background:
```
davidrapson, line 6: @luke-hill Drawing your attention to these changes in particular. Added context in the commit description here https://github.com/citizensadvice/design-system/pull/1141/commits/e081367665511b4919ed8368bc4c0c07a8d27912 but might be worth chatting about it too just so you understand the wider goal here.

engine/previews/citizens_advice_components/targeted_content_preview/adviser.html.haml
```diff
@@ -0,0 +1,44 @@
+= render(CitizensAdviceComponents::TargetedContent.new(type: :adviser,
+  title: "Students or self-sufficient people",
+  id: "targeted-content-adviser")) do
+  %p Students and people who can support themselves without working (either with their own money, or with help from family or friends) can have a right to reside.
+  %p One condition of this type of right to reside is that they won’t be an ‘unreasonable burden on the social assistance system’ - either they’re unlikely to claim benefits or, if they do, they won’t need them for long.
+
+  -# @TODO: Replace with table component once available
```
davidrapson, line 7: Ended up inlining the table example to avoid this being blocked on the table component but we should come back and replace this with a call to the component when that's been migrated.

engine/app/components/citizens_advice_components/badge.rb
```diff
@@ -0,0 +1,27 @@
...
+        fallback: nil
+      )
+    end
+
+    def call
+      content_tag(:span, t(".#{type}"), class: "cads-badge cads-badge--#{type}")
+    end
+
+    def render?
+      @type.present?
```
luke-hill, line 24: if we're using attr readers then ideally we'd refer to those over the iVars proper (Better style).

engine/app/components/citizens_advice_components/targeted_content.rb
```diff
@@ -0,0 +1,74 @@
...
+    )
+      super
+      @title = title
+      @id = id
+      @type = fetch_or_fallback(
+        allowed_values: %i[public adviser],
+        given_value: type,
+        fallback: :public
+      )
+      @heading_level = (heading_level || 2).to_i.clamp(2, 6)
```
luke-hill, line 24: you could reduce the cyclomatic score by abstracting this out to a method below. Maybe `detect_heading_level` as a name?

engine/spec/components/targeted_content_spec.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "public_targeted_content" do
```
luke-hill, line 3: traditionally you would place these inside rspec support. Usually your folder structure would be

```
spec
  support
    shared_contexts
      stuff
    shared_examples
      stuff
```

One file for each shared example. Usually named accordingly. So this one could be `public_targeted_content`

engine/spec/components/targeted_content_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+  let(:title) { "Example title" }
+  let(:heading_level) { nil }
+
+  it "renders content block" do
+    expect(component.text).to include "Example content"
+  end
+
+  it_behaves_like "public_targeted_content"
+
+  it "has a h2 title by default" do
```
luke-hill, line 56: `has a correct h2 title`

engine/spec/components/targeted_content_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    expect(component.at("h2").text.strip).to eq(title)
+  end
+
+  context "when missing type" do
+    let(:type) { nil }
+
+    context "production rails env" do
```
luke-hill, line 63: should we add a context for non-production if the behaviour is diff?

engine/spec/components/targeted_content_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      it "has heading clamped to minimum" do
+        expect(component.at("h2")).to be_present
+      end
+    end
+
+    context "with heading_level over limit" do
+      let(:heading_level) { 7 }
+
+      it "has heading clamped to minimum" do
```
luke-hill, line 96: maximum

testing/features/components/targeted_content.feature
luke-hill: for parity can we keep the cuke / file name anchored.

luke-hill (changes requested):
Few stylistic tweaks but think we've missed a set of unit tests for one of the components.

testing/features/components/targeted_content.feature
davidrapson: Not sure I follow. Do you mean the "Given an anchored targeted content component is on the page" wording?

luke-hill: yep, basically that.

engine/spec/components/targeted_content_spec.rb
davidrapson: Good to know. I reckon I'll keep them inlined for now as there's only one set and used within this file. Buy we should totally do that if we end up with more of these that cross specs 👍

engine/spec/components/targeted_content_spec.rb
davidrapson: Yeah good shout. This is technically covered by the behaviour in `fetch_or_fallback` but doesn't hurt to be explicit.

engine/spec/components/targeted_content_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+    end
+
+    context "with heading_level over limit" do
+      let(:heading_level) { 7 }
+
+      it "has heading clamped to maximum" do
+        expect(component.at("h6")).to be_present
+      end
+    end
+  end
```
davidsauntson, line 116: Do we want to test for an non-integer heading level too like `ew-david`?

davidsauntson (approved):
Super dooper 🩰 
One small additional test suggestion but not a blocker

engine/spec/components/targeted_content_spec.rb
davidrapson: I think I'm happy with this wording. Intending to capture that we default to a `h2` if no `heading_level` is provided.

luke-hill: hang on though, that's not validating that here. If we want to validate that shouldn't it validate that the h2 is present? By doing something like

`expect(component.at("h2")).not_to be_nil`

luke-hill (commented):
Could we also copy to the prose part of the feature something along the lines of what we discussed in your SHA.

a la - `NB: We only test one of the variants here because the Default variant is a subset of the Anchored variant` or something better

davidrapson:
> Could we also copy to the prose part of the feature something along the lines of what we discussed in your SHA.

Done 👍 Went with:

> Note: We test the "anchored" example because this is equivalent to the "default" example but with enough surrounding content to allow us to test all scenarios, including opening by hash.

PR #1146, opened 2021-05-18
Create automated_release_process.feature [NP-1787]

thomasgower:
Requirements for automated release process added

### DIFF
--- a/testing/features/automated_release_process.feature
+++ b/testing/features/automated_release_process.feature
@@ -21,3 +21,7 @@ So that so that I can prevent errors and make manual changes after the process h
 As a user of the design system in a Citizens Advice product team
 I need to be aware of new design system releases and what has changed since the last release
 So that I have an up-to-date understanding of the Citizens Advice design system to apply to my work
+
+As a design system developer
+I need to be able to specify whether the release is a major, minor or patch release
+So that users of the design system know what types of changes are included in the release
--- a/testing/features/automated_release_process.feature
+++ b/testing/features/automated_release_process.feature
@@ -6,9 +6,9 @@ As a design system developer
 I need to be able to release updates to the design system using an automated process 
 So that I can save time and improve the consistency of release artefacts and processes
 
-As a design system developer 
-I need to be able to specify that a design system release is "pre-release" or "stable release"
-So that users understand that the design system is ready for either the development or production environments
+As a developer using the design system components
+I need to know whether a given release is "pre-release" or "stable"
+So that I know whether to use it in production-ready code
  
 As a design system developer 
 I need to have the option to release design system updates manually
--- a/testing/features/automated_release_process.feature
+++ b/testing/features/automated_release_process.feature
@@ -2,9 +2,13 @@ Feature: Automated Release Process
 
 This feature is to automate the design system release process, removing repeatable manual developer actions. 
 
+As a design system developer
+I need to be able to release updates to the design system using an automated process 
+So that I can save time and improve the consistency of release artefacts and processes
+
 As a design system developer 
-I need to be able to specify that a design system release is Alpha or Public
-So that users understand that the design system is ready for either the public or develop environments
+I need to be able to specify that a design system release is "pre-release" or "stable release"
+So that users understand that the design system is ready for either the production or development environments
  
 As a design system developer 
 I need to have the option to release design system updates manually
@@ -14,6 +18,6 @@ As a design system developer
 I need to be able to stop the automated design system release process
 So that so that I can prevent errors and make manual changes after the process has started
  
-As a user of the design system
-I need to be aware of new design system releases and any changes
+As a user of the design system in a Citizens Advice product team
+I need to be aware of new design system releases and what has changed since the last release
 So that I have an up-to-date understanding of the Citizens Advice design system to apply to my work
--- a/testing/features/automated_release_process.feature
+++ b/testing/features/automated_release_process.feature
@@ -17,3 +17,6 @@ So that so that I can... prevent errors / make manual changes after the process 
 As a user of the design system [who?]
 I need to be aware of new design system releases and any changes
 So that I have an up-to-date understanding of the Citizens Advice design system to apply to my work
+
+Test TesT
+Teotarjgnwprigj
--- a/testing/features/automated_release_process.feature
+++ b/testing/features/automated_release_process.feature
@@ -2,7 +2,7 @@ Feature: Automated Release Process
 
 This feature is to automate the design system release process, removing repeatable manual developer actions. 
 
-As a design system developer
+As a maintainer of the design system components
 I need to be able to release updates to the design system using an automated process 
 So that I can save time and improve the consistency of release artefacts and processes
 
@@ -10,11 +10,11 @@ As a developer using the design system components
 I need to know whether a given release is "pre-release" or "stable"
 So that I know whether to use it in production-ready code
  
-As a design system developer 
+As a maintainer of the design system components
 I need to have the option to release design system updates manually
 So that I am not reliant on 3rd party services to release updates automatically e.g. Github / Jenkins
  
-As a design system developer 
+As a maintainer of the design system components
 I need to be able to stop the automated design system release process
 So that so that I can prevent a bad release from being published after the process has started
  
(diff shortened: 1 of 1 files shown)


testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Automated Release Process
+
+This is to automate the design system removing manual developer actions. 
+    
+As a design system developer 
+I need to be able to specify that a design system release is Alpha or Public
+So that users understand that the design system is ready for either the public or develop environments
```
thomasgower, line 7: - who are the users of the design system / do we need to specify here

testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Automated Release Process
+
+This is to automate the design system removing manual developer actions. 
+    
+As a design system developer 
+I need to be able to specify that a design system release is Alpha or Public
```
thomasgower, line 6: - specify release number / define as a full or partial release? Not sure what the right terms are here

testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Automated Release Process
+
+This feature is to automate the design system release process, removing repeatable manual developer actions. 
+
```
thomasgower, line 4: Adding here as a comment in case we need to add the original requirements on Jira ticket from Dan in some form:
- As a Design System Developer I can release the Design System using an automated process so that the release artifacts and process are consistent.
- As a Design System Developer I can release the Design System using an automated process so I can save time and do something else whilst the release is happening.

testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,19 @@
...
+I need to be able to specify that a design system release is Alpha or Public
+So that users understand that the design system is ready for either the public or develop environments
+ 
+As a design system developer 
+I need to have the option to release design system updates manually
+So that I am not reliant on 3rd party services to release updates automatically e.g. Github / Jenkins
+ 
+As a design system developer 
+I need to be able to stop the automated design system release process
+So that so that I can prevent errors and make manual changes after the process has started
```
thomasgower, line 15: I'm unclear on the primary benefit of being able to stop the automated process, if this is a requirement. Have put some options here, any thoughts?

testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,19 @@
...
+ 
+As a user of the design system
+I need to be aware of new design system releases and any changes
```
thomasgower, line 18: This is drawn from the 'ideal process' but may be a requirement we want to capture.

User could be defined more narrowly as: "as a member of the #design-system slack channel" if this is too vague currently

testing/features/automated_release_process.feature
davidrapson: This is a good question. We're using https://github.com/citizensadvice/design-system/releases as the central point for this and that uses the phrase "pre-release". So perhaps "is a stable or pre-release" might do?

testing/features/automated_release_process.feature
davidrapson: For the purposes of the code part (the component library) our main users are other product teams in the organisation. e.g. Content platform, casebook, digital workplace and energy teams mainly. But in theory any product team. 

But there's also users in the sense of _software_ that depends on the design system code.

davidrapson: Not sure if we want to cover the wider design system community here or if this should be technology-specific.

testing/features/automated_release_process.feature
davidsauntson: I didn't use helpful language when we were discussing it - I would go with `pre-release` and `stable release`.

testing/features/automated_release_process.feature
davidsauntson: I think changing `public or develop` to `production or development` is probably more accurate but I am splitting hairs.

testing/features/automated_release_process.feature
davidsauntson: Fine to add these I think - you can combine them into one tho?

testing/features/automated_release_process.feature
davidsauntson: I think `what has changed since the last release` instead of `any changes` would be more explicit.

testing/features/automated_release_process.feature
davidsauntson: I guess we would want to do this if we have made some sort of mistake that hasn't been captured before the release starts, eg a wrong version number or mis-labelling as `pre-release` instead of `stable release` perhaps?

So maybe change `prevent errors and make manual changes` to `prevent a bad release from being published` - @davidrapson  what do you think?

davidsauntson (commented):
🎈 Great!

I think `users` is fine as a term - they could be other devs, designers, product specialists etc.

I think we need another scenario that is very similar to the `pre-release` / `stable-release` one where the developer can specify what kind of version increment the release is, in-line with semver. Something like:

```
As a design system developer
I need to be able to specify whether the release is a major, minor or patch release
So that users of the design-system know what types of changes are included in the release
```

testing/features/automated_release_process.feature
thomasgower: Yep - can combine into 1. Only reservation was about specifying the solution here but I suppose it's the core of the feature!

testing/features/automated_release_process.feature
thomasgower: Thanks @davidsauntson - have changed to "production or development" - gotta split those hairs!

@davidrapson The way I understood it for this one it would be the product teams needing to be notified of changes.

In terms of the way other software (as users) depends on the code, would automating the release process change anything here in how it interacts with the design system - ie is there a requirement from a "user = software" perspective? I'd assume not, but am still getting to grips with the different pieces of the puzzle!

testing/features/automated_release_process.feature
thomasgower: updated, thanks.

testing/features/automated_release_process.feature
thomasgower: added below

testing/features/automated_release_process.feature
thomasgower: Updated to "prevent a bad release from being published"

testing/features/automated_release_process.feature
thomasgower: Updated, thanks

davidsauntson (approved):
✨

testing/features/automated_release_process.feature
```diff
@@ -0,0 +1,27 @@
...
+
+This feature is to automate the design system release process, removing repeatable manual developer actions. 
+
+As a design system developer
+I need to be able to release updates to the design system using an automated process 
+So that I can save time and improve the consistency of release artefacts and processes
+
+As a design system developer 
+I need to be able to specify that a design system release is "pre-release" or "stable release"
+So that users understand that the design system is ready for either the development or production environments
```
pedantic-git, line 11: Is the developer the user here, or is the person consuming the design system the user? E.g.

```
As a developer using the design system components
I need to know whether a given release is "pre-release" or "stable"
So that I know whether to use it in production-ready code
```

pedantic-git (commented):
Approving anyway, but a couple of thoughts.

1. As well as this, I wonder if "As a design system developer" is vague. (Does it mean someone developing the components or developing _using_ the components? What about contributors from outside the core maintenance team?) Maybe it should be "As a maintainer of the design system components". (Aside: I also think we should probably start referring to this repo as the "design system components" or "design system component library", because the design system is a process, a state of mind, and a group of people, rather than just this piece of technology.)
2. See my comment inline.

pedantic-git (approved):

testing/features/automated_release_process.feature
thomasgower: Yes - good question and I think your version better captures the core need, will update.

thomasgower: re you're first point I'll also update to be "As a maintainer of the design system components"

On the repo renaming, I think the more specific we can be the better, but will leave to others to change the name in case I break it!

pedantic-git: Oh yeah - I don't think we should actually rename the repo... just what we refer to it as in these requirements. But that's one for another day!

PR #1151, opened 2021-05-19
Correct case for dependency-name include

davidrapson:
So close and yet so far…one more try.


davidsauntson (approved):

davidrapson:
For our future selves, although the validator doesn't seem to run on pull requests changing the config this https://dependabot.com/docs/config-file/validator/ supports the version 2 syntax for one-off checks.

davidrapson:
image

PR #1168, opened 2021-05-24
Scope table styles to cads-table

davidrapson:
Noticed whilst looking at a public-website issue with tables that we include the top-level `table` element as part of our tables styles.

For our components we probably want to narrow the scope wherever possible. Including the base table element can cause knock-on effects if you're not expecting it.

This can be seen even in the storybook docs, where having element level styles conflicts with storybook's own tables.

Before | After
------------ | -------------
image Hybrid storybook / design system table styles | image Unaffected storybook table styles

I think this is pretty safe to do based on our current usage, but shout if I'm missing something.


davidsauntson (approved):
I'm just really upset that there's no table emoji

davidrapson:
> I'm just really upset that there's no table emoji

(╯°□°)╯︵ ┻━┻

PR #1170, opened 2021-05-26
chore/remove slash divs NP-2109

davidsauntson:
Replace slash division with `math.div` as slash division will be removed in dart sass 2.


davidrapson (approved):
👍 

But/and I reckon we should add a note to the docs somewhere that this only works with dart sass.

That's fine as it looks like all the other implementations are deprecated at this point. But it does mean that you need to be running at least version `5.2.2` of webpacker as previous versions bundle `node-sass`.

Get this error if I explicitly set the version in `demo/` to ~`5.1`:

image

davidrapson:
Actually it looks like the demo app is on `5.2.1` so we might want to bump that here at the same time.

davidrapson:
Hooray this failed 🦜 

image

davidrapson:
Oh, of course this will need the `@rails/webpacker` node module updating too as that's what actually loads dart-sass.

davidrapson (approved):
⭐

PR #1171, opened 2021-05-26
Add demo app workflow

davidrapson:
Prompted by https://github.com/citizensadvice/design-system/pull/1170 wanted to try adding a workflow to install the demo and and run a webpacker build as a smoke test for using the design system with a full Rails app.


davidsauntson (approved):
I am a fan 🦮

davidrapson:
@davidrapson I think this will cause https://github.com/citizensadvice/design-system/pull/1170 to fail now. But if you run a `bundle update` in the demo app that should bring webpacker up to date. Kinda useful to see it fail first though.

PR #1173, opened 2021-05-27
chore: add codeowners

davidsauntson:
Add The Davids(TM) as global code owners.  When this is merged they should be added as reviewers to any PR with `master` as the target.


davidrapson (approved):
👍 Mind raising a test PR after this is merged to check the behaviour?

PR #1186, opened 2021-06-01
Clarify storybook bug

luke-hill:
Ronseal PR

### DIFF
--- a/testing/features/components/navigation.feature
+++ b/testing/features/components/navigation.feature
@@ -1,3 +1,4 @@
+# NB: Due to a Storybook issue, these tests won't run on Old Safari / iOS
 @not_ios13 @not_ios12 @not_safari
 Feature: Navigation component
 


testing/features/components/navigation.feature
```diff
@@ -1,3 +1,4 @@
+# NB: Due to a Storybook issue, these tests won't run on Old Safari / iOS
```
davidrapson, line 1: What's the actual issue? Longer term we could probably start running these tests against the `demo/` app which is a much simpler and more realistic environment than storybook but we should chat about that first 🤓

davidrapson (approved):
👍 


(Small aside, I'm always unsure what you mean by "Ronseal PR", if possible adding a real description even if the change _feels_ obvious is super helpful. Thanks 🙏)

testing/features/components/navigation.feature
luke-hill: Go for it. Happy to let you guys decide on this. You probably know better.

P.S. Showing my age but: https://www.youtube.com/watch?v=OkGaq9xiQZY

davidrapson: Hah, I see. Didn't make that connection.

Yeah, that probably makes sense it'd be some compatibility thing. Storybook doesn't have a very good cross-browser story. No need to switch now, but as we've migrate more components into the Rails engine there will be a pretty decent test environment within the demo app. Something to come back to.

luke-hill: I've popped a new Phase I ticket in the backlog for this. Happy to tackle this whenever you feel is appropriate. I don't envisage it being a lot of work our end to migrate the tests (Probably just altering some page urls / ruby setup code).

PR #1189, opened 2021-06-01
Feat/typography NP-2027

davidsauntson:
Initial pass at the typography update as per the figma designs here: https://www.figma.com/file/J9qfbS9n1pUM2dpRQnTXUD/Design-system?node-id=4%3A3

Would be good to get lots of 👁️ on this!

@minniessuyuhuang there are some components that are not in the design system but explicitly mentioned in the figma files (OISC and on this page) - I will address those once this is merged and the design-system version updated in public website.


scss/4-elements/_typography.scss
```diff
@@ -1,59 +1,62 @@
 %cads-h1 {
   font-size: 2.5rem;
-  line-height: 1.2em; // Line-height is variable
+  line-height: 3.25rem;
   font-weight: $cads-font-weight__bold;
   color: $cads-font-colour__headings;
   margin: 0 0 $cads-spacing-7 0;
```
minniessuyuhuang, line 7: H1, h2, h3 have`letter-spacing: -0.3px` on all the breakpoints.

scss/4-elements/_typography.scss
```diff
@@ -1,59 +1,62 @@
...
   }
 }
 
 %cads-h2 {
   font-size: 2rem;
-  line-height: 1.25em; // Line-height is variable
+  line-height: 2.75rem;
   font-weight: $cads-font-weight__bold;
   color: $cads-font-colour__headings;
   margin: 0 0 $cads-spacing-4 0;
```
minniessuyuhuang, line 21: This needs `letter-spacing: -0.3px`.

scss/4-elements/_typography.scss
```diff
@@ -1,59 +1,62 @@
...
 
 %cads-h3 {
-  font-size: 1.4rem;
-  line-height: 1.25em; // Line-height is variable
+  font-size: 1.5rem;
+  line-height: 2rem;
   font-weight: $cads-font-weight__bold;
   color: $cads-font-colour__headings;
-  margin: 0 0 $cads-spacing-3 0;
+  margin: 0 0 $cads-spacing-4 0;
```
minniessuyuhuang, line 35: This needs `letter-spacing: -0.3px`.

scss/4-elements/_typography.scss
```diff
@@ -1,59 +1,62 @@
...
 
 %cads-paragraph {
   @include cads-typographic-scale-text();
 
   font-weight: $cads-font-weight__regular;
-  margin: 0 0 $cads-spacing-4 0;
+  margin: 0 0 $cads-spacing-5 0;
 }
 
 %cads-paragraph-small {
```
minniessuyuhuang, line 62: The small paragraph should have the same margin specs as the normal paragraph (24px margin-bottom) on all the breakpoints.

scss/4-elements/_typography.scss
```diff
@@ -1,59 +1,62 @@
...
-  line-height: 1.25em; // Line-height is variable
+  line-height: 2.75rem;
   font-weight: $cads-font-weight__bold;
   color: $cads-font-colour__headings;
   margin: 0 0 $cads-spacing-4 0;
 
   @include cads-media-breakpoint-down(sm) {
     font-size: 1.5rem;
+    line-height: 2rem;
     margin: 0 0 $cads-spacing-3 0;
```
minniessuyuhuang, line 25: This could be removed as the margin should be the same as the md and lg breakpoints (`margin: 0 0 $cads-spacing-4 0`)

scss/4-elements/_typography.scss
```diff
@@ -66,7 +69,7 @@
 h2,
 h3,
 h4 {
-  @include cads-adjacent-heading-margin($cads-spacing-6);
+  @include cads-adjacent-heading-margin($cads-spacing-7);
```
minniessuyuhuang, line 72: Interested to know why H1 is not included here?

minniessuyuhuang (commented):
Thanks David S, I have left some comments.

scss/4-elements/_typography.scss
davidsauntson: This mixin adds a margin to the top of these heading elements when they appear after certain elements in a `.cads-prose` container.  The 'certain elements' I refer to are listed in the mixin:  

https://github.com/citizensadvice/design-system/blob/3e48703f9f6aea9bbb86f22122874c0063847792/scss/2-tools/_typography.scss#L24

It's responsible for the margin on the top of this heading element, for example: https://www.citizensadvice.org.uk/immigration/staying-in-the-uk-if-youre-from-the-eu#h-check-when-to-apply-for-the-eu-settlement-scheme

`h1`s will never have any of these elements above them so we don't need the top margin (currently).

minniessuyuhuang (approved):

davidrapson (commented):
Gave this a a check with `npm link` and all looked pretty good aside from two things:

- As mentioned on chat, some lists in contentful are coming out as separate `ul`s so this new spacing makes that mistake pretty obvious (fine, we can fix it at the source)
- This also affects the spacing of the header making the main header noticeably deeper. Might be intentional, but maybe we need to adjust for that.

Otherwise looked good. One 🤓 about unitless line-height but just a thought nothing blocking.

Will defer to @minniessuyuhuang for the approval on this one.

davidsauntson:
@davidrapson @minniessuyuhuang 

I've updated to use unitless line-heights.  As they stand at the moment this leads to some awkward numbers - eg `1.3` recurring and in once instance an irrational!  Nothing against them personally but browsers will round in an odd and inconsistent way.   

For example:
```scss
h3 {
  font-size: 1.5rem;
  line-height: 1.33;
}
```
will result in a line-height of `31.92px`, which Chrome will round to `31px` (presumably because no one has taught it about significant figures). 

I think leaving the line-heights in `rem`s is probably the best option.  Whaddya reckon?

minniessuyuhuang:
> @davidrapson @minniessuyuhuang
> 
> I've updated to use unitless line-heights. As they stand at the moment this leads to some awkward numbers - eg `1.3` recurring and in once instance an irrational! Nothing against them personally but browsers will round in an odd and inconsistent way.
> 
> For example:
> 
> ```css-scss
> h3 {
>   font-size: 1.5rem;
>   line-height: 1.33;
> }
> ```
> 
> will result in a line-height of `31.92px`, which Chrome will round to `31px` (presumably because no one has taught it about significant figures). You can see the differences in the failed backstop report: https://jenkins.devops.citizensadvice.org.uk/job/design-system/job/PR-1189/4/BackstopJS_20Report/
> 
> I think leaving the line-heights in `rem`s is probably the best option. Whaddya reckon?

Thanks @davidsauntson. I think for now we should just stick to `rem` to keep things simple. If we want to use unitless, we can do it in another ticket (and maybe update the design in Figma to simplified the decimal points).

davidrapson:
> I think leaving the line-heights in rems is probably the best option. Whaddya reckon?

ooh yeah that makes sense seeing it context. I guess that makes most sense when we're setting proportions (1.3 / 1.5) etc. rather than trying to hit a specific scale. Those numbers get a bit gnarly don't they. Cheers for trying it though 👍

davidrapson (approved):

PR #1191, opened 2021-06-02
Update browser versions being tested

luke-hill:
Whilst trying to debug the IE / Big Sur issues (Raised tickets for both on Browserstack), I figured I'd bump the versions being tested


davidrapson (approved):
👍 But worth forcing things to run on this PR ahead of merging?

luke-hill:
OOoh yes. Good point. Always a good idea.

luke-hill:
Yep still all good. Will remove commit and merge

PR #1193, opened 2021-06-03
Feat/breadcrumb view component DS-7

davidsauntson:
Add the breadcrumbs component to the design system engine:
- new view component
- rspec
- previews
- updates `mdx` docs
- removes hard-coding and translating 'Home' link (this will be duplicated in old design system HAML in a different PR)

### DIFF
--- a/engine/app/components/citizens_advice_components/breadcrumbs.html.haml
+++ b/engine/app/components/citizens_advice_components/breadcrumbs.html.haml
@@ -1,0 +1,10 @@
+%nav{ class: "cads-breadcrumbs cads-breadcrumbs--#{@type.to_s.dasherize}", "aria-label": "breadcrumbs" }
+  %ul.cads-breadcrumbs__list
+    - @links.each_with_index do |link, index|
+      %li.cads-breadcrumbs__crumb
+        - if index == links.size - 1
+          %span.cads-breadcrumb{ aria: { current: "location" } }
+            = link[:title]
+        - else
+          %a.cads-breadcrumb{ href: link[:url] }
+            = link[:title]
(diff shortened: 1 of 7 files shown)


engine/previews/citizens_advice_components/breadcrumbs_preview.rb
```diff
@@ -0,0 +1,77 @@
...
+      render(
+        CitizensAdviceComponents::Breadcrumbs.new(
+          type: :collapse,
+          links: [
+            {
+              title: "Home",
+              url: "/"
+            },
+            {
+              title: '
```
davidrapson, line 57: I this meant to have the line-break?

davidrapson (approved):
👍 Looks great. Couple of 🤓 suggestions but nothing major.

engine/app/components/citizens_advice_components/breadcrumbs.html.haml
```diff
@@ -0,0 +1,10 @@
+%nav{ class: "cads-breadcrumbs cads-breadcrumbs--#{@type.to_s.dasherize}", "aria-label": "breadcrumbs" }
+  %ul.cads-breadcrumbs__list
+    - @links.each_with_index do |link, index|
```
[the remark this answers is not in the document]
davidsauntson: Ah yeah that makes sense.  So `attr_reader` generates a getter behind the scenes and then I'm using the instance variable that effectively ignores it - is that correct?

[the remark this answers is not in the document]
davidrapson: Yep, spot on.

engine/previews/citizens_advice_components/breadcrumbs_preview.rb
davidsauntson: No that was rubocop

engine/previews/citizens_advice_components/breadcrumbs_preview.rb
```diff
@@ -54,8 +54,7 @@ def long
               url: "/"
             },
             {
-              title: '
-              My Grandmother Asked Me to Tell You She’s Sorry',
+              title: "My Grandmother Asked Me to Tell You She’s Sorry",
```
davidrapson, line 57: It's OK 👵, I forgive you.

engine/spec/components/breadcrumbs_spec.rb
```diff
@@ -68,4 +72,24 @@
       end
     end
   end
+
+  context "links are passed with the old style hash format" do
```
davidrapson, line 76: ❤️

davidrapson (approved):
👍 Looks good to me

We might want to come back and generate the example files and update storybook to use them; but we don't need to do that here.

PR #1194, opened 2021-06-03
Sync ruby versions

davidrapson:
In https://github.com/citizensadvice/design-system/pull/1187 ruby was updated to `2.7.3` but only in some places.

This builds on that and makes sure we use the same version everywhere.

In the process I've merged the engine workflow in with the main one as they don't need to be separate workflow files and it allows us to share the version more easily.


.github/workflows/ci-workflow.yml
```diff
@@ -39,6 +40,23 @@ jobs:
...
+  engine:
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        working-directory: engine
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
```
davidrapson, line 52: This version doesn't come from the `ruby-version` largely because we don't specific a ruby version file in the engine directory, but also because the engine isn't "locked" to a version of ruby like the top-level build environment is. In fact longer term it would be good for this to run as a matrix build and test against `2.7.2`, and `3` as well for example.

davidrapson: Sneaky experiment showing what I mean here: https://github.com/citizensadvice/design-system/pull/1195

luke-hill: Oki. I mean I guess it's fine. Given this in theory will be much faster than jenkins I'm definitely not against it.

The only caveat to this, is all "consumers" of DS currently is basically PW and the smart meter tool. Both running 2.7.2 or 2.7.3 I believe. But definitely happy to give this a whirl.

luke-hill (approved):

.github/workflows/ci-workflow.yml
davidrapson: Not 100% sure I follow the concern here. As far as I understand here you'd never lock a gems ruby version to a specific point release. We use `required_ruby_version` to specify a boundary, but otherwise flexible. Here we _are_ testing against `2.7.3` explicitly; but it's a pretty reasonable to be testing the engine against upcoming versions too.

FWIW we're already running this exact workflow in actions, I've just moved where it lives at the same time as syncing the version up.

luke-hill: I'm saying atm we don't have people not running 2.7 using this. So it's fine to add this, but it's not a priority.

davidrapson: Oh I see yeah, with you there. For now this does just test 2.7.3.

davidsauntson (approved):

PR #1196, opened 2021-06-04
Browserstack GitHub actions

davidrapson:
Re-instates Browserstack workflow. Unsurprisingly this fails on a credentials error as the browser stack keys will need updating in the secrets.


davidsauntson (approved):
👏🏻  Love that journey for us.
https://media.giphy.com/media/JRUbQVADfqzrFnOkZr/giphy.gif

davidrapson (commented):
I can't approve this as I raised it but **so** happy with this. I'd check two things with @luke-hill before we hit merge:

1. Happy sticking with the `latest` versions for now (my preference) or should we pin to the same versions that are in the Jenkinsfile
2. Do the build names in the BrowserStack UI look OK?

davidsauntson:
> I can't approve this as I raised it but **so** happy with this. I'd check two things with @luke-hill before we hit merge:
> 
> 1. Happy sticking with the `latest` versions for now (my preference) or should we pin to the same versions that are in the Jenkinsfile
> 2. Do the build names in the BrowserStack UI look OK?

1 - yeah I prefer the latest I think but will merge and check in with Luke when he's back tomorrow
2 - yup I peeped and they and both the build and sessions names are 🆗

.github/workflows/browserstack-workflow.yml
```diff
@@ -0,0 +1,74 @@
+name: BrowserStack
+
+# BrowserStack tests are run against the deployed GitHub pages URL
+# The closest event we can use to trigger when these are updated is page_build.
+# on: [page_build]
+on: [push]
```
davidsauntson, line 6: @davidrapson  we want this to on page build, right?

davidrapson: Oh good point. Yeah that's the ideal event I think. A build scoped to `master` is fine as a fallback if `page_build` doesn't do what we think but in theory it's closest to the thing we're testing in BrowserStack.

davidsauntson: Cool, I just checked the docs again and it's a push to any github pages enabled branch so ✅

luke-hill:
There's a few things I want to change from this. But given this is just running alongside things for now it can stay as is (As this will run everything twice, once for Jenkins once for GHA I believe).

Do we have a follow-up ticket to remove the Jenkinsfile? If so we can add the required changes there. Some of the changes I'll want to get done soon as they break some existing features we kinda relied on.

PR #1207, opened 2021-06-08
feature: Targeted Content (Audit of Variants) [NP-2117]

luke-hill:
Added a couple of missing scenarios for the Targeted Content Variants

### DIFF
--- a/testing/features/components/step_definitions/common_steps.rb
+++ b/testing/features/components/step_definitions/common_steps.rb
@@ -55,3 +55,7 @@ Then("I can collapse the search bar") do
 
   expect(@component).not_to have_search_field
 end
+
+Then("a/an {string} label is present above the title") do |label|
+  expect(@component.label.text).to eq(I18n.t("cads.callout.#{label.downcase}"))
+end
--- a/testing/features/components/step_definitions/common_steps.rb
+++ b/testing/features/components/step_definitions/common_steps.rb
@@ -55,7 +55,3 @@ Then("I can collapse the search bar") do
 
   expect(@component).not_to have_search_field
 end
-
-Then("a/an {string} label is present above the title") do |label|
-  expect(@component.label.text).to eq(I18n.t("cads.callout.#{label.downcase}"))
-end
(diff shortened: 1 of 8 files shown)


testing/features/components/step_definitions/common_steps.rb
```diff
@@ -55,3 +55,7 @@
 
   expect(@component).not_to have_search_field
 end
+
+Then("a/an {string} label is present above the title") do |label|
+  expect(@component.label.text).to eq(I18n.t("cads.callout.#{label.downcase}"))
```
davidrapson, line 60: I know we've circled round this in the past but could we avoid using the internal I18n keys in the tests. Looks like `expect(@component.label.text).to eq(label))` might have a comparable effect where we pass the expected label directly?

davidrapson (changes requested):
👍 Nice to get the extra scenarios covered. Small request about how we're testing the label text. Although I can see you're mostly moving that code around so shout if you'd rather revisit that.

testing/features/components/step_definitions/common_steps.rb
luke-hill: For some of them (such as here), we can do that. But for some we need to keep it in. Up to you what we do for consistency here.

davidrapson: Let's take it out please. We probably should catch up about language tests specifically in the design system because the new engine / view components change the boundaries of that a little bit anyway.

testing/features/components/step_definitions/targeted_content_steps.rb
```diff
@@ -46,3 +58,13 @@
   expect(@component.additional_information.close.text)
     .to eq(I18n.t("cads.targeted_content.close_label"))
 end
+
+Then("an Adviser label is present in the expandable pane") do
+  expect(@component.label.text).to eq("Adviser")
```
davidrapson, line 63: ❤️

davidrapson (approved):
👍 LGTM

PR #1208, opened 2021-06-08
feature: Input Audit (Additional Variant) [NP-2115]

luke-hill:
Performs the Audit of code / scenarios for the Input form/s


davidrapson (commented):
I'm not sure what this is trying to test? The step looks like it's checking if a list of class names are unique, what are you looking to check there?

luke-hill:
You'll notice as this scenario is written differently I didn't want to do the whole verbose steps, so I just opted to compress the steps to stipulate there are different types of Input, good for clarifying as a spec.

davidrapson:
I guess what I mean is this block:

```
input_types = @form.inputs.map { |input| input["class"] } 
expect(input_types).to be_uniq
```

Is checking for uniq elements. So this test can be satisfied even if the class names are nonsense; just so long as they are unique. Is there another way we could frame this so it asserts something more direct? Or is a visual test sufficient given the material change here is a visual one?

luke-hill:
I would say a combination of what you've said. I think visual automation should do the bulk of the heavy lifting here. I wouldn't want to test that they look different as previously doing that (Colour testing), was pushed back and into the VRTs.

The fact they have different styling also makes test maintenance easier, we just check they're different (in some way)

davidrapson:
Yeah, that makes sense. I definitely don't think we should be doing visual tests in the selenium tests.

To scratch my own itch [I made a commit](https://github.com/citizensadvice/design-system/commit/115af65e3d0661c0aeb9f054c9ae48e3769b8bfc) based off this branch which renamed the size classes to random names in the examples. Wanted to verify my assumption that the steps here would pass for anything as long as the class names are unique.

image
https://github.com/citizensadvice/design-system/actions/runs/922275607

As you can see the selenium tests pass as the changes meet the loose conditions that the classnames are unique, despite the fact that the usage is now wrong. Visual regression tests catch this.

Fairly comfortable that the visual test is sufficient; and that any future rspec tests we add when we migrate form inputs to the engine can cover the component interface side of things. As it stands this scenario is kinda like the button scenario in that it's testing something with loose guarantees and only really in the docs.

What might be useful is if you could demonstrate a scenario, or version of these steps which shows a failure?

testing/features/forms/step_definitions/input_steps.rb
```diff
@@ -51,3 +51,10 @@
 Then("the text box has a value of {string}") do |text|
   expect(@form.input.value).to eq(text)
 end
+
+Then("I am able to use Inputs of a variety of widths") do
+  @form = Input::FixedWidths.new.tap(&:load)
+  input_types = @form.inputs.map { |input| input["class"] }
+
+  expect(input_types).to be_uniq
```
davidrapson, line 59: Sorry! One more 🤓 question. Does this `be_uniq` come from the fact that there is an `Array#uniq?` core extension in the `ca_testing` gem?

luke-hill: Sorry I figured I may as well explain how this works. And in general metaprogramming.

Ruby heavily uses metaprogramming or programming-on-the-fly. In it you don't know what the code will be until **after** runtime or compilation time (Ruby kinda compiles). As such often we like to do things like this in ruby.

RSpec if loaded (Specifically rspec expectations and rspec matchers), will then scan your codebase for all booleans (Things ending in `?`), and specifically will generate a series of rspec matchers with associated names. The name for all matchers becomes as follows.

- Any boolean named `has_xxx?` creates a matcher called `have_xxx`
- Any boolean named `xxx?` creates a matcher called `be_xxx`

There are also a whole host of other things. RSpec heavily promotes metaprogramming / BDD so it kinda fits hand in glove with step defs.

luke-hill:
Yep you've got it spot on. So the tests here would catch for something small. That's the purpose of the tests here. Also as they serve some nice documentation purposes too.

testing/features/forms/step_definitions/input_steps.rb
davidrapson: Cheers. Relatively familiar with that pattern. Not 100% sure why we maintain our own core extensions; but good to know that assumption is correct.

davidrapson:
Happy to close this one then? On the documentation point, given the thing that these tests run agains _is_ documentation I think we're fine as is. Unless these tests can assert something that would usefully fail.

luke-hill:
So I'm confused. I was thinking the opposite? Like we're good to merge this now>?

testing/features/forms/step_definitions/input_steps.rb
luke-hill: This is a method that doesn't exist in any library, also these kinds of methods (The ones written), are good extensions to have, so they're added to the ca_testing gem (Alongside patches and driver methods).

davidsauntson:
I think you both agree that all the test does is check to see if the input's class names are unique.

Looking at the code in the input haml file, there is nothing that either verifies what is passed into as `size` and there is nothing that mandates `size` be present either.  So this test isn't really covering any of the code in the haml file.  It's not assessing the actual size of the inputs either, so it's also not testing any of the CSS code.

IMHO there are three things we need to check for the input as it relates to `size`, and these are:
 - that not passing in a `size` parameter renders the input at full width
 - that passing in a `size` parameter as specified in the docs renders an input of the correct width
 - that passing in an invalid `size` parameter renders the input at full width (ie, does nothing)

The code that defines this is covered in the CSS and so I think the VRTs are sufficient here.

davidrapson (approved):
Approving as I guess this is a tiny description change at this point. The discussion was instructive though.

PR #1209, opened 2021-06-08
feature: Button Audit [NP-2113]

luke-hill:
Added simple spec to validate link button w3c props

### DIFF
--- a/testing/features/components/buttons.feature
+++ b/testing/features/components/buttons.feature
@@ -26,3 +26,10 @@ Feature: Button components
 
     Scenario: Ensure button is of correct type
       Then the Tertiary button acts as a button
+
+  Rule: Link Buttons
+    Background:
+      Given the Link Button components are on the page
+
+    Scenario: Ensure buttons are of correct type
+      Then the Link buttons act as buttons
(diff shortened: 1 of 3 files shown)


davidrapson (commented):
Interesting. The button "component" is really a single set of class names and there's not a distributed template for them. You can also legitimately use the classes on a range of different types. More so for the regular `cads-button` class but the point still stands.

So "correct" type for all these button specs is really very contextual and down to usage. All this test will do is assert that the docs have a button with type `button`. Not sure this test tells us much. Wonder if these sorts of checks are better scoped to tests in products themselves?

luke-hill:
Hmmm interesting. Doesn't that mean then that it would fail a11y checks down the line? Also these tests already existed so I just updated them as per the audit tasks I had created.

What do you advise here? I'd like something in here, even if it's rudimentary as I'm trying to cover all the components/forms as the tests run subsecond.

davidrapson (approved):
> Doesn't that mean then that it would fail a11y checks down the line?

It depends on usage. I guess bear in mind that these are fundamentally just a set of CSS styles. So what semantics are used are down to individual projects. In this case our docs aim to pave the path.

Approving and will leave the call to you. This test is ultimately pretty harmless.

luke-hill:
It's a good point you raise. I'll merge for now, but yeh this is a good learning point for me. Thanks for the context.

PR #1215, opened 2021-06-10
feature: Breadcrumbs Test Audit [NP-2112]

luke-hill:
It appears that some of these are identically styled and functional (So I just removed duplicate tests and placed notes)

If I'm wrong and they are distinct components I'll re-add the tests in.

### DIFF
--- a/testing/features/components/breadcrumbs.feature
+++ b/testing/features/components/breadcrumbs.feature
@@ -5,6 +5,7 @@ Feature: Breadcrumbs component
 
   Rule: Default Breadcrumb
     This collapses down when viewing on small screen sizes (Mobile's e.t.c.)
+    This is the same as the Long Titles Breadcrumb component
     Background:
       Given a Default Breadcrumbs component is on the page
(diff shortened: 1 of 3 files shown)


testing/features/components/breadcrumbs.feature
```diff
@@ -4,6 +4,8 @@ Feature: Breadcrumbs component
   previous content pages they have traversed through
 
   Rule: Default Breadcrumb
+    This collapses down when viewing on small screen sizes (Mobile's e.t.c.)
```
davidsauntson, line 7: typo nitpick: this should be `(mobiles etc)`

davidsauntson (approved):
Nice!

You are correct - they are the same component (one just requires a wrapping div if used at the top of the page so no need to test them separately).

testing/features/components/breadcrumbs.feature
luke-hill: it should!

PR #1229, opened 2021-06-14
Fix/header link margins  NP-2027

davidsauntson:
Fixes the header link margin issue introduced in the typography update.


davidrapson (approved):
👍

PR #1230, opened 2021-06-14
fix: remove hard-coded link from breadcrumbs DS-7

davidsauntson:
Removes the hard-coded home link from the 'old' haml partials for the smart metering team.  Updates docs and examples.


davidrapson (approved):
👍 Nice, thanks for fixing this here too.

PR #1238, opened 2021-06-15
fix: add missing env var

davidsauntson:
So I _think_ this will fix the missing `BROWSERSTACK_BUILD_NAME` key issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -275,7 +275,8 @@ def define_browserstack_tests() {
       withEnv([
         "BROWSER=${it.browser}",
         "BROWSERSTACK_CONFIGURATION_OPTIONS=${it.config}",
-        'BROWSERSTACK=true'
+        'BROWSERSTACK=true',
+        "BROWSERSTACK_BUILD_NAME=Design system (Jenkins) ${env.BRANCH_NAME}_${getSha()}"
       ]) {
         withTestingNode("Browser Tests for ${it.browser} on ${it.config}", true, isMobile) {
           try {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -275,7 +275,8 @@ def define_browserstack_tests() {
       withEnv([
         "BROWSER=${it.browser}",
         "BROWSERSTACK_CONFIGURATION_OPTIONS=${it.config}",
-        'BROWSERSTACK=true'
+        'BROWSERSTACK=true',
+        "BROWSERSTACK_BUILD_NAME=Design system (Jenkins) ${env.BRANCH_NAME}_${getSha()}"
```
davidrapson, line 279: As `DOCKER_TAG=${env.BRANCH_NAME}_${getSha()}` you could arguably do ` "BROWSERSTACK_BUILD_NAME=Design system (Jenkins) ${env.DOCKER_TAG}"`. But this is a pretty short-lived fix so 🤷‍♂️

davidrapson (approved):
Looks right to me. But worth confirming the syntax works by forcing the browserstack tests to run on this PR.

Jenkinsfile
davidsauntson: forcing a re-run here: https://github.com/citizensadvice/design-system/pull/1239
looks like a goer: https://jenkins.devops.citizensadvice.org.uk/blue/organizations/jenkins/design-system/detail/PR-1239/1/pipeline

davidsauntson: it failed but only because no components could  load on iPhone IIX: The Last iPhone.

PR #1243, opened 2021-06-16
fix: asset size reported in MiB not MB

davidsauntson:
Can't quite believe I spent time on this but the asset hyperlink view component reports 6,444,516 bytes as 6.15MB.  I sort of don't really care about the MiB / MB holy war but after you've downloaded the file, your OS will report it as 6.4MB rather than 6.15MB.  So I think they should match.

http://localhost:3000/immigration/applying-for-settled-status/

image

image

### DIFF
--- a/engine/app/components/citizens_advice_components/asset_hyperlink.html.haml
+++ b/engine/app/components/citizens_advice_components/asset_hyperlink.html.haml
@@ -2,4 +2,4 @@
   = description
   %span.cads-asset-type<
     %span.cads-icon_file
-    \ #{number_to_human_size(size)}
+    \ #{number_to_human(size, precision: 2, units: { unit: 'B', thousand: 'KB', million: 'MB' })}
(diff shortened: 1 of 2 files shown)


davidrapson (approved):
Approving as it's a judgement call. But I think, sadly, Mac OS and Windows use opposite units so you can never win. I'd err towards the one that's accurate to Windows users personally.

Probably worth checking, just because the previous version is simpler and defers to Rails defaults.

davidsauntson:
Oh! I read somewhere that windows would also be 6.4MB but I've tested it and you are of course correct!  Since it's not consistent I agree we should leave it as it is then.

davidrapson:
> Oh! I read somewhere that windows would also be 6.4MB but I've tested it and you are of course correct! Since it's not consistent I agree we should leave it as it is then.

It's a nightmare right! I think Mac OS changed with snow leopard or something.

PR #1244, opened 2021-06-16
Fix/breadcrumb options

davidsauntson:
Adds some options into the `Breadcrumbs` view component:
 - `current_page` if the breadcrumbs represent the location of the current page (controls `aria-current`)
 - `full_width` if the breadcrumbs should take up the full width of the viewport (wraps in `.cads-breadcrumbs-wrapper`

Also adds `pry` into development dependencies and makes rubocop count hashes and arrays as 1 line long when assessing class and method length

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,6 +21,7 @@ Style/MixinUsage:
     - 'scripts/haml.rb'
 
 Metrics/MethodLength:
+  CountAsOne: ['array', 'hash']
   Exclude:
     - 'scripts/haml.rb'
     - 'engine/previews/**/*.rb'
@@ -33,3 +34,6 @@ Metrics/BlockLength:
     - 'spec/**/*.rb'
     - 'engine/spec/**/*.rb'
     - 'demo/spec/**/*.rb'
+
+Metrics/ClassLength:
+  CountAsOne: ['array', 'hash']
--- a/engine/app/components/citizens_advice_components/breadcrumbs.html.haml
+++ b/engine/app/components/citizens_advice_components/breadcrumbs.html.haml
@@ -1,10 +1,10 @@
-- haml_tag_if @full_width, ".cads-breadcrumbs-wrapper" do
+- haml_tag_if full_width?, ".cads-breadcrumbs-wrapper" do
   %nav{ class: "cads-breadcrumbs cads-breadcrumbs--#{type.to_s.dasherize}", "aria-label": "breadcrumbs" }
     %ul.cads-breadcrumbs__list
       - links.each_with_index do |link, index|
         %li.cads-breadcrumbs__crumb
           - if index == links.size - 1
-            %span.cads-breadcrumb{"aria-current": ("location" if @current_page) }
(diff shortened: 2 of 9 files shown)


engine/app/components/citizens_advice_components/breadcrumbs.html.haml
```diff
@@ -1,10 +1,11 @@
...
-  %ul.cads-breadcrumbs__list
-    - links.each_with_index do |link, index|
-      %li.cads-breadcrumbs__crumb
-        - if index == links.size - 1
-          %span.cads-breadcrumb{ aria: { current: "location" } }
-            = link[:title]
-        - else
-          %a.cads-breadcrumb{ href: link[:url] }
-            = link[:title]
+- haml_tag_if @full_width, ".cads-breadcrumbs-wrapper" do
```
davidrapson, line 1: By convention we might want to define a `full_width?` method on the component rather than using `@full_width`.

.rubocop.yml
```diff
@@ -21,6 +21,7 @@ Style/MixinUsage:
     - 'scripts/haml.rb'
 
 Metrics/MethodLength:
+  CountAsOne: ['array', 'hash']
```
davidrapson, line 24: If this stops Rubocop being such a nerd then ❤️

engine/app/components/citizens_advice_components/breadcrumbs.rb
```diff
@@ -4,10 +4,12 @@ module CitizensAdviceComponents
...
     attr_reader :type
 
-    def initialize(links:, type: :collapse)
+    def initialize(links:, type: :collapse, current_page: true, full_width: true)
       super
-      @type = fetch_or_fallback(allowed_values: %i[collapse no_collapse], given_value: type, fallback: :collapse)
       @links = links
+      @type = fetch_or_fallback(allowed_values: %i[collapse no_collapse], given_value: type, fallback: :collapse)
+      @current_page = fetch_or_fallback_boolean(current_page, fallback: true)
+      @full_width = fetch_or_fallback_boolean(full_width, fallback: true)
```
davidrapson, line 12: Makes sense to me to make this `true` by default 👍

engine/app/components/citizens_advice_components/breadcrumbs.html.haml
```diff
@@ -1,10 +1,11 @@
...
+  %nav{ class: "cads-breadcrumbs cads-breadcrumbs--#{type.to_s.dasherize}", "aria-label": "breadcrumbs" }
+    %ul.cads-breadcrumbs__list
+      - links.each_with_index do |link, index|
+        %li.cads-breadcrumbs__crumb
+          - if index == links.size - 1
+            %span.cads-breadcrumb{"aria-current": ("location" if @current_page) }
```
davidrapson, line 7: Same as above, maybe define a `aria_current?` method or similar.

davidrapson (approved):
👍 Couple of tiny suggestions, but nothing blocking as it's not about the public interface. Smashing.

PR #1247, opened 2021-06-16
Add status badges

davidrapson:
Because the main CI jobs and Browserstack are handled by different workflows I thought it might be useful to use Github's provided status badges to help us disambiguate a bit at a glance.

We can still see this through status checks and the actions runs but might as well have something up front in the README too.

image


davidsauntson (approved):
🦡

luke-hill (approved):

PR #1249, opened 2021-06-17
Remove pull request template

davidrapson:
Remove pull request template on the basis that most of the information in it is either outdated or explaining basics:

- References NP project board which is now a little too specific
- Review requests are automatically handled by codeowners now
- The "do not merge" label is rarely if ever used or we use Drafts instead.

Hopefully this isn't too controversial. Open to writing a new version if we'd rather keep something.


luke-hill (approved):
Prob should kill the old one from PW too also.

davidsauntson (approved):
Yeah fine with this - we will need to write smth for the contribution guidelines anyway, so when we've done that we can revisit whether we want a template again. 👍🏻

PR #1251, opened 2021-06-17
feat/table view component DS-18

davidsauntson:
Create the Table view component.

- view component haml / rb
- preview with examples
- rspec
- updates old docs with new example html and versions
- updates vrt urls and img refs

**Note on the testing**
All public website does at the moment is perform a [home-brew snapshot test](https://github.com/citizensadvice/public-website/blob/develop/spec/lib/renderers/table_data_renderer_spec.rb) which kinda feels right for something like testing table markup is correct.  I'm happy to replicate this here, so have added a new gem for snapshots ([`rspec-snapshot`](https://github.com/levinmr/rspec-snapshot)).

It is admittedly a bit unloved BUT works for this purpose and has v recently (-2days) changed ownership to someone who seems keen so hopefully that's ok!


engine/previews/citizens_advice_components/table_preview/long_table.html.haml
```diff
@@ -0,0 +1,6 @@
+.cads-page-content
```
davidrapson, line 1: Does this need all the layout classes? Guessing this is what the previous example did?

davidrapson (approved):
The component side looks fantastic, so definitely happy to get this in as is 👍 .

Couple of thoughts on the snapshot gem. The generated examples are kind of already a form of snapshots right? Does it make sense to rely on them and keep the tests themselves simpler?

> It is admittedly a bit unloved BUT works for this purpose and has v recently (-2days) changed ownership to someone who seems keen so hopefully that's ok!

Assuming the question above ☝️ is me oversimplifying then I say let's go with this. Pretty low risk and we can always pull it out if it stops working or is otherwise abandoned. Could we add something to the design system board with a deadline in 2-3 months time to check if it looks more active then?

davidsauntson:
> Couple of thoughts on the snapshot gem. The generated examples are kind of already a form of snapshots right? Does it make sense to rely on them and keep the tests themselves simpler?

I thought a bit about that and figured the example html wouldn't end up being permanent in the long term (probs we would replace with `render_inline` in the actual docs when move off storybook?).  This does feel like a long way away but figured it would be nice to have something that we won't have to change later.  I also figured we'd end up writing our own helper in the end anyway, since we wouldn't want to repeat the file reading and whitespace-ignoring logic all the time. 

Unless you mean not testing the html and just relying on the VRTs?

> Could we add something to the design system board with a deadline in 2-3 months time to check if it looks more active then?
👍🏻

engine/previews/citizens_advice_components/table_preview/long_table.html.haml
davidsauntson: There was no example outside of the sample pages so I just added some in to illustrate the scrolling behaviour.  Probably don't need _all_ of them...

davidrapson:
I think I was suggesting something even lazier, and relying on the diff in github for the examples. But you're right we probably won't need them forever. Your rationale makes total sense 👍

PR #1253, opened 2021-06-18
refactor: Further DRY out driver code

luke-hill:
Just delegates more stuff out to ca_testing

(Note I found a small bug in our migration of code to CaTesting, so there's a 2.0.1 fix going in as well (But for now this can still go in AS IS)


davidrapson (approved):
👍

PR #1254, opened 2021-06-18
bugfix: IOS browser caps and local BS runs

luke-hill:
- bugfix: Allow ios to use non aliased method name
- bugfix: Allow local browserstack runs


davidrapson (approved):
👍 good catch

PR #1255, opened 2021-06-18
Add ContactDetails view component

davidrapson:
- Adds new component
- Adds rspec tests
- Updates documentation
- Rewords example text to use a fake organisation instead of a real one
- Updates visual regression tests
- Updates callout stories to use generated examples and remove welsh language scenarios (moved to engine specs)


davidsauntson (approved):
nice! 📡

PR #1261, opened 2021-06-21
chore: update, audit

davidsauntson:
- update deps, audit fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8283,9 +8283,9 @@
           "dev": true
         },
         "postcss": {
-          "version": "7.0.35",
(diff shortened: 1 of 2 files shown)


davidrapson (approved):
👍 Although now wondering is as the CI time is much faster, and the lists seem a little smaller at the mo if we could get away with merging dependabots PRs directly again?

davidsauntson:
> 👍 Although now wondering is as the CI time is much faster, and the lists seem a little smaller at the mo if we could get away with merging dependabots PRs directly again?

Yeah but what will I do on Monday mornings?

PR #1262, opened 2021-06-21
Experimental: Support outside of Rails

davidrapson:
Adds the minimal changes required to support our experiment with using bridgetown for documentation.


davidsauntson (approved):
👏🏻

engine/lib/citizens_advice_components/engine.rb
```diff
@@ -9,5 +9,10 @@ class Engine < ::Rails::Engine
     isolate_namespace CitizensAdviceComponents
 
     config.view_component.preview_paths << CitizensAdviceComponents::Engine.root.join("previews")
+    config.autoload_paths = %W[
```
davidrapson, line 12: Before we merge this I want to make sure I understand the different between `autoload_paths` and `autoload_once_paths` and confirm this all works fine when linked to a Rails app.

davidsauntson: Cool - Zeitwerk will complain if it is asked to load an `autoload_once_path` that it has already loaded, but not sure of the ramifications beyond that.  Seemed to be an when running the demo app and my cads repo at the same time - but I don't know how Zeitwerk manages instances of itself.

davidrapson: Most I've been able to find is https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#rails-autoloaders which suggests `autoload_paths` is the way to go.

davidrapson:
Raised a PR to the public website to check that all tests pass with this change. If that looks good will merge this.

PR #1264, opened 2021-06-21
Mark nested callouts as future_release

davidrapson:
Adjusted the example heading level for nested callouts in https://github.com/citizensadvice/design-system/pull/1255 so this feature needs to be marked as future_release.

### DIFF
--- a/testing/features/components/callouts.feature
+++ b/testing/features/components/callouts.feature
@@ -28,6 +28,7 @@ Feature: Callout components
       And an "Adviser" label is present above the callout title
 
   Rule: Nested Callouts
+    @future_release @v4_1_2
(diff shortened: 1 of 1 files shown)


luke-hill (approved):

davidrapson:
Just tweaked this to tag the scenario instead. Turns out you can't tag a rule right now.

luke-hill:
Sorry, yes this is still the huge bugfix I did (Albeit only 1 line of code). It "will" come.

technically it was enabled in cucumber6, but it requires 3 other things to be merged before I can merge mine.

See here for my secondary fix: https://github.com/cucumber/cucumber-ruby/pull/1549

davidsauntson (approved):

PR #1267, opened 2021-06-21
WIP: Header component

davidrapson:
Starting to think about this for digital workplace, so making a stab at a component structure for the header. Specifically wanted to make it easy to swap out parts of the header so we can handle use cases like https://github.com/citizensadvice/design-system/pull/1071 without resorting to workarounds.

Makes heavy use of slots to keep things flexible. Allows things like:

- Opting out of the search form
- Providing a custom block for the account link, e.g. to replace the plain link with a form.
- Skip links and secondary links can be provided by slots too to keep the interface consistent

Definitely needs some thought around the API, currently quite a few slots with some variant of the word `links`. Needs some clearer names. Also need to consider how the header search form relates to the inline one; perhaps we can still share a component.

**TODO**:

- [x] Convert logo to slot. Use slots for everything for a consistent API
- [x] Improve names of slots for different kinds of links
- [x] API for account link with custom HTML
- [x] Settle on an approach for the search form
- [x] Test including navigation as part of header
- [x] Add rspec tets
- [ ] Generate examples from previews
- [ ] Document usage


engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
...
+      - if skip_links.present?
+        %nav.cads-skip-links
+          - skip_links.each do |link|
+            = link_to(link.title, link.url, class: "cads-skip-links__link")
+
+      .cads-grid-col-md-2.cads-header__logo-row
+        %a.cads-logo{ href: homepage_url, title: t("citizens_advice_components.shared.logo_title") }
+        %button.cads-header__search-reveal.js-cads-search-reveal.cads-icon_search{ toggle_button_attributes }
+      .cads-grid-col-md-10.cads-header__search-row
+        %ul.cads-header__links.js-cads-copy-into-nav
```
davidrapson, line 13: Wondering if it makes sense to bundle the global navigation in with this component. Does it make sense for them to be used separately?

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
+%header.cads-header
+  .cads-grid-container
+    .cads-grid-row
+      - if skip_links.present?
```
davidsauntson, line 4: naming: `skip-links` is good

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
...
+        %nav.cads-skip-links
+          - skip_links.each do |link|
+            = link_to link.title, link.url, class: "cads-skip-links__link"
+
+      .cads-grid-col-md-2.cads-header__logo-row
+        = logo
+        = toggle_button
```
davidsauntson, line 11: naming: might want to disambiguate from the nav toggle if you bring it in, so `search_form_toggle_button` (to go with `search_form` below)

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
...
+
+      .cads-grid-col-md-10.cads-header__search-row
+        %ul.cads-header__links.js-cads-copy-into-nav
+          - links.each do |link|
```
davidsauntson, line 15: naming: `header_links`?

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
...
+            %li.cads-header__links-item
+              - if link.current_site?
+                %span.cads-header__text= link.title
+              - else
+                = link_to link.title, link.url, class: "cads-header__hyperlink"
+
+          - if custom_account_link.present?
```
davidsauntson, line 22: naming: how about `sign_out`? (this will be the slot that custom sign out html goes into?)

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -0,0 +1,43 @@
...
+            %li.cads-header__account-link
+              = custom_account_link
+          - elsif account_link.present?
```
davidsauntson, line 25: naming: `sign_in_link`?

davidsauntson (commented):
Looking good so far.  Some thoughts on adding nav to header:
- would need to be optional as not all sites will want the nav
- ✅ there is interaction between the two components anyway and some nasty JS copying of links, so it would be nice to eliminate that and bring them together in the same component
- ❌ it will make the component quite big?

Probably the pro outweighs the con so maybe give it a bash and see how it feels by `npm link`ing to the public website and implementing it as a user would?

Added some naming suggestions alongside the code.

engine/app/components/citizens_advice_components/header.html.haml
davidrapson: Managed to rework this so there's a single slot for the account link now (the slot can take either arguments or a custom block). Keen to keep it general as it handles both cases. Think `account_link` feels right, or maybe `sign_in_out` to cover both cases?

davidrapson:
> Probably the pro outweighs the con so maybe give it a bash and see how it feels by npm linking to the public website and implementing it as a user would?

Ended up trying something where the navigation exists as a separate component, but we delegate to it as a slot within the header. Feels like a good mix of both worlds. You can still use the nav standalone; but there's a simpler API available if using with the header.

Will try it out within a project to get a feel for things.

engine/app/components/citizens_advice_components/navigation.html.haml
```diff
@@ -0,0 +1,6 @@
+.cads-navigation-full-width-wrapper
+  %nav#cads-navigation.cads-navigation.js-cads-greedy-nav
+    %ul.cads-navigation__list
+      - links.each do |link|
+        %li.cads-navigation__list-item
+          = link_to link[:title], link[:url], title: link[:title], class: "cads-navigation__link"
```
davidrapson, line 6: @davidsauntson Can you remember if there was a reason we include the `title` attribute for the navigation when it's the same as the link text?

davidsauntson: No it pre-dates both of us (https://github.com/citizensadvice/design-system/commit/5e903a52154094d79ee5e030f45e026ada022e9c#diff-398089f5ad0ebdc8766c211d098101d5a68b52d635a13bd2c975021524377a99) so happy for it to go

davidrapson:
> Will try it out within a project to get a feel for things.

The API for the navigation ended up feeling pretty good. However, realised it's a chunky change to the docs so kinda stretching the scope of these migrations. Going to keep them separate for now but we should revisit down the line.

davidrapson:
Going to raise a cleaned up version of this now the navigation component is in.

PR #1268, opened 2021-06-22
feat: use html for error summary

davidsauntson:
Not a lot to do for this since we aren't using this as a component with a defined interface at the moment.  I did consider creating a view component proper (eg with a list of errors and input ids) but rather than pre-empting the interface requirements I think it's best to wait until we look at forms as a whole.


davidrapson (approved):
💯 Approving pre-emptively because this is totally the right call.

PR #1269, opened 2021-06-22
Update asset hyperlink example

davidrapson:
We already had the component and the example file but hadn't yet updated the docs.

- Replace haml example with generated HTML version
- Add component documentation


davidsauntson (approved):
✨

PR #1270, opened 2021-06-23
feature: input view component DS-13

davidsauntson:
Thanks to @davidrapson  and @luke-hill  for comments on the first draft.  This is now ready for review.

This PR creates `CitizensAdviceComponents::Input` and its first born `CitizensAdviceComponents::TextInput` child.

- create `FormField` component responsible for rendering common form field structure markup
- create `Input` component to be used as a base for future input components that share commonalities (like a text input and textarea)
- create a `TextInput` to be used for text based inputs like email, search, text, password etc.  Can be of fixed widths.
- supporting rspec and vrts
- update the existing docs to use example html and include usage guidance for the view components
- move Welsh language test of the `optional` UI copy from features to rspec


davidrapson:
> making the FormField component usage transparent to the consumer (don't want them to have to explicitly use it especially as not using it means you can't see the error message)

Not sure if you managed to dig out my original experiment but I ended up doing pretty much the same thing https://github.com/citizensadvice/design-system/pull/594/files#diff-2ef4cdb27ad70e8c2ea7a8d5e692f658b6161bc411862b7c0a646acc248b8581 so that pattern makes sense to me. It should be there if you need it, but otherwise transparent to the consumer.

The other thing I tried in there which might still be useful is the `BaseInputComponent`. Which helped with the fact that `input` and `textarea` are basically the same.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -0,0 +1,88 @@
...
+
+module CitizensAdviceComponents
+  class Input < Base
+    attr_reader :name, :label, :error_message, :hint, :width, :value
+
+    def initialize(name:, label:, type:, options: nil)
+      super
+      @name = name
+      @label = label
+      @type = type
```
davidrapson, line 11: > making type required and not providing a default value of text (want people to think about what inputs they are creating more (eg email, search, number etc) and not just make everything a text field
> not validating type (although there are only c. 20 so I could still do this)

I think in this case what you have here would work. A named argument should throw an error if one is not provided. We could probably write a custom error message though if you want to guide people more.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      set_options(options)
```
davidrapson, line 13: Curious why this is a separate hash? Feels like these could all still be named arguments? Possibly missing your intent though.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -0,0 +1,88 @@
...
+        allowed_values: allowed_width_values,
+        given_value: options[:width],
+        fallback: nil
+      )
+      @value = options[:value]
+      @additional_attributes = options[:additional_attributes]
+    end
+
+    def allowed_width_values
+      %w[2ch 4ch 8ch 16ch] << nil
```
davidrapson, line 32: This syntax is fun, adding `nil` to the list? Wonder if this could be `[nil, :2ch, :4ch, :8ch, :16ch]`? I think using symbols here makes sense, consistent with the other components.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    def optional?
+      @optional
+    end
+
+    def required?
+      !@optional
```
davidrapson, line 40: Maybe use the previous method here? `!optional?`

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -0,0 +1,88 @@
...
+        value: value,
+        required: required?,
+        "aria-invalid": error?,
+        class: ("cads-input--#{width}" if width?),
+        "aria-describedby": (error_id.to_s if error?)
+      }
+    end
+
+    def input_attributes
+      if @additional_attributes.present?
```
davidrapson, line 81: Nice!

davidrapson (commented):
⭐ Thanks so much for picking this one up. Inputs are a thorny one. A few little comments but looking great so far.

engine/app/components/citizens_advice_components/input.rb
davidrapson: 100% agree with the point about making people think about what type they use.

engine/app/components/citizens_advice_components/form_field.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class FormField < Base
+    def initialize(error: false)
+      super
+      @error = fetch_or_fallback_boolean(error, fallback: false)
+    end
+
```
davidrapson, line 9: Maybe add: 

```
def render?
  content.present?
end
```

engine/spec/components/input_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    it "renders the input at the correct width" do
+      expect(subject.css(".cads-input--4ch")).to be_present
+    end
+  end
+
+  context "when an invalid width is specified" do
+    let(:width) { "banana" }
+
+    context "production rails env" do
```
luke-hill, line 127: from a testing perspective there probably should be a non-production unit test also.

You can also get hints for gaps like this by using something like `simplecov` I'm not sure if we use it in public-website. It's a code-coverage gem that's reasonably simple to setup.

If you change the below line to         `allow(Rails.env).to receive(:production?).and_return(production?)` then you can then specify the flag switch in the same way you do other items.

davidrapson: This is really testing a single helper https://github.com/citizensadvice/design-system/blob/master/engine/app/lib/citizens_advice_components/fetch_or_fallback_helper.rb so it would be nice to cover this in a what that doesn't feel like we need the same boilerplate every time. Any suggestions for using shared examples or similar approaches here?

engine/app/components/citizens_advice_components/input.rb
davidsauntson: Lots of (>4) params is a bit of a code smell and also enforced by rubocop.  Option object params is one of the refactoring approaches (https://refactoring.guru/introduce-parameter-object), felt like it made sense here since they are also all optional.

engine/app/components/citizens_advice_components/input.rb
davidsauntson: Isn't it tho 😄   Agree it could be nicer.  Alas if you start a symbol with a number you have to enclose the symbol text in quotes, eg `:"2ch"` which feels a bit awkward to me and probably not immediately obvious to users?  Maybe it should just be a simple array `[nil, "2ch", "4ch", ...]`?

davidsauntson: OR could rename, eg `two-ch` and then just update the CSS.  Probably this I think.

engine/app/components/citizens_advice_components/input.rb
davidsauntson: Yeah I think I'll enumerate the options here and then have a custom error message referencing mdn or smth 👍🏻

engine/app/components/citizens_advice_components/input.rb
davidrapson: Ooh interesting. I've always interpreted Introduce Parameter Object more for when you've got like parameters being passed together, e.g. passing `start` and `end` when you really want a `range`. I'd argue in this case we've only moved the arguments around, the component still has the same number of arguments we've just put them in a hash. But I don't feel strongly here, and the distinction between required and optional parameters seems reasonable here. Totally your call 👍

engine/app/components/citizens_advice_components/input.rb
davidrapson: Ooh good point 👍  Yeah we should rename, the "2ch" format makes sense as implementation details but it doesn't need to be how we ask you to name them. Maybe go super explicit and say `:two_characters`? We can always map that to the terser class names behind the scenes, or rename.

engine/spec/components/input_spec.rb
davidsauntson: Could we test the fallback helpers directly and then not test them in the components?  Feels like the coverage would be there?  

Would be keen to know what you think though @luke-hill - basically what is the best way to test that invalid params are handled correctly? (as defined by these helpers here: https://github.com/citizensadvice/design-system/blob/master/engine/app/lib/citizens_advice_components/fetch_or_fallback_helper.rb and implemented by the components here: https://github.com/citizensadvice/design-system/blob/8665e1715ae977b8b81e64cabfa2aaa4ca276e05/engine/app/components/citizens_advice_components/callout.rb#L9)

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -32,6 +36,29 @@ def allowed_width_values
       %w[2ch 4ch 8ch 16ch] << nil
     end
 
+    def allowed_type_values
+      %i[
```
davidrapson, line 40: I wonder if should limit this to text based values? Not sure about this component handling stuff like colour / file types etc. that feel like they should be their own input types in future.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -66,7 +93,7 @@ def error_id
 
     def base_input_attributes
       {
-        type: @type,
+        type: @type.to_s.dasherize,
```
davidrapson, line 99: Ah Rails

engine/spec/components/input_spec.rb
luke-hill: Yep, it's quite common to write helper tests. If this is just consuming that.

The theory behind simplecov (If you are interested), is just each LOC is tested once.

If you want to test the helper a good way (My recommendation is).

```
let(:dummy) do
  Class.new do
    include MyHelper
  end
end

subject(:named_subject) { dummy.new }


context "when good" do
  it "passes" do
    expect(dummy).to be_awesome
  end
end

context "when crap" do
  it "fails" do
    expect(dummy).to raise_error(ArgumentError).with_message("Be better!")
  end
end
```

luke-hill: EDIT: When testing them, the folder structure should mimic the lib. So the tests would probably live in `spec/helpers/myhelper_spec.rb`

davidrapson: Perfect. In fact it looks like Primer—where we pinched `fetch_or_fallback` from—does a similar thing albeit in minitest.

davidrapson: Raised this https://github.com/citizensadvice/design-system/pull/1273. Thanks for the nudge in the right direction!

testing/features/forms/input.feature
```diff
@@ -35,16 +35,6 @@ Feature: Input forms
...
-      And there is a hint
-      And there is an optional label
-      But there is no error message
-
-    Scenario: Welsh Input can have an optional label
-      Given the language is Welsh
-      When I type "Hello" into the text box
-      Then the text box has a value of "Hello"
-      And there is a label
-      And there is a hint
```
luke-hill, line 47: given they are of the same speed, can we leave all properties it shows. So for this one can we leave `And there is a hint` *Unless we're changing the physical property*

testing/features/forms/input.feature
```diff
@@ -56,15 +46,4 @@ Feature: Input forms
       When I type "Hello" into the text box
       Then the text box has a value of "Hello"
       And there is a label
-      And there is a hint
```
luke-hill, line 59: Same here with hint / optional label (unless caveat wrong)

luke-hill (commented):
Seems legit. Few questions

engine/app/components/citizens_advice_components/input.html.haml
```diff
@@ -0,0 +1,14 @@
+= render CitizensAdviceComponents::FormField.new(error: error?) do
+  %label.cads-form-field__label{id: label_id, for: input_id }
+    = label
+    - if optional?
+      %span.cads-form-field__optional= t(".optional")
+  - if hint?
+    %p.cads-form-field__hint= hint
+  - if error?
+    %p.cads-form-field__error-message{ id: error_id, data: { testid: "error-message" } }
```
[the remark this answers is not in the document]
davidsauntson: I think we settled on `data-testid` across the board, so this should be ok.

engine/app/components/citizens_advice_components/text_input.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def initialize(width: nil, **args)
+      @width = fetch_or_fallback(
+        allowed_values: allowed_width_values,
+        given_value: width,
+        fallback: nil
+      )
+      args[:options]&.merge(width: @width)
+      @base_input_args = args
+      super(@base_input_args)
```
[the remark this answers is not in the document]
davidsauntson: The `base_input_args` are passed into the haml template for the `Input.new()` call in the haml file here: https://github.com/citizensadvice/design-system/pull/1270/files#diff-7a46c55e3a358c2b0eaeec8d9da67fd14bd7a836c1c6b536312d1600c538e958R1

testing/features/forms/input.feature
davidsauntson: The diff is a bit mad - basically I:
 - removed the Welsh test as it's in the rspec now
 - remove the hint and the optional label from the error example as it's not rendered with them anymore (covered by the texts above tho still)

luke-hill: So we've actually changed the properties of a couple of these forms then?

engine/config/locales/cy.yml
```diff
@@ -1,20 +1,23 @@
+---
```
davidrapson, line 1: What does `---` do?

engine/previews/citizens_advice_components/contact_details_preview.rb
```diff
@@ -1,7 +0,0 @@
-# frozen_string_literal: true
```
davidrapson, line 1: Did you mean to delete this?

scss/6-components/_forms.scss
```diff
@@ -126,27 +126,31 @@
 //
 // Modifier: ~2 characters wide
 //
+.cads-input--two-chars,
```
davidrapson, line 129: If we wanted to avoid the new class names we could have a mapping in the component from symbol to class name:

```
{
  :two_chars: "2ch",
  :four_chars: "4ch"
}[width]
```

testing/features/forms/input.feature
davidsauntson: Yeah - the optional text is only added on the optional example (it was in error example too before), the hint text is only added in the hint example and fixed width example (it was in error example too before).

davidrapson (approved):
👍 Looking great. Couple of little questions, but nothing particularly blocking. Although you may have deleted the contact details preview by mistake?

scss/6-components/_forms.scss
davidrapson: To clarify: the new class names are clearer so happy to migrate over to them if we want 👍

engine/config/locales/cy.yml
davidrapson: Hold on, I can see this is a result of running `normalize keys`

luke-hill (approved):
Testing-wise this is good. Top stuff. No idea on JS though 😆

engine/config/locales/cy.yml
luke-hill: Per the yml spec most files "should" start with this sorta thing. It's also auto-generated by a lot of CLI stuff. It's kinda annoying though

davidrapson:
> No idea on JS though

I was about to ask "What JS??" and then realised it's the storybook docs 😆. It all looks right to me!

engine/previews/citizens_advice_components/contact_details_preview.rb
davidsauntson: Not so much

PR #1271, opened 2021-06-23
Test cleanup

davidrapson:
A bit of cleanup post Github Actions migration. Mostly removing old drivers, methods, and config files that no longer apply.

- **Remove some old grid environment variables** — We no longer use the grid diver for actions
- **Single config for pa11yci** — Github Actions can use the same localhost URL so we only need a single config
- **Remove old backstop config** — Holdover from Jenkins / docker-composer setup
- **Remove test helpers** — Cleans up a few places where we had some unused test helper code


testing/.env.example
```diff
@@ -30,12 +30,6 @@ BROWSERSTACK_DEBUG_MODE=false
 # i.e. iPad7th_13 --> iPad 7th generation, Running Browser Version(iOS), 13
 BROWSERSTACK_CONFIGURATION_OPTIONS=Windows_10_83
 
-# Whether you want to run on the Selenium Grid or not?
-GRID=false
```
luke-hill, line 34: I'm a bit against removing these from the example file, as these env params will still work and can be used locally by someone spinning up a grid.

We will also definitely be doing some more work on this once 4.0.0 is out as I'm hearing then the edge image for docker will be a bit better (Possibly).

luke-hill (commented):
Actually you can possibly disregard my point if you want us to entirely stop using grid for DS.

I'm not sure how I feel, not necessarily against it. Not necessarily for it. Sorta fenced. 🤷

testing/.env.example
davidrapson: Oh interesting, I hadn't considered someone using grid locally. We don't have any need for it in github actions land anymore. So perhaps the best thing to do is instead to update the examples to they are more generic; not referring to docker-compose endpoints that no longer exist.

luke-hill: Tell you what. To avoid holding this up. Lets delete these for now. And "if" we want to use them again in DS, we'll just re-add them in.

davidrapson: Hah! I went the other way 😄.

davidsauntson (approved):

PR #1273, opened 2021-06-24
Add spec for fetch_or_fallback helper

davidrapson:
Ports over the library test code and helper method from Primer. Converts from minitest syntax to rspec equivalents.

- Adds a single spec for fetch_or_fallback to cover the common scenarios.
- Introduces a `without_fetch_or_fallback_raises` helper, copied from Primer, which allows us to test component specific fallback behaviours.

Thanks to @luke-hill for the nudge in the right direction here.


engine/spec/components/base_spec.rb
```diff
@@ -1,50 +0,0 @@
-# frozen_string_literal: true
-
-RSpec.describe CitizensAdviceComponents::Base, type: :component do
```
davidrapson, line 3: We were already using the base component spec to do a version of this. But testing the module directly in the new spec file is much clearer.

engine/spec/rails_helper.rb
```diff
@@ -12,8 +12,23 @@
...
 require "view_component/test_helpers"
 
+module TestHelpers
+  # For components using fetch_or_fallback,
+  # we want to ensure that the correct
+  # fallback behavior works in production.
+  # The helper enables us to easily test
+  # this production-only behavior.
+  def without_fetch_or_fallback_raises
+    CitizensAdviceComponents::FetchOrFallbackHelper.fallback_raises = false
```
luke-hill, line 22: maybe not one for this PR, but if you want to you could do.

```
fallback_raises = CitizensAdviceComponents::FetchOrFallbackHelper.fallback_raises
# logic
ensure
  CitizensAdviceComponents::FetchOrFallbackHelper.fallback_raises = fallback_raises
```

luke-hill (approved):

davidsauntson (approved):

PR #1295, opened 2021-06-28
Migrate i18n-tasks to engine

davidrapson:
By moving the i18n-tasks setup into the engine we can use the built-in rspec test and rely on a simpler config. The previous setup was more focussed on the old haml files.

One hurdle was that i18n-tasks and view component disagree slightly on how to resolve translations that are used within component classes. There's only a handful of these so the simplest thing to do is to write them out long-form in these cases.

Taking a bit of a punt here that we're OK removing the top-level config now. It got a bit confusing having both at once, but happy to try and keep that if this feels preemptive.


.github/workflows/ci-workflow.yml
```diff
@@ -43,7 +43,6 @@ jobs:
           bundler-cache: true
       - run: bundle exec rubocop --format github
       - run: bundle exec haml-lint
```
davidrapson, line 45: There's a similar job to do at some point for `haml-lint`. Added a card to the design system board for this.

davidsauntson (approved):
👍🏻 I actually prefer the long form locale keys.

PR #1296, opened 2021-06-28
Search component

davidrapson:
Adds component version of the search field.

The main change here is to treat this like the other form inputs. Rather than including the whole form element I've limited this to the field itself.

**TODO:**

- [x] Add component and template
- [x] Rspec tests
- [x] Previews
- [x] Update documentation
- [x] Remove welsh language from functional tests

### DIFF
--- a/styleguide/components/search/search-docs.mdx
+++ b/styleguide/components/search/search-docs.mdx
@@ -12,6 +12,25 @@ import { Story, Preview } from '@storybook/addon-docs/blocks';
   <Story id="components-search--search-with-value" />
 </Preview>
 
+## Using with Rails
+
+If you are using the `citizens_advice_components` gem you call the component from within a template using:
+
+```rb
+= render(CitizensAdviceComponents::Search.new(value: "Current search term"))
+```
+
+Unlike the legacy haml partial the component will only render the input and button, not the surrounding from to give you more control over how the field is used.
+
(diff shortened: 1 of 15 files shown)


styleguide/components/search/search-docs.mdx
```diff
@@ -12,6 +12,25 @@ import { Story, Preview } from '@storybook/addon-docs/blocks';
...
 
+## Using with Rails
+
+If you are using the `citizens_advice_components` gem you call the component from within a template using:
+
+```rb
+= render(CitizensAdviceComponents::Search.new(value: "Current search term"))
+```
+
+Unlike the legacy haml partial the component will only render the input and button, not the surrounding from to give you more control over how the field is used.
```
davidsauntson, line 23: typo: `from` &rarr; `form`

bike shedding but might be better to flip the sentence to emphasis that you'll have to render this in your own `form` as it's a fairly major change from the haml partial.

davidsauntson (approved):
🔍 nice

styleguide/components/search/search-docs.mdx
davidrapson: Good shout. The other approach we could take is to allow a `content` block in the form. But based on how you're using this component in public website limiting it to the field seems most flexible.

PR #1297, opened 2021-06-28
Page review component

davidrapson:
In a bit of a flow for migrating these, so doing one more whilst I have the time. Migrates the page review component.

**TODO:**

- [x] Add rspec scenarios
- [x] Add component interface
- [x] Confirm approach to passing date (formatted or as date object)
- [x] Add preview
- [x] Document component API
- [x] Update functional tests to remove welsh language scenarios

### DIFF
--- a/engine/app/components/citizens_advice_components/page_review.rb
+++ b/engine/app/components/citizens_advice_components/page_review.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class PageReview < Base
+    def initialize(page_review_date:, date_format: nil)
+      super
+      @date_format = date_format || "%d %B %Y"
+      @page_review_date = page_review_date
+    end
+
+    def call
+      content_tag(
+        :p,
+        t("citizens_advice_components.page_review.body_html", date: formatted_date),
+        class: "cads-page-review"
+      )
+    end
+
+    def formatted_date
+      I18n.l(@page_review_date, format: @date_format)
+    end
+
+    def render?
+      @page_review_date.present?
+    end
+  end
+end
--- a/engine/citizens_advice_components.gemspec
+++ b/engine/citizens_advice_components.gemspec
@@ -25,5 +25,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency     "haml-rails", [">= 2.0.0", "< 3.0"]
   spec.add_runtime_dependency     "rails", [">= 6.0.0", "< 7.0"]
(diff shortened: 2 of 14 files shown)


engine/citizens_advice_components.gemspec
```diff
@@ -25,5 +25,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency     "haml-rails", [">= 2.0.0", "< 3.0"]
   spec.add_runtime_dependency     "rails", [">= 6.0.0", "< 7.0"]
+  spec.add_runtime_dependency     "rails-i18n", [">= 6.0.0", "< 7.0"]
```
davidrapson, line 28: This is a requirement of translating the date within the component using `I18n.l`

engine/app/components/citizens_advice_components/page_review.rb
```diff
@@ -0,0 +1,27 @@
...
+    def call
+      content_tag(
+        :p,
+        t("citizens_advice_components.page_review.body_html", date: formatted_date),
+        class: "cads-page-review"
+      )
+    end
+
+    def formatted_date
+      I18n.l(@page_review_date, format: @date_format)
```
davidrapson, line 20: 💭 Thoughts wanted on this API. My preference is to have the component accept a date object, so we can define the default format. We previously required people to pass a pre-formatted string. Any strong feelings either way?

styleguide/components/page-review/page-review.stories.mdx
```diff
@@ -1,21 +1,33 @@
...
 
+## Using with Rails
+
+If you are using the `citizens_advice_components` gem, you can call the component from within a template using:
+
+```rb
+= render(CitizensAdviceComponents::PageReview.new(page_review_date: Date.new(2020, 9, 21)))
+```
+
+Unlike the legacy haml partial the component accepts a date object rather than a formatted string. This allows us to define a default format. This can be customised by passing a `date_format` option.
```
davidsauntson, line 22: Would be nice to link to the `I18n` docs list of formats here (if there is one)?

engine/app/components/citizens_advice_components/page_review.rb
davidsauntson: Yeah I think this is fine - we could parse a UTC `Z` string or something but the component is always going to be in ruby land anyway so seems a bit overkill and we may as well use a date object 👍🏻

davidsauntson (approved):
🍸 🍣 

An ideal date.

PR #1298, opened 2021-06-28
Feat/textarea DS-32

davidsauntson:
Adds the `Textarea` component, the newest and most favourite child of `Input` component.
- create `Textarea` component, spec and previews
- update `Input` to allow `nil` as a type (`type` is ~~not a valid~~ a read-only attribute on a `textarea`)
- update docs
- update VRTs

Separate PR: rationalise preview method names across input components so they align with these ones


engine/spec/components/textarea_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    let(:rows) { 10 }
+
+    it "renders the textarea with the correct number of rows" do
+      expect(subject.css("textarea").attribute("rows").value).to eq("10")
+    end
+  end
+
+  context "when an invalid number of rows is specified" do
```
davidrapson, line 55: You might not need to use shared examples here. Perhaps something like:

```rb
context "when an invalid number of rows is specified" do
  let(:rows) { "banana" }

  it "uses fallback rows" do
    without_fetch_or_fallback_raises do
      expect(subject.css("textarea").attribute("rows").value).to eq("8")
    end
  end
end
```

Which uses the `without_fetch_or_fallback_raises`.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -10,7 +10,7 @@ def initialize(name:, label:, type:, options: nil)
       @label = label
       @type = fetch_or_fallback(
         allowed_values: allowed_type_values,
-        given_value: type.to_sym,
+        given_value: type.present? ? type.to_sym : nil,
```
davidrapson, line 13: Might be able to get away with `given_value.presense` here. Should return the symbol or nil in this case.

davidrapson (approved):
📝 

Couple of 🤓 comments but nothing blocking

engine/spec/components/textarea_spec.rb
davidsauntson: How did I forget about this! 🙈

PR #1301, opened 2021-06-29
Navigation view component [DS-14]

davidrapson:
## Background

https://citizensadvice.atlassian.net/browse/DS-14

## Changes

Adds view component version of navigation component.

Started this in https://github.com/citizensadvice/design-system/pull/1267 but decided to leave the navigation out of the header component for now. I think the interface makes sense longer term, but would require too much documentation rework to be worthwhile now. Still need the navigation though so: here it is.

**TODO:**

- [x] Add rspec tests
- [x] Add component definition and template
- [x] Add component previews
- [x] Update documentation
- [x] Remove old haml examples


davidsauntson (approved):
🧭 👍🏻

PR #1303, opened 2021-06-29
Remove language switcher from Storybook

davidrapson:
**Background**

We're pretty much at the point now where all our welsh language specs are in rspec component tests. I think we should strip the remaining locale code out now to make the last stages of migration simpler.

**Changes**

- Remove any remaining welsh language scenarios from functional tests
- Remove reliance on `I18n` in functional tests
- Remove `switch_language` behaviour
- Remove language switcher from Storybook

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -26,31 +26,3 @@ export const parameters = {
     },
   },
 };
-
-export const globalTypes = {
-  locale: {
-    name: 'Locale',
-    description: 'I18n locale',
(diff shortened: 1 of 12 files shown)


davidsauntson (approved):
braf 👍🏻 

There aren't any `@not_local` tests right? And so we we aren't going to gain anything by forcing a run of the browserstack tests on this branch.

davidrapson:
Yeah, I think we're good there.

PR #1304, opened 2021-06-29
Feat/radio buttons DS-33

davidsauntson:
Adds the 'RadioGroup' view component.

- creates view component
  - uses a `renders_many` slot and `RadioButton` view component to render the inputs and labels proper
- add previews, rspec
- update docs


engine/app/components/citizens_advice_components/radio_group.html.haml
```diff
@@ -0,0 +1,16 @@
...
+      = legend
+      - if optional?
+        %span.cads-form-field__optional= t("citizens_advice_components.input.optional")
+    - if hint?
+      %p.cads-form-field__hint= hint
+    - if error?
+      %p.cads-form-field__error-message= error_message
+    - radio_buttons.each do |radio|
+      .cads-radio-group__item
+        %input.cads-radio-group__input{ radio.attributes(name) }
```
davidsauntson, line 14: Wanted a user to only have to pass `name` once per group - wonder if there is a nicer way of sharing the `name` between classes?

engine/app/components/citizens_advice_components/radio_group.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    def hint?
+      @hint.present?
+    end
+
+    def fieldset_attributes
+      {
+        class: [
+          ("cads-radio-group--#{@size.to_s.dasherize}" if size?),
```
davidsauntson, line 54: I know that passing `:regular` and `:list` aren't really doing anything but felt this was better than bools as it gives us more room in the future.

engine/app/components/citizens_advice_components/radio_group.rb
```diff
@@ -0,0 +1,93 @@
...
+        ]
+      }
+    end
+
+    class RadioButton < Base
+      attr_reader :label
+
+      def initialize(label:, value:, checked: false, **additional_attributes)
+        super
+        @label = label
```
davidsauntson, line 65: This is where I had accidentally added commas.  So `label` was an array of all the arguments, but all the arguments also had their own values.  It was a fun time.

davidrapson (approved):
💯 Looks spot on to me. Like the use of slots for radio options. Feels right as it's a group.

engine/app/components/citizens_advice_components/radio_group.rb
davidrapson: Yeah, a list of states is always better than booleans 👍

PR #1306, opened 2021-06-30
Header view component [DS-12]

davidrapson:
## Background

https://citizensadvice.atlassian.net/browse/DS-12

Clean version of https://github.com/citizensadvice/design-system/pull/1267 now that all other components have been migrated.

## Changes

Everything is handled [using slots](https://viewcomponent.org/guide/slots.html). There's things like the logo that could be handled as arguments to the component, but using slots for everything feels consistent.

For example:

```rb
render CitizensAdviceComponents::Header.new do |c|
  c.logo(title: "Citizens Advice homepage", url: "/")
  c.skip_links(skip_links)
  c.header_links(header_links)
  c.search_form(search_action_url: "/search")
  c.account_link(title: "Sign in", url: "/sign-in")
end
```

### Logo slot

This is the only slot required by default. The slot takes a `title` and optional url (defaults to "/"). Made the title required because each product should really have a different title. Currently the partial renders "Citizens Advice homepage" even for adviser products where a more specialised name makes sense.

```rb
c.logo(title: "Citizens Advice homepage", url: "/")
```

### Skip links

Skip links are optional. We provide a default set but also let you provide your own.

```rb
c.skip_links([
  { title: "Skip to content", url: "#main" },
  { title: "Skip to webchat", url: "#chat" }
])
```

### Header links

Similar API to the old Haml partial, but as a slot.

```rb
c.header_links([
  { title: "Public site", url: "/", current_site: true },
  { title: "AdviserNet", url: "?advisernet" },
  { title: "CABlink", url: "?CABlink" },
  { title: "BMIS", url: "?BMIS" },
  { title: "Cymraeg", url: "?lang=cy" }
])
```

### Search form slot

Made the search form optional. In order to use it you need to tell us your `search_action_url`, and optionally a `search_param` which defaults to `:q`.

```rb
c.search_form(search_action_url: "/search", search_param: :search)
```

### Account link slot

After a good chat with @davidsauntson we've decided to accept a completely custom block to override the account link e.g. to replace it with a custom button. This has the effect that we also require consumers to handle styling too. Will require some updates to the advice sites to account for this.

The slot can be used one of two ways. As a plain link:

```rb
c.account_link(title: "Sign in", url: "/sign-in")
```

Or by passing a custom block to render your own HTML

```rb
c.account_link do
  = form_tag("/sign-out") do
     = button_tag("Sign out", class: "cads-button cads-button__tertiary")
```

To simplify things a bit I've made a couple of tweaks to the styles, particularly around how we use the `.cads-header__acount-link` class so help with this change. Made the equivalent markup changes to the old haml partial.

## How to review

Really appreciate 👀 on this. Play around with it. Try and break the API. Perhaps worth pairing on a trail run updating public advice to use this to test out the component API?

## The reward 🥳 

When we merge this is the last view component we need to migrate!! Last step is to extract out the sample pages and we can 🗑️ all the haml code from storybook.

## TODO

- [x] Rspec tests added 
- [x] View component created 
- [x] Translations migrated
- [x] Support rendering custom HTML for account link
- [x] Component previews created, matching the stories in the storybook docs
- [x] Documentation added


demo/app/frontend/packs/application.scss
```diff
@@ -1,5 +1 @@
 @import '~@citizensadvice/design-system/scss/lib.scss';
-
-#content {
```
davidrapson, line 3: I removed this as for header and footer component previews it helps for them to be flush. Will revisit this in a future PR as it would be nice to specific either for previews.

engine/previews/citizens_advice_components/header_preview/full_example.html.haml
```diff
@@ -0,0 +1,9 @@
+= render CitizensAdviceComponents::Header.new do |c|
+  :ruby
+    c.logo(title: "Citizens Advice homepage", url: "/")
+    c.skip_links(skip_links)
+    c.header_links(header_links)
+    c.search_form(search_action_url: "/search")
+  - c.account_link do
+    = form_tag("/sign-out/out", authenticity_token: false) do
```
davidrapson, line 8: This `authenticity_token: false` is important for previews, so we always generate stable HTML. But it's a good thing as it proves that we can use the regular form helpers like `form_tag` and `button_to` and rails will inject `authenticity_token` if needed.

davidrapson:
Thanks backstop! Let me fix this border wonkiness before this gets reviewed.

image

engine/spec/components/header_spec.rb
```diff
@@ -0,0 +1,158 @@
+# frozen_string_literal: true
+
+RSpec.describe CitizensAdviceComponents::Header, type: :component do
```
davidrapson, line 3: @luke-hill Mind reviewing my rspec scenarios?

engine/app/components/citizens_advice_components/header.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class Header < Base
+    renders_many :skip_links, "SkipLink"
+    renders_many :header_links, "HeaderLink"
+
+    renders_one :account_link, "AccountLink"
+
+    renders_one :logo, lambda { |title:, url: "/"|
```
luke-hill, line 10: Might get caught up on rubocop / linting but usually stabby lambdas use the `{}` and long ones use `do end`

i.e. 
```
renders_one :logo, -> (title, url) { puts title + url }
```

```
renders_one :logo, lambda do |title, url|
  puts title + url
end
```

engine/spec/components/header_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+      render_inline(described_class.new)
+    end
+
+    it "does not render" do
+      expect(component.at("header")).not_to be_present
+    end
+  end
+
+  context "with logo" do
+    subject(:component) do
```
luke-hill, line 15: Using subject will change the value of the `subject` invocation. It also gives you access to a shorthand one liner.

i.e. you can say `it { is_expected.to be_present }` When you have a defined subject.

Probably best to alter this one to be `let`

engine/spec/components/header_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    subject(:logo) { component.at(".cads-logo") }
+
+    it "renders a logo" do
+      expect(logo).to be_present
+      expect(logo.attr("title")).to eq "Logo title"
+      expect(logo.attr("href")).to eq "/homepage"
+    end
+  end
+
+  context "with skip links" do
```
luke-hill, line 30: with provided skip links

luke-hill (commented):
Seems good, couple of extra generic questions (All non-blocking, so feel free to merge without).

- Is it common using this library to wrap all sub-classes into one file. It feels a little odd seeing that.
- When writing rspec, having multiple contexts feels weird. Traditionally the first block start would be `describe` with the method instance/class that you're testing. Here maybe you could use describe to state what part of the render your testing (This would also avoid you needing to prefix everything with with/when)

engine/app/components/citizens_advice_components/header.rb
davidrapson: Yeah I had this the second way but rubocop complained!

luke-hill: Interesting. It shouldn't have done, but I may have missed some config we might have setup in the style repo.... I'll have a quick look at this.

davidrapson:
> Is it common using this library to wrap all sub-classes into one file. It feels a little odd seeing that.

Yeah it's pretty common for this library. It's modelled a bit on React where it's fairly common to have internal components within the same file. See https://viewcomponent.org/guide/slots.html#rendering-collections for an example. It's also possible to pull things out into separate classes, but that's mostly done when it makes sense for the sub-component to be used standalone (we did this with a little `Badge` component).

> When writing rspec, having multiple contexts feels weird. Traditionally the first block start would be describe with the method instance/class that you're testing. Here maybe you could use describe to state what part of the render your testing (This would also avoid you needing to prefix everything with with/when)

Oooh interesting. I guess I'd not considered these as "methods" as they are all parts of the same component interface; but they really are just method calls! Will give this a go and see how it feels.

engine/app/components/citizens_advice_components/header.rb
davidrapson: In fact trying the second way throws a `tried to create Proc object without a block` error.

luke-hill: I "think" this might be a library / ruby 2.7 interop bug. But I'm not 100% sure. This "should" work / be interchangeable, I'll look into this for my own learnings but feel free to disregard for now if you want.

demo/app/frontend/packs/application.scss
davidsauntson: No probs - FYI I added it for the field component as it needs to not be flush in the error state.

davidsauntson (approved):
This is great 👏🏻

davidrapson:
Just tested this with `npm link` using the visual regression tests in public website and all looks good.

PR #1307, opened 2021-07-02
Test engine against multiple Ruby + Rails versions [DS-37] [DS-38]

davidrapson:
## Background

Different product teams use different versions of ruby and may upgrade to newer versions of Rails at different cadences. Let's make sure we support more than just Ruby 2.7.3 + Rails 6.1!

### Engine changes

- Updates engine and demo gemfiles to support running against a custom version of rails or `main`
- Adds a matrix build to run against Ruby 2.7 and 3.0 + Rails 6.1 and `main`
- Fixes some Ruby 3 compatibility issues

This gives us a 4x matrix build for each ruby + rails combination. Nice!

### Workflow changes

One bit of complexity here is that branch protection required checks rely on the job name. The matrix build generates four different names that would all need to be added to the required checks list. This is a bit unweildy!

Thankfully we can happily move the `demo-app` build to be dependent on the engine matrix and limit the required status checks only to this dependent build. This is conceptually more in line with the separation between the engine and the storybook build. It was a bit misleading to have the `demo-app` dependent on storybook before.

Takes us from this:

image

To this:

image

## Before merging

- [ ] Remove old `engine` job from required status checks. Rely on check for dependent `demo-app` job.

### DIFF
--- a/demo/Gemfile.lock
+++ b/demo/Gemfile.lock
@@ -213,8 +213,5 @@ DEPENDENCIES
   web-console (>= 3.3.0)
   webpacker (~> 5.0)
 
-RUBY VERSION
-   ruby 2.7.3p183
-
 BUNDLED WITH
    2.1.4
--- a/engine/app/components/citizens_advice_components/text_input.html.haml
+++ b/engine/app/components/citizens_advice_components/text_input.html.haml
@@ -1,2 +1,2 @@
-= render CitizensAdviceComponents::Input.new(base_input_args) do
+= render CitizensAdviceComponents::Input.new(**base_input_args) do
   %input.cads-input{ input_attributes }
--- a/engine/spec/components/footer_spec.rb
+++ b/engine/spec/components/footer_spec.rb
@@ -3,9 +3,7 @@
 RSpec.describe CitizensAdviceComponents::Footer, type: :component do
   subject(:component) do
     render_inline(CitizensAdviceComponents::Footer.new(feedback_url: feedback_url)) do |c|
-      columns.each do |column|
-        c.column(column)
-      end
+      c.columns(columns)
     end
   end
 
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -44,30 +44,6 @@ jobs:
       - run: bundle exec rubocop --format github
       - run: bundle exec haml-lint
 
-  engine:
-    runs-on: ubuntu-20.04
-    defaults:
-      run:
-        working-directory: engine
-    strategy:
-      fail-fast: false
-      matrix:
-        rails_version: ['6.1', 'main']
-        ruby_version: ['2.7', '3.0']
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: ${{ matrix.ruby_version }}
-          working-directory: engine
-      - run: |
-          bundle config path vendor/bundle
-          bundle update
-          bundle install --jobs 4 --retry 3
-          bundle exec rake spec
-        env:
-          RAILS_VERSION: ${{ matrix.rails_version }}
-
   # Actions can run in parallel at the job or the workflow level.
   # In order to run our build-dependent tests in parallel we need a way
   # to build the docs and then "seed" each stage with a copy.
@@ -85,7 +61,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-      - run: npm run build --if-present
+      - run: npm run build
       - run: npm run docs:build
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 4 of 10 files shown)


engine/app/components/citizens_advice_components/text_input.html.haml
```diff
@@ -1,2 +1,2 @@
-= render CitizensAdviceComponents::Input.new(base_input_args) do
+= render CitizensAdviceComponents::Input.new(**base_input_args) do
```
davidrapson, line 1: @luke-hill This feels like something you might know? I _think_ I understand why I need the splats here, but trying to read up on the differences in keyword arguments in Ruby 3 made my head hurt. Any background here?

engine/spec/components/footer_spec.rb
```diff
@@ -3,9 +3,7 @@
 RSpec.describe CitizensAdviceComponents::Footer, type: :component do
   subject(:component) do
     render_inline(CitizensAdviceComponents::Footer.new(feedback_url: feedback_url)) do |c|
-      columns.each do |column|
-        c.column(column)
-      end
+      c.columns(columns)
```
davidrapson, line 6: We can also do:

```
columns.each do |column|
   c.column(title: column[:title, links: column: column[:links])
end
```

But using the plural form of the method for passing an array is more conventional for view component slots.

engine/app/components/citizens_advice_components/text_input.html.haml
luke-hill: So basically the issue is ambiguity. The biggest issue is that in ruby certain things are inferred or automatic (i.e. every method has a block argument, but often it is nil).

The problem is if you have a method with positional / named / hash arguments you need a way of differentiating them. Often people used to cheat and pass a hash of the named arguments. However this would work in **both instances**

i.e.
```
def foo(a: 1, b: 2)
  "a is #{a}, b is #{b}"
end
```

Here we have 2 keyword arguments so there's no real issue. Furthermore if you pass in a hash, it'll try find the keywords a and b, again no real issue.

However if you have a bit of a ambiguous one you can get issues. A simple (passing one), is the following

```
def bar(a = 1, **opts)
  "a is #{a}, opts is #{opts}"
end
```

Here we want a positional argument first, then a hash of all remaining options, with this method signature we have no issues until we pass in 2 things that look like a hash i.e.

```
bar("x" => 3, y: 5) # This gives different answers on ruby 2.6 and ruby 2.7
```

Because we're not sure whether this is one hash (And one argument), and should go as `a` or it's two hashes and 2 arguments and should overwrite a=1 and leave y:5 as the opts.

The full descriptive article (From the horses mouth), is this one. https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

However in short (TL;DR)

1) Nothing changes if you **only** have properly defined methods and you switch to using **kwargs on invocations.
2) If you expect a hash of options passed in. Be more specific now, don't be lazy
3) If you have a method that has positional, named and then a hash of options at the end, things may be different from ruby 2.6->2.7 (2.7 to 3 basically goes from throwing warnings to crashing).

A shorter explanation of the final point 3 is (From 2.6->2.7)

See https://github.com/site-prism/site_prism/pull/49 & https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

> When a method call passes a Hash at the last argument, and when it passes no keywords, and when the called method accepts keywords, a warning is emitted. To continue treating the hash as keywords, add a double splat operator to avoid the warning and ensure correct behavior in Ruby 3.

luke-hill (dismissed):
Whilst it's a bit more complex / bigger. If you look how we tackled this on cucumber-rails using Appraisals then it'll probably save you some hassle / issues. Specifically using specific gemfiles for each version you want to tackle.

https://github.com/cucumber/cucumber-rails/blob/main/.github/workflows/build.yml

davidrapson:
> Whilst it's a bit more complex / bigger. If you look how we tackled this on cucumber-rails using Appraisals then it'll probably save you some hassle / issues. Specifically using specific gemfiles for each version you want to tackle.

Thanks, that's great to know. This is following the example https://github.com/primer/view_components which feels like pretty good reference point for our use case. So happy keeping it as is for now. Good to know there's something like appraisals to turn to if our needs get more complex though.

demo/Gemfile.lock
```diff
@@ -213,8 +213,5 @@ DEPENDENCIES
   web-console (>= 3.3.0)
   webpacker (~> 5.0)
 
-RUBY VERSION
```
luke-hill, line 216: This will cause problems as this will flag a diff every time. If you use appraisals it's actually easier than this way. In fact you don't even need to use appraisals if you wanted manually to upload the gemfiles (Just appraisals regenerates them for you)

davidrapson:
Lets come back to this.

davidsauntson:
Linking this to new github issue https://github.com/citizensadvice/design-system/issues/1394

PR #1323, opened 2021-07-05
Sample pages [DS-39]

davidrapson:
## Background

This depends on https://github.com/citizensadvice/design-system/pull/1306 before we can merge it. It's also the last change we need to make before we can completely remove haml from Storybook.

https://citizensadvice.atlassian.net/browse/DS-39

## Changes

I've made a few choices here that are worth explaining:

### Use erb for page templates in demo app

I wanted to scratch an itch and try using plain erb for the templates within the demo app. I think it's useful to have a fairly "stock" demo app to test that we're not leaking too much haml related stuff from the engine.

It had the interesting side-effect of being _really_ pleasant to work with. It turns out editor support for plain old HTML is _way_ better than Haml 😅

### Render sample pages using high_voltage

Uses https://github.com/thoughtbot/high_voltage to render the sample pages. Gives us a nice interface for accessing them and allows us to drop templates into `views/pages` without having to write a custom controller. Updates the rake task to support these.

### Consolidates sample page templates

The sample pages we had were a bit lean. It didn't seem hugely worth replicating them exactly as is so I've consolidated:

- **Content page** — A content heavy template with an example of most of the components in it.
- **Form page** — Replicates the form page pretty much as is.


davidrapson:
Note to self: Squash this on merging to clean up the history.

davidsauntson (approved):
🦖 This is really great!  High Voltage looks cool.

You can also delete `styleguide/haml_locals.rb` too now, which is a file I particularly hate.  I did this locally and everything is ok - it was just for the sample pages afaik.

davidrapson:
> You can also delete styleguide/haml_locals.rb too now, which is a file I particularly hate.

Oh don't worry, PR incoming to delete more than just that!

PR #1324, opened 2021-07-06
Footer: Use :title instead of :label in links

davidrapson:
Spotted whilst demoing the new components. We use `:title` for the other components that accept lists of links, like navigation and breadcrumbs. Update the new footer component to be consistent.


davidsauntson (approved):
👍 nice!

PR #1325, opened 2021-07-06
Docs cleanup

davidrapson:
To try and separate the concept of contributing in the broad sense and developer docs we are removing the CONTRIBUTING.md file from the public docs and instead breaking down tech docs into smaller guides. This leaves space for a future, broader contributing guide within storybook.


davidrapson:
Merged as discussed with @davidsauntson last week. We can do a post-merge review on this later.

PR #1348, opened 2021-07-20
fix: prevent search jump

davidsauntson:
Fixes a bug where focussing on a header link causes the search box to jump up and down.

Added a VRT as something similar happened ages ago.

**Before**
before

**After**
after


davidrapson (approved):
👟

PR #1349, opened 2021-07-21
Remove all ruby dependencies from storybook

davidrapson:
Now that all examples are generated from the demo Rails app we no longer need to wrangle storybook to support haml templates or any ruby code.

- Removes need for custom webpack loaders
- Removes top-level haml and rails gems
- Removes unused helpers like haml_locals and story-helpers


davidsauntson (approved):
Lush 💣

PR #1350, opened 2021-07-21
Rationalise lint tasks [DS-26]

davidrapson:
## Background

Our previous rubocop and haml-lint setup was designed based on the previous storybook setup with these tools being run at a high level with only a few rules. Once https://github.com/citizensadvice/design-system/pull/1349 is merged we can improve our lint set up a bit.

## Changes

- Rather than having a single top-level lint config which covers a mix of different things, lets specialise and have one for the engine code and one for the testing code. Allows them to tailor to fit their respective needs.
- Removes the `ruby-lint` job in favour of running the linters directly in their respective jobs (e.g. `engine` now runs rubocop, haml-lint, and rspec all together)
- Fixes a few component style issues flagged by the new specialised configs 🥳 
- Removes some old rake tasks that are a holdover from the previous setup / jenkins-land.

We'll need to remove `ruby-lint` from the list of required jobs before we merge this.


davidrapson:
@luke-hill Have removed the exclude line to show what's failing https://github.com/citizensadvice/design-system/runs/3123354747?check_suite_focus=true#step:6. Seems reasonable as we're installing to `vendor/bundle` and `haml-rails` includes some sample haml files. Makes sense to exclude that.

luke-hill:
That's fine, but those files shouldn't be checked at all. So I'm not really sure what's going on. https://github.com/citizensadvice/citizens-advice-style-ruby/blob/main/default.yml#L11 is where we say ignore all vendored files. Maybe there's a bug there?

davidrapson:
Could it be because this is in a child directory? so `engine/vendor/bundle`?

luke-hill:
Possibly yeh. but then why would the overridden one work? I don't know.

If you're happy with it I can approve. I'm doing a bit of a deep-dive into Casebook currently

luke-hill (approved):

davidrapson:
Yep cheers. I think it makes sense as this is `haml-lint` specifically failing rather than `rubocop` and a little exclude fixes it so all good to me.

davidsauntson (approved):
Smashing 🧹

engine/.rubocop.yml
```diff
@@ -0,0 +1,28 @@
+inherit_gem:
+  citizens-advice-style:
+    - default.yml
+    - default-rails.yml
+
+AllCops:
+  SuggestExtensions: false
+  Exclude:
```
luke-hill, line 8: if you remove these lines does it fail? Quick way to test would be just running locally so you don't need to push up

davidsauntson:
Removed the `ruby-lint` task from the list of required tasks for ya 😄

PR #1351, opened 2021-07-21
Experimental: Run backstop against demo app

davidrapson:
This is a proof of concept to see if we can run backstop against the demo app. Using the targeted content component as an example, specifically so we can test using `page.setJavaScriptEnabled(false)` for no-JS scenarios.


davidrapson:
@davidsauntson This is a bit of a punt, but working well locally at least. Mind checking this out and running `npx backstop test --docker --config=backstop-config-demo-app.js`. Not expecting to merge this, just want to rule out if it's possible and if we can disable JS in the visual regression tests.

davidrapson:
Closing this because I've managed to goof something and deleted all of the original images 😬 . Does seem like it's reasonably simple to switch when we want to though.

PR #1352, opened 2021-07-21
Fix/basic account link form styles

davidsauntson:
Fixes a big with the sign out button in the greedy nav dropdown.

I know that we agreed consumers will be responsible for styling the account link HTML if they provide it - but seems foolish to consign the existing styles for the simple sign out button example to the dustbin.

Instead I have created a subclass of the `AppHeaderComponent` called `FormAppHeaderComponent` that is functionally the same, just renders the button in the sign out form in the `account_link` slot.  Since the styles already exist I think it's nice to surface in the examples and docs (I've used it on the forms sample page).


davidrapson (approved):
👍 This makes total sense to me. A slightly less nuclear option than handing it all over to consumers.

styleguide/examples/sample_pages/form-sample.html
```diff
@@ -43,8 +43,29 @@
...
+              <form
+                class="cads-header__account-form"
+                action="/sign-out"
+                accept-charset="UTF-8"
+                method="post"
+              >
+                <input
+                  type="hidden"
+                  name="authenticity_token"
+                  value="3SoLRI9T1WxgJ9CG7CyrMF7Gy3M5w7z570z8IE6X9wSY9ajYt2RusmGm2zlE8n-L3sqnZ6Y1uNqibhiSpyH6YA"
```
davidrapson, line 56: Ooh, just spotted you'll need to set `authenticity: false` on the form to strip this value.

davidrapson (changes requested):

davidrapson (approved):

PR #1353, opened 2021-07-21
Refactor/minor world tidy

luke-hill:
This does 3 things
- Remove some Worldable methods / Modules (These weren't used regularly and can be moved out of the global NS or deleted full stop
- Removed up some code that was not relevant any more (api tests / parameter types)
- Upgraded to cucumber v7, and with it, taggable rules now works, so I've removed some code in the feature files that is no longer needed


davidrapson (approved):
Lovely 💯. Always a fan of less code. Rubocop is mean and calling something "useless" but merge away once you've appeased the robot.

celonicolosi (approved):

testing/features/components/navigation.feature
```diff
@@ -13,10 +13,10 @@ Feature: Navigation component
     Scenario: All navigation links are shown
       Then the dropdown toggle is not present
 
+  @small_screen
   Rule: Default Navigation - small screens
```
davidrapson, line 17: Oh nice, can we tag rules in Cucumber v7 then?

luke-hill: Yep, It's all fixed now. Was good to get it here in the end.

Also v7 has a completely rewritten message buffer, which is now written using pure custom json, as opposed to using googles protobuf

PR #1355, opened 2021-07-21
Add rubocop-rspec

davidrapson:
Adds our own little automated @luke-hill by configuring `rubocop-rspec` 🤖 . Fixes a few of the simpler common issues for `RSpec/FilePath` conventions and `RSpec/ContextWording` and then uses `--auto-gen-config` to create a `.rubocop_todo.yml` file for stuff that needs a little more thought.

@luke-hill Have you used this before? Any tips on settings we should tweak, or is all of this decent stuff?


davidsauntson (approved):
Oh this is nice - deffo going to learn a lot from this 🤖

luke-hill (approved):
As an introduction this is fine, all setup as I'd expect. Again this is something we should probably introduce to the styling gem proper as/when I bother to look into it again.

Think you're seeing a couple of the ones I mentioned before like having describe's mention the core functionality / method name

davidrapson:
@luke-hill Cheers. Yeah I ended up carrying on yesterday and fixing the rest of the warnings so going to push the rest of that up in a moment. Each time I hit something and thought "hmm, is this really a problem?" I tried the suggested way and it made the tests a lot clearer. 💯 would lint again.

luke-hill:
Yeh. What I would say. Being totally honest. Is whilst I agree with 95% of the standard rubocop configuration, I would say for rubocop rspec that number is more like 70% or so.

See places like this: https://github.com/site-prism/site_prism/blob/main/.rubocop.yml#L30-L31 or https://github.com/site-prism/site_prism/blob/main/.rubocop.yml#L19-L23

engine/spec/components/radio_group_spec.rb
```diff
@@ -1,168 +0,0 @@
...
-    context "when invalid layout parameter is passed" do
-      let(:layout) { :bananas }
-
-      it "renders the radio buttons in list layout" do
-        without_fetch_or_fallback_raises do
-          expect(subject.css(".cads-from-group--inline")).to_not be_present
-        end
-      end
-    end
-  end
```
davidrapson, line 154: @davidsauntson FYI I spotted that these two tests were broken. `cads-from-group--inline` is a typo and we ask if it's **not** present so could never fail. Fixed that in the corrected version.

engine/spec/components/citizens_advice_components/header_spec.rb
```diff
@@ -6,39 +6,41 @@
       render_inline(described_class.new)
     end
 
-    it "does not render without any slots" do
+    specify "does not render without any slots" do
```
davidsauntson, line 9: If these are aliases, shall we just use `it`?   It (lol) certainly reads better here

davidsauntson (approved):
Fine with this.  Need educate myself on some of the changes - no more shared examples?

davidrapson:
> no more shared examples

Not specifically, just in our cases the warnings encouraged separating / specialising the tests to the point where they just weren't needed any more. Shared examples are legit still.

PR #1356, opened 2021-07-22
Update getting started guide

davidrapson:
Updates the getting started guide to reflect that `citizens_advice_components` will soon be the recommended way of using design system components within Rails apps.

Restructures the guide to make it clear you can use the styles without using Rails.


davidsauntson (approved):
📜

PR #1357, opened 2021-07-22
docs: initial roadmap

davidsauntson:
Adds a small roadmap for future versions of the design system into storybewk.

### DIFF
--- a/styleguide/documentation/roadmap.stories.mdx
+++ b/styleguide/documentation/roadmap.stories.mdx
@@ -1,0 +1,27 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Roadmap" />
+
+# Future version roadmap
+
+Here is what will be happening with the design system over the next few versions.
+
+## v4.2.0+
+
+- all existing components will also be available as [view components](https://viewcomponent.org/)
+- no additional functionality will be added to haml partials
+- only modification to haml partials will be bugfixes and any WCAG 2.X A or AA enhancements
+
+We are currently investigating ways of surfacing deprecation warnings for these haml partials.
+
+## v5.0.0+
+
+- haml partials will be removed from the design system
+- consumers of the design system must use the view components or the HTML and CSS examples provided
+- ongoing work to improve SASS documentation
+
+## Documentation
+
+Currently the desgin system is documented using storybook. Work to replace this with a more suitable site constructed using the design system itself is ongoing.
+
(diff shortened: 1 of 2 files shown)


styleguide/documentation/roadmap.stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+Here is what will be happening with the design system over the next few versions.
+
+## v4.2.0+
+
+- all existing components will also be available as [view components](https://viewcomponent.org/)
+- no additional functionality will be added to haml partials
+- only modification to haml partials will be bugfixes and any WCAG 2.X A or AA enhancements
+
+We are currently investigating ways of surfacing deprecation warnings for these haml partials.
```
davidrapson, line 15: If we wanted to we can do this by adding the following to partials:

```
:ruby
   Rails.logger.warn("Upgrade me!")
```

But we'd a) have to add that to **every** partial as there's no central code b) would rely on people upgrading to v4.2 **and** noticing it in their logs. Personally wouldn't worry too much as we can chat directly to the team(s) that need to upgrade right now. But will leave it to your judgement.

This kind of thing is definitely easier going forward as we have ruby classes to play with and so we can use standard patterns for deprecation.

davidrapson (approved):
👍 Lovely! With a little suggestiong about deprecation warnings.

styleguide/documentation/roadmap.stories.mdx
luke-hill: Usually in most Gem frameworks you would have deprecation policies / notices so you don't need to check logs. The 3 quickest ways you could do this are....

1) Create a new deprecation class/model, whereby you surface this behaviour in multiple ways dependent on level of deprecation. See https://github.com/site-prism/site_prism/blob/main/lib/site_prism/deprecator.rb OR https://github.com/SeleniumHQ/selenium/blob/64447d4b03f6986337d1ca8d8b6476653570bcc1/rb/lib/selenium/webdriver/common/logger.rb#L115-L130

2) use just a portion of the above. The simplest would be either `p "foo"` or `Kernel.warn("foo")`

3) Alter the rails logger implementation here to push notices to the console if the level is at warn or higher.

davidrapson: Perhaps I should clarify: This was specifically looking at deprecations patterns for the old loose haml partials, which as a major limitation don't have ant companion classes or models _at all_. So the only place we could do anything is through code in the partials themselves. Hence the hacky suggestion.

luke-hill: I think your suggestion is probably the right way to do it then. I don't think it's really a hack, as it's basically the way I do it in site_prism (Just with a package-global method). So here you could just do `Kernel.warn` or `p` if you wanted to surface it out to the users console.

PR #1358, opened 2021-07-23
Remove broken tests scripts

davidrapson:
These have been broken for a while and as far as I can tell are not used in any current workflows.

Realised I've never run these, always ran pa11y and backstop with the low-level commands. @davidsauntson Shout if that's mistaken and you'd rather fix these up.


davidsauntson (approved):
Nice - yeah I also just run the actual commands rather than the npm scripts so happy to 🔥

PR #1389, opened 2021-08-03
Fix/input value

davidsauntson:
~~Allows consumers to specify values for `Textarea` and `TextInput` components.  Revolutionary!~~

`TextInput` now renders the value passed into the component.  Updated unit tests and docs to show versions with values.

Probs didn't need the examples in the docs but I did them out of habit :-)

### DIFF
--- a/engine/app/components/citizens_advice_components/input.rb
+++ b/engine/app/components/citizens_advice_components/input.rb
@@ -4,10 +4,11 @@ module CitizensAdviceComponents
   class Input < Base
     attr_reader :name, :label, :error_message, :hint, :value
 
-    def initialize(name:, label:, type:, options: nil)
+    def initialize(name:, label:, type:, value: nil, options: nil)
       super
       @name = name
       @label = label
+      @value = value
       @type = fetch_or_fallback(
         allowed_values: allowed_type_values,
         given_value: type.presence,
@@ -83,7 +84,7 @@ module CitizensAdviceComponents
         type: (@type.to_s.dasherize if @type.present?),
         id: input_id,
         name: name,
-        value: value,
+        value: @value,
         required: required?,
         "aria-invalid": error?,
         "aria-describedby": (error_id.to_s if error?)
--- a/engine/app/components/citizens_advice_components/input.rb
+++ b/engine/app/components/citizens_advice_components/input.rb
@@ -4,7 +4,7 @@ module CitizensAdviceComponents
   class Input < Base
     attr_reader :name, :label, :error_message, :hint, :value
 
-    def initialize(name:, label:, type:, value: nil, options: nil)
+    def initialize(name:, label:, type:, options: nil)
       super
       @name = name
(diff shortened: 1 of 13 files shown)


davidrapson (approved):
👍 A truly wild idea!

davidrapson:
(Also how lovely is it to be able to add unit test coverage for this kind of thing now)

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -83,7 +84,7 @@ def base_input_attributes
         type: (@type.to_s.dasherize if @type.present?),
         id: input_id,
         name: name,
-        value: value,
+        value: @value,
```
luke-hill, line 87: This already seemed to come from the above (Line27 then Line5). This also seems to be getting overwritten potentially.

davidrapson: Oh yeah! Wait were we already supporting the `value` through `options`?

davidsauntson: 😬  Let me check

davidsauntson: Well yeah, I've lost the plot 😵  I didn't believe in past-David.
The unit tests still stand so I'll just update this PR to add those.

davidsauntson: Great spot - thanks @luke-hill  😄

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -4,7 +4,7 @@ module CitizensAdviceComponents
   class Input < Base
     attr_reader :name, :label, :error_message, :hint, :value
 
-    def initialize(name:, label:, type:, options: nil)
+    def initialize(name:, label:, type:, value: nil, options: nil)
```
luke-hill, line 7: Given this lVar is now no longer used you can drop this too. But as it's symbolised it won't flag if you don't (Linting might show it up though).

engine/app/components/citizens_advice_components/textarea.html.haml
```diff
@@ -1,2 +1,3 @@
 = render CitizensAdviceComponents::Input.new(base_input_args) do
   %textarea.cads-textarea{ input_attributes }
+    = value
```
luke-hill, line 3: Do you want to validate that the value is non-blank? Because in the `set_options` method it could be fetched as a nil value, but then I guess that's actually not a problem for an auto-entered value right 🤷

luke-hill (approved):
Groovy stuff. Couple of questions but they can probably be ignored.

engine/app/components/citizens_advice_components/textarea.html.haml
davidsauntson: Yeah it's fine if it's nil, it just renders nothing which is correct

PR #1390, opened 2021-08-04
Add github-actions to dependabot rules

davidrapson:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot

Interested to see what it picks up.


luke-hill:
If we want to automate this more we could try using automerge? I know they have it set up on some of the casebook stuff.

PR #1407, opened 2021-08-05
Issue templates

davidsauntson:
A couple of suggested issue templates so I can demo them in the next design system working group.

Not worried too much about the text right now - just want to show the feature to the group.


davidrapson (approved):
👍

PR #1415, opened 2021-08-10
Breadcrumbs render check

davidrapson:
## Summary of problem

Breadcrumbs render even if there's only one item. We should probably only render if there is more than one.

## Steps to replicate

Provide a single item to the `Breadcrumbs` component, e.g. just a Home link

```rb
render CitizensAdviceComponents::Breadcrumbs.new(links: [{ title: "Home", url: "/" }])
```

## Expected behaviour

We should probably only render the breadcrumb trail if there is more than one item.

## Actual behaviour

It renders a single item. Because it's also the "last" item it renders without a link and on small screens is hidden apart from a border.

## Context

**Single item:**

image

**Missing on mobile:**

image

(A bit hard to tell from the screenshot but all that's left is the border)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 **Bugfixes**
 - `TextInput` and `Textarea` components: allow consumers to specify a value (!)
+- 🥖 Breadcrumbs: Don't render when there is only one item.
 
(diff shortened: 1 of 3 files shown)


davidsauntson (approved):
🔗 🎉

PR #1433, opened 2021-08-18
Update CI browsers

luke-hill:
Chrome v89 was a good version to automate. For firefox the a similar one was v88. For the win10 vers just use stable ones that are reasonably new


davidrapson (approved):
👍 Makes sense.

Probably worth temporarily setting `on: [page_build]` in the workflow to `on: [push]` to check this works before merging.

luke-hill:
I've tested these ones locally, so if they fail now, then it's something else.

PR #1434, opened 2021-08-20
Experimental: Docs tools

davidrapson:
A little experiment to add basic sassdoc and yard support.


davidsauntson (commented):
Nice 🍶 

I think there is something to having some of this documentation in the static site we will create, rather than these independent doc sites.  Particularly for SASS colours and the documentation of `.new` for view components.  Understanding how to use `.new` is pretty vital and would maybe feel a bit lost if we only surface it in the yard generated html (I find those doc sites pretty dense).

davidrapson:
Yeah totally, GDS do some stuff with sassdoc where the use the generated JSON to make some custom templates. Primer do a similar thing with yard so we can pull out bits to use in a more readable way I reckon.

engine/app/components/citizens_advice_components/targeted_content.rb
```diff
@@ -4,6 +4,20 @@ module CitizensAdviceComponents
...
+    #     title: "Example targeted content",
+    #     type: :adviser
+    #   ) do %>
+    #     Example content
+    #   <% end %>
+    #
+    # @param title [String] Title
+    # @param id [String] A HTML-friendly ID
+    # @param type [Symbol] Type, either :public or :adviser with a fallback of :public
+    # @param heading_level [Integer] Optional, sets the heading level of the title. Defaults to `2`
```
davidsauntson, line 19: If we specify these defaults in the signature, yard will pick the up automatically.

PR #1446, opened 2021-08-24
Dependency audit and Storybook prune

davidrapson:
This does a bit of dependency audit:

- Runs `npm audit fix`
- Streamlines the storybook build — Removes add-ons we're not using. Controversially including the `a11y` add-on on the basis we have `pa11y` running already and it should keep the storybook build streamlined to have minimal add-ons.
- Removes `js-beautify` and `systemjs` which are installed but unused.


.storybook/main.js
```diff
@@ -1,14 +1,4 @@
 module.exports = {
   stories: ['../styleguide/**/*.stories.@(js|mdx)'],
-  addons: [
-    {
-      name: '@storybook/addon-essentials',
```
davidsauntson, line 5: I have seen others use the viewport tool this provides, might be worth leaving it for now?

davidrapson: Ah good shout. I'm mostly trying to avoid the churn of having the whole essentials module. Looks like we can pull in https://storybook.js.org/addons/@storybook/addon-viewport/

Issue #1396, opened 2021-08-05
SVG icon system


davidsauntson:
Switch to using plain SVGs for icons rather than a font https://css-tricks.com/icon-fonts-vs-svg/ Aside from the resilience and performance benefits might help with this too.

Add icons as individual SVGs inside bounding box

Updating icons in existing components

PR #1449, opened 2021-08-27
Feat/svg icons #1396

davidsauntson:
👋🏻 Early draft of svg icon components - would be good to get some feedback before going too much further 😄 

- makes a base icon class that handles parameters (currently only size)
- specify small and large sizes as per the design system guidelines
- allow no size so people can override in their CSS
- supply default colour in the CSS

If we use the preview to generate examples of the two sizes, people can just grab the the actual SVG from there, rather than providing them as files too?  Then they'll have access without having to use a view component.

Would also like to provide a component for plus/minus and up/down arrows in the same file, like you did with targeted content, along with some predefined `--is-open` modifiers for some basic open/close functionality.

### DIFF
--- a/engine/app/components/citizens_advice_components/icons/base.rb
+++ b/engine/app/components/citizens_advice_components/icons/base.rb
@@ -10,6 +10,7 @@ module CitizensAdviceComponents
 
       def size
         return if @size.blank?
+
         size
       end
     end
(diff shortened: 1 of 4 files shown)


davidsauntson:
Closing PR until I pick this pack up again

PR #1462, opened 2021-08-31
Dependabot audit fix and config updates

davidrapson:
Upgrade demo app dependencies to fix a warning and adds demo app to dependabot at monthly cadence. Also uses this an excuse to simplify the dependabot config:

- Remove labels (use default `dependencies` label)
- Remove open-pull-requests-limit config, the default is fine
- Remove timezone, we don't mind when pull requests are opened
- Remove target-branch, dependabot uses the default in github


davidsauntson (approved):
Cool - I quite like the specific dependency labels but not massively attached.

PR #1482, opened 2021-09-10
Update browsers

luke-hill:
Done a reasonable update of browsers


davidrapson (approved):
👍 

Looks reasonable. Remind me again, what's the issue with using latest -1 for evergreen browsers?

luke-hill:
The issue is under the hood it does 3 transformations. So this is a bit more optimal, but primarily it gives us control.

The -1 / -2 / +1 e.t.c. relies on something they set.

PR #1492, opened 2021-09-14
Experimental: Run pa11y against demo app directly

davidrapson:
Proof of concept to test running pa11y against the demo server directly rather than running against storybook. Updates the demo-app job to start a test server and then run pa11y against that in a single job.

An alternative option might be to use https://github.com/dequelabs/axe-core-gems and run the accessibility check as part of the selenium tests. Maybe that's a preferable approach?


testing/wcag/pa11yci.js
```diff
@@ -1,13 +1,83 @@
...
     },
     ignore: [
-      'WCAG2AA.Principle2.Guideline2_4.2_4_1.G1,G123,G124.NoSuchId'       // ignore missing anchor links in design-sytem
-    ]
+      // ignore missing anchor links in design-sytem
+      'WCAG2AA.Principle2.Guideline2_4.2_4_1.G1,G123,G124.NoSuchId',
+    ],
   },
-  urls: getComponentUrls('http://localhost:6006'),
+  urls: [
```
davidrapson, line 15: We could probably generate this list from the component previews if we wanted. Maybe the generate examples rake task. I don't mind this being a manual list though.

davidrapson:
Thinking about this, I reckon I'd prefer to revisit this by moving the selenium tests over to the demo app first and trialing https://github.com/dequelabs/axe-core-gems then.

PR #1493, opened 2021-09-16
Update ca_test gem

celonicolosi:
This gets the repo using v3 of the ca_testing gem.
EDIT: This will need v3.0.2 of the gem releasing first.


luke-hill:
You can't do this work until "after" releasing the changes to ca_testing. But all seems fine.

Gemfile.lock
```diff
@@ -167,6 +169,7 @@ GEM
     zeitwerk (2.4.2)
 
 PLATFORMS
+  x86_64-darwin-19
```
luke-hill, line 172: This probably shouldn't be here. Not sure though

celonicolosi: This went automatically after the bundle install or bundle update (don't remember each one right now). But I will remove it.

luke-hill (changes requested):
You haven't made any of the necessary changes from the breaking updates. Those need updating as well

testing/features/support/drivers/browserstack.rb
```diff
@@ -14,7 +14,15 @@ def register
...
-      browser_type_class.new.register
+      # This is missing from the selenium repo proper. It has been added to all relevant spots
+      # HERE: https://github.com/SeleniumHQ/selenium/pull/9862
+      # Remove once we are only supporting Selenium 4
+      #
+      # LH - Sep '21
+
+      require "selenium/webdriver/common/platform"
+
+      CaTesting::Drivers::Browserstack.new(browser, configurable_capabilities).register
```
luke-hill, line 25: This will need a new method here to push in the device specific caps but only when they are provided

testing/features/support/drivers/browserstack.rb
```diff
@@ -33,22 +41,16 @@ def valid_browserstack_drivers
...
-      %i[
-        android
-        chrome
-        internet_explorer
-        ios
-      ].include?(browser)
+    def configurable_capabilities
+      {
+        build_name: PayloadValuesGenerator.build_name,
+        project_name: "Public-Website tests",
```
davidrapson, line 47: Can we rename this so it doesn't reference public website?

celonicolosi: Sure, thanks for noticing David!

luke-hill (approved):

davidrapson (approved):

PR #1507, opened 2021-09-21
Configure reek

davidrapson:
Bit of an experiment in configuring reek https://github.com/troessner/reek

Generated a `.reek.yml` file using `reek --todo` and then tried making a few of the suggested changes to see if it catches much. All the recommended changes made a lot of sense and caught a couple of quirks in the code so I think this looks reasonable.

@luke-hill I imagine you might have some experience with reek. Any tips for if we do go ahead and use this?


luke-hill:
So in essence three things to bare in mind.

1) It's hardly used now.
2) Rubocop does all the things reek does
3) Rubocop has a community guide where people can suggest items that can be added to future style iterations.

Is there anything in particular you get from reek that you don't get from rubocop. i.e. is there an offence which has been setup from one that you don't get on the other (To my knowledge there aren't really any).

davidrapson:
Closing this after a chat, looks like many of these suggestions are either misleading or even false positives for our use case so we can probably do without it.

PR #1510, opened 2021-09-22
Add guards for empty table rows

davidrapson:
Found myself adding a bit of guard code in some table renderer code that feels like it could stand to live in the design system code.

If empty rows are passed to the table we will render them. It makes sense to guard against this and to add a basic `render?` check to the table so we don't render broken tables.

- Add render? check for missing header and rows
- Flatten empty row data

---

Also used this as an excuse to update our spec code to use the latest recommended pattern from the view component docs Based on https://viewcomponent.org/guide/testing.html#rspec-configuration

Uses the provided helper methods along with capybara matchers which lets us write things like this:

```
expect(rendered_component).to have_css "caption", text: "Example caption"
```

Rather than needing to manually wrangle nokogiri objects.


engine/spec/components/citizens_advice_components/table_spec.rb
```diff
@@ -1,55 +1,79 @@
...
+        rows: sample_rows,
+        caption: "Example caption"
+      )
+    end
 
-  it "renders correctly" do
-    expect(component.to_html).to match_snapshot("table")
+    it "renders a caption" do
+      render_inline component
+      expect(rendered_component).to have_css "caption", text: "Example caption"
```
luke-hill, line 63: Not a problem, but just worth pointing out because capybara uses xpath under the hood, you can just use any of the capybara query matchers isolated. (So long as you're happy to). So if you wanted you could just then composite this to

`expect(rendered_component).to have_text "Example caption"` as well as increasing the amount so for instance

`expect(rendered_component).to have_css "caption", text: "Example caption", count: 1`

davidsauntson: <3

engine/spec/components/citizens_advice_components/table_spec.rb
```diff
@@ -1,55 +1,79 @@
...
-      ]
-    ]
+  context "when missing headers" do
+    subject(:component) do
+      described_class.new(header: [], rows: sample_rows)
+    end
+
+    it "does not render" do
+      render_inline component
+      expect(rendered_component).to have_no_table
```
davidsauntson, line 24: Is `rendered_component` an automagic shorthand thing?

davidsauntson (approved):
Oh nice!  Great that we can use standard capybara selectors too 👍🏻

engine/spec/components/citizens_advice_components/table_spec.rb
davidrapson: Yep, although not especially magic. It's a reader for the output of `render_inline` provided by `ViewComponent::TestHelpers`. https://viewcomponent.org/guide/testing.html#rspec-configuration

PR #1524, opened 2021-09-29
Feat/svg icons 1396

davidsauntson:
Adds icon svg view components to the design system:

- create `CitizensAdviceComponents::Icons::Base`, which renders outer svg and associated props
- create component for each icon and preview all icons in a table
- add placeholder icon CSS for sizing and default colours
- add svg files into `/assets` (see below)

This PR is already quite noisy so I want to complete the following tasks in separate PRs:
- tidy up `/assets` (not used as far as I can tell), add it into the package, test in public-website and update icon docs
- icon slots for `CitizensAdviceComponents::Button` (and anything else?)
- add combined +/- up/down arrow icons with supporting CSS for easy toggling

Partial fix for #1396


engine/app/components/citizens_advice_components/icons/arrow_down.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render CitizensAdviceComponents::Icons::Base.new(size: size, icon_name: self.class.name) do %>
```
davidsauntson, line 1: Having to pass in the class name in every icon component isn't the DRY-est, but wasn't sure of a way of getting this from inside the base class.  I did have

```
def human_class_name
  self.class.name.....
end
```

But that's always just `base`, since I'm always creating a `Base` component with the `do` block.

davidrapson: No reason the class couldn't reasonably define an `icon_name` method.

```
class ArrowLeft < CitizensAdviceComponents::Icons::Base
   def icon_name
      "arrow-left"
   end
end
```

Also just wondering if having the component inherit from `CitizensAdviceComponents::Icons::Base` but then also rendering `CitizensAdviceComponents::Icons::Base` is a little confusing here. Wondering if there's something slightly back to front here.

engine/app/components/citizens_advice_components/icons/base.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module CitizensAdviceComponents
+  module Icons
+    class Base < CitizensAdviceComponents::Base
+      attr_reader :size
+
+      def initialize(size: nil, icon_name: nil)
+        super
+        @size = fetch_or_fallback(allowed_values: [nil, :small, :large], given_value: size, fallback: nil)
+        @icon_name = icon_name
```
davidrapson, line 11: Oh now I can't see where `icon_name` is used at all?

davidrapson (commented):
Looking great so far. Only thing I'm a little stuck on is the icon name behaviour. I think I might be missing where that's used.

And had a little question about the composition + inheritance for the base class. Wondering if the base icon class we inherit from is a different concept to the base icon component we render inside each icon.

engine/app/components/citizens_advice_components/targeted_content.html.haml
```diff
@@ -1,9 +1,11 @@
...
-          = render(CitizensAdviceComponents::Badge.new(type: :adviser))
-      = title
+  .heading-wrapper
+    - haml_tag("h#{heading_level}", heading_attributes) do
+      .cads-targeted-content__title-text
+        - if adviser?
+          .cads-targeted-content__badge
+            = render(CitizensAdviceComponents::Badge.new(type: :adviser))
+        = title
+    %button= "hi'"
```
davidrapson, line 9: 👋  Hiya

engine/previews/citizens_advice_components/icons_preview/icons.html.haml
```diff
@@ -0,0 +1,134 @@
+-# haml-lint:disable ViewLength
+.cads-grid-container
+  .cads-grid-row
+    .cads-grid-col-8
+      %table.cads-table
```
davidrapson, line 5: We could probably generate the rows in this table if they all follow the same pattern but I don't mind the longhand.

engine/app/components/citizens_advice_components/icons/base.rb
davidrapson: Oh wait. It's used in `human_class_name`. Yeah probably makes more sense for the specialised class to define that method itself.

davidsauntson: For posterity: it's used to render the correct CSS class, eg `cads-icon--arrow-left`.  Ultimately in `css_classes`, via `human_class_name`

engine/app/components/citizens_advice_components/targeted_content.html.haml
davidsauntson: LOLWUT

engine/app/components/citizens_advice_components/targeted_content.html.haml
```diff
@@ -1,9 +1,10 @@
 .cads-targeted-content.js-cads-targeted-content{ attributes }
-  - haml_tag("h#{heading_level}", heading_attributes) do
-    .cads-targeted-content__title-text
-      - if adviser?
-        .cads-targeted-content__badge
-          = render(CitizensAdviceComponents::Badge.new(type: :adviser))
-      = title
+  .heading-wrapper
```
davidrapson, line 2: Where does this `.heading-wrapper` class come from? Isn't this the class added dynamically by references?

davidrapson (approved):
⬆️ ⬆️ ⬇️ ⬇️ 🅱️ 🅰️ ⭐ 

Looks great! I reckon the `heading-wrapper` class in targeted content should go, or be a separate change if we do need it but otherwise fab.

PR #1525, opened 2021-09-29
Update more specs to use capybara matchers

davidrapson:
Update more specs to use capybara matchers. We don't have to use these for everything, it's fine to mix and match nokogiri with capybara where it makes sense. But generally speaking it's feeling more natural to write.


davidsauntson (approved):
👏🏻  big fan

PR #1526, opened 2021-09-29
Button component with icon slots

davidrapson:
## Background

Now that https://github.com/citizensadvice/design-system/pull/1524 has been merged we have the opportunity to provide a view component for buttons with support for icons.

## Changes

- Adds button component with support for left and right icon slots
- Updates icon component to default to `:small` size. Shouldn't be possible to have a `nil` size.
- Updates icon styles to use `currentColor` for the fill rather than setting a default fill colour that needs configuring.
- Updates docs to use generated examples and to document view component options

```haml
= render CitizensAdviceComponents::Button.new do |c|
  Next
  - c.icon_right do
    = render CitizensAdviceComponents::Icons::ArrowRight.new
```

## Screenshots

image


scss/4-elements/_buttons.scss
```diff
@@ -67,6 +67,13 @@ $button-shadow-size: 4px;
   &:active {
     border-color: $cads-language__primary-button-hover-colour;
   }
+
+  .cads-icon {
+    fill: currentColor;
```
davidsauntson, line 72: Yeah I think this is much more sensible - let's replace the default one in https://github.com/citizensadvice/design-system/blob/4100d28eea4d17f72cccc114922003579d0e63e7/scss/6-components/_icons.scss#L8 and associated settings

davidsauntson (commented):
I love it.

Left and right slots is good for me - we have designs for both currently.  Both at the same time would be bad but not too bothered about preventing people from doing that

engine/app/components/citizens_advice_components/icons/base.rb
```diff
@@ -5,9 +5,13 @@ module Icons
     class Base < CitizensAdviceComponents::Base
       attr_reader :size
 
-      def initialize(size: nil)
+      def initialize(size: :small)
         super
-        @size = fetch_or_fallback(allowed_values: [nil, :small, :large], given_value: size, fallback: nil)
+        @size = fetch_or_fallback(
+          allowed_values: [nil, :small, :large],
```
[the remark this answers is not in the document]
davidrapson: If you _really_ need an arbitrary size then that can always be done via component-specific CSS. Makes sense to standardise by default.

engine/app/components/citizens_advice_components/button.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module CitizensAdviceComponents
+  class Button < Base
+    renders_one :icon_left
+    renders_one :icon_right
+
+    attr_reader :variant, :type, :additional_attributes
+
+    def initialize(variant: :primary, type: :button, **additional_attributes)
```
davidsauntson, line 10: Oh interesting re `**additional_attributes`!  Just when I thought I knew ruby.  What are the advantages of doing it this way over how we did it in inputs etc?

davidsauntson (approved):

engine/app/components/citizens_advice_components/button.rb
davidrapson: The `**` kinda means the rest of the arguments has. Mostly makes the API a little simpler. 

```
= render CitizensAdviceComponents::Button.new(variant: :secondary, "data-thing": "test", "aria-label": "custom label")
```

vs.

= render CitizensAdviceComponents::Button.new(variant: :secondary, additional_attributes: { "data-thing": "test", "aria-label": "custom label"} )

PR #1527, opened 2021-09-29
Remove legacy haml partials

davidrapson:
We've been using the citizens_advice_components gem for all components for a while now. We're probably at the point where keeping the legacy haml partials around is limiting as we make more and more changes to view components.


davidrapson:
One to discuss tomorrow. Only thing I noticed whilst doing this is that we're missing view component docs for callouts.

davidsauntson:
Yeah fine - this will take us to 5.0.0 according to the roadmap.

davidrapson:
> Yeah fine - this will take us to 5.0.0 according to the roadmap.

Makes sense. As we're also looking at the SVG icons and other new components now feels like the time to do that.

davidsauntson (approved):
🔥

PR #1528, opened 2021-09-29
Improve SCSS tests

davidrapson:
Does a bit of work to improve the testability of our SCSS files.

- Structure SCSS tests around grouping (e.g. settings, grid)
- Add usage tests for mixins to assert output and confirm they can be imported standalone

Main focus is to treat sass tests more like we would unit tets.

This is also a good excuse to use `@use` to include any unresolved variables. We could stick with `@import` for a while longer but as we require dart sass the `@use` syntax is preferred.


davidsauntson (approved):
Wow!  This is great 🍰

PR #1531, opened 2021-09-30
Feat/assets tidy up

davidsauntson:
Stop it and assets tidy up.

- moves icons into `assets/svgs`
- moves fonts into `assets/fonts`
- adds `/assets` into package
- update Open Sans to v27
- remove support for legacy browsers

I've `npm link`ed into PW but would be good to get another pair of eyes on it.  Is there anything else I need to do to get `/assets` into the actual package?  I think it's just adding it to `package.json` but not 💯


davidrapson (approved):
👍  npm link test worked for me and yeah all that's needed to include the files is the package.json update.

scss/3-generics/_fonts.scss
```diff
@@ -2,39 +2,26 @@
...
-      format('embedded-opentype'),
-    /* IE6-IE8 */ url('../../fonts/open-sans-v17-latin-regular.woff2')
-      format('woff2'),
-    /* Super Modern Browsers */
-      url('../../fonts/open-sans-v17-latin-regular.woff') format('woff'),
-    /* Modern Browsers */ url('../../fonts/open-sans-v17-latin-regular.ttf')
-      format('truetype'),
-    /* Safari, Android, iOS */
-      url('../../fonts/open-sans-v17-latin-regular.svg#OpenSans') format('svg'); /* Legacy iOS */
+  src: local(''),
```
davidrapson, line 10: Only thing that might be worth keeping is the `local('Open Sans Regular'), local('OpenSans-Regular')` sources. This allows locally installed versions to be used if available.

davidsauntson: Ah yes, oops!  Thanks

PR #1532, opened 2021-09-30
Docs/icons

davidsauntson:
Some docs for the SVG icon components.


davidrapson (approved):
Docs look great. Only question is if we should merge the two docs as they reference each other anyway but happy to go with it as is 👍

PR #1533, opened 2021-09-30
Remove callout aria label

davidrapson:
We have a `title` property for callouts which confusingly maps to an `aria-label`.

Based on the real world usage of this there's no scenario where we provide this that isn't otherwise covered by the content or headings within the callout itself. It's either set to the same value as a heading or not set at all. Fairly sure this is a bug and we shouldn't be setting an aria-label here.

### DIFF
--- a/engine/app/components/citizens_advice_components/callout.html.haml
+++ b/engine/app/components/citizens_advice_components/callout.html.haml
@@ -1,4 +1,4 @@
-= tag.section(class: "cads-callout cads-callout--#{type}", "aria-label": title) do
+= tag.section(class: "cads-callout cads-callout--#{type}") do
   - if show_badge?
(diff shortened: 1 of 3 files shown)


davidrapson:
Edit: Tested this out with voice over and the only place this title gets set is adviser callouts and it just reads the heading out twice (once as the aria label, and then again for the heading). Definitely confusing things to set one.

davidsauntson (approved):
I think it came from a DAC report, but it should probably actually be a `labelled-by` rather than a repetition of the heading content.  

This is fine as is - in future, I wonder if we should allow consumers to pass in an id for the element that labels the section?

davidrapson:
Might come back to this one, feels like it's worth doing the `labelled-by` before we remove the existing label.

PR #1538, opened 2021-10-01
Fix: Match search button colour on hover

davidrapson:
When you hover over the search button the border colour doesn't change to match the background like regular buttons do. Updates the styles to match.


davidsauntson (approved):
🙈

PR #1557, opened 2021-10-05
chore: seperate form scss

davidsauntson:
Small housekeeping task to split the form components out into their own SCSS files, ahead of the additional components (checkboxes, select etc) that will be introducing soon.


davidrapson (approved):
👍

PR #1561, opened 2021-10-05
Feat/combined icons

davidsauntson:
Finishes #1396

Adds some combined icons for +/- and  &uarr; / &darr;, along with some supporting CSS for toggling between the two states.  There is some illustrative javascript in the examples, I think it's best to let people roll their own when it comes to using the combined icons though.


davidsauntson:
> Ooh interesting. I think I like this approach but curious about usage. This is the kind of thing that's often useful to inline into client-side code so curious if the view component wrappers help here.
> 
> Is it worth coming up with a more full fledged example of using the toggles in practice? Or is that already targeted content?

It's for convenience really.  There are a few use cases on public website where this would be useful already: 
- search filter and template letter show/hide for up and down arrows
- on this page, expandable sidebar components, targeted content for plus/minus

PR #1563, opened 2021-10-07
Test against multiple ruby versions

davidrapson:
A bit of a re-do of https://github.com/citizensadvice/design-system/pull/1307 but with just the Ruby version changes.

Includes a couple of Ruby 3 syntax fixes. The `**` stuff is down to https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

> When a method call passes a Hash at the last argument, and when it passes no keywords, and when the called method accepts keywords, a warning is emitted. To continue treating the hash as keywords, add a double splat operator to avoid the warning and ensure correct behaviour in Ruby 3.

Gist seems to be—based on the context Luke gave in the original PR—that Ruby 3 requires less ambiguity.

One bit of complexity here is that branch protection required checks rely on the job name. The matrix build generates four different names that would all need to be added to the required checks list. This is a bit unwieldy!

Thankfully we can happily move the demo-app build to be dependent on the engine matrix and limit the required status checks only to this dependent build. This is conceptually more in line with the separation between the engine and the storybook build. It was a bit misleading to have the demo-app dependent on storybook before.

TODO:

- [x]  Remove old engine job from required status checks. Rely on check for dependent demo-app job.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -57,6 +57,41 @@ jobs:
       - run: bundle exec haml-lint
       - run: bundle exec rake spec
 
+  demo-app:
+    needs: engine
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: ${{ env.NODE_VERSION }}
+
+    # We link the @citizensadvice/design-system package
+    # from the root but we only build the lib/ directory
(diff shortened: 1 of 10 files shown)


davidsauntson (approved):
Think this makes sense.  So we'll:
- always have one engine job for each version of ruby
- none of the engine jobs will be a required status check
- the demo-app will remain a status check
- if either of the engine jobs fail, so will the demo-app job and the branch won't be mergeable

PR #1577, opened 2021-10-12
Update greedy nav tests to use jest jsdom environment

davidrapson:
Bundles this upgrade which highlighted this issue https://github.com/citizensadvice/design-system/pull/1566

For some reason in our greedy nav tests we were manually constructing JSDOM instances rather than using the built-in jsdom environment that jest provides. Switch to this and remove the explicit dependency.


davidsauntson (commented):
The tests obviously still pass but - I know that jest doesn't reset the DOM after each test.  I can see you've manually set the `innerHTML`, but the greedy nav code also adds classes to the `html` element.  I see we weren't passing an`html` to the constructor before, and I don't think one is added automatically so it's probs fine.

davidrapson:
Might come back to this one, it's probably worth keeping the explicit html element for now.

PR #1612, opened 2021-10-27
Feat/checkboxes

davidsauntson:
Adds checkbox components to the design-system:

- a `CheckboxGroup` component
- a `CheckboxSingle` component
- create a `Checkable::Base` class, that is then subclassed by `Checkable::Radio` and `Checkable::Checkbox`
- refactor common ruby and haml into `FormGroup` component, use this for both checkboxes and radios
- refactored the SCSS to create common classes in `/forms/form-groups.scss` (just went with `.cads-form-group`
- created `/forms/checkboxes.scss` and `/forms/radios.scss` to provide type specific styles
- rspecs etc

This is already a lot so I'll do docs in a different PR.

http://localhost:3000/rails/view_components/citizens_advice_components/checkbox_group
http://localhost:3000/rails/view_components/citizens_advice_components/checkbox_single


engine/spec/components/citizens_advice_components/checkbox_group_spec.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
davidrapson: We could potentially use rspec shared examples for this stuff too…but tbh this feels nicer.

scss/6-components/forms/checkboxes.scss
```diff
@@ -0,0 +1,20 @@
+.cads-checkbox-group {
```
davidrapson, line 1: Might be tempted to use `cads-form-group--checkbox` to signal it's a BEM modifier, which is what this looks like. Although tbh this reads OK to me as well.

davidrapson (approved):
Lovely! One little comment about a class name (wondering out loud if cads-checkbox/radio-group should be a --modifier class) but that's a style preference so nothing blocking.

scss/6-components/forms/checkboxes.scss
davidsauntson: yup good shout

PR #1629, opened 2021-11-09
Docs/checkboxes

davidsauntson:
Adds checkbox documentation and vrts to the design-system


davidrapson (approved):
Looks great!

Might just be a backstop thing, but noticed the error message for a single checkbox looks a little wonky (maybe this is intentional?)

image

PR #1659, opened 2021-11-24
Design system docs

davidsauntson:
All aboard the train to Bridgetown.

This PR introduces the `design-system-docs` folder, which contains an installation of the Bridgetown v1.0.0 release candidate. It also:
- has the design system gem and npm package installed (using `file:`)
- has the `CitizensAdviceComponents` loaded and used to generate the header and nav
- a `cads_components` collection and one example component page available on http://localhost:4000/cads_components/targeted_content
- updates the linting rules for prettier, eslint and stylelint

You can see the additional parts by comparing these two commits: https://github.com/citizensadvice/design-system/compare/e08acf..edb858

image


design-system-docs/Gemfile
```diff
@@ -0,0 +1,30 @@
...
+# This will help ensure the proper Bridgetown version is running.
+#
+# To install a plugin, simply run bundle add and specify the group
+# "bridgetown_plugins". For example:
+#
+#   bundle add some-new-plugin -g bridgetown_plugins
+#
+# Happy Bridgetowning!
+
+gem "bridgetown", "~> 1.0.0.alpha10"
```
davidrapson, line 19: Fancy!

design-system-docs/bridgetown.config.yml
```diff
@@ -0,0 +1,31 @@
...
+# https://learnxinyminutes.com/docs/yaml/
+#
+
+url: '' # the base hostname & protocol for your site, e.g. https://example.com
+
+permalink: pretty
+
+template_engine: erb
+
+collections:
```
davidrapson, line 23: Think this makes sense for now, but do we have to have the `cads_` prefix in URLs? Is it work trying out the new Resources feature as we're already running the 1.0 alpha https://www.bridgetownrb.com/docs/resources

davidrapson (approved):
🎉 This looks fantastic.

Would be amazing if we could nix the `cads_` prefix from the URL but totally fine for now.

And do wonder if we should remove `yarn`. Feels confusing to switch between plain `npm` for the package and then `yarn` for the docs. But we should just go with the defaults rather than changing it right now.

design-system-docs/bridgetown.config.yml
davidsauntson: Yeah probs - there is already a collection called `components` that comes with the installation, and it's where bridgetown expects you to create your custom view components or whatevs, so I just prefixed to get the ball rolling.  Once for the ideation sesh, for sure.

davidrapson: Ah! Yeah that rings a bell. Cool, let's leave it like this for now.

PR #1662, opened 2021-11-25
Fix/callout last child margin

### LINKED ISSUE #1463: Margin removed from all last children inside a callout
## Summary of problem
The callout SCSS contains:

```scss
.cads-callout :last-child {
 margin-bottom: 0
}
```

Which removes the margin from _any_ last child inside a callout.  This causes a problem when wrapping headers and adding an adjacent button - since this is also a last child, the margin gets removed.  The intent of this ruleset was to remove the final margin from the content of the callout, so should be changed to:

```scss
.cads-callout > :last-child {
 margin-bottom: 0
}
```

marianayovcheva:
Fixes #1463


davidrapson (approved):
🥳

davidsauntson (approved):
😀  Nice!

PR #1673, opened 2021-12-01
Add rel=external styling for external links

davidrapson:
Our current prose styles have a bug that match any URL with https:// and citizensadvice.org.uk in it e.g. https://example.com/u/citizensadvice.org.uk/

Using string matches and hardcoding domains isn't especially reliable. In a future release we should remove these styles and favour using rel=external as a styling hook. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-external


scss/5-objects/_prose.scss
```diff
@@ -54,11 +54,20 @@
...
+  a[rel~='external']::after,
+  // These older rules have a bug that match any URL
+  // with https:// and citizensadvice.org.uk in it
+  // e.g. https://example.com/u/citizensadvice.org.uk/
+  // Using string matches and hardcoding domains isn't especially
+  // reliable. In a future release we should remove these styles
+  // and favour using rel=external as a styling hook.
+  // https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-external
   a[href^="https://"]:not([href*="citizensadvice.org.uk"])::after,
   a[href^="http://"]:not([href*="citizensadvice.org.uk"])::after
```
davidsauntson, line 67: I think we should also make this domain a variable.  My hunch is we will need to keep it for a while as can't mark a link as external in Contentful (AFAIK), so not sure how we'd do this.  Unless we get the content-api to sort it out.  Either way, our external won't be the same as someone else's so we'll need to let people edit it.

davidrapson: On the first one we definitely can determine if a link is external in our rendering code (will DM you a link to the renderer code that prompted this) but you are right it might not _always_ be possible to set `rel=external` so a config value for this might be a good idea. Maybe something like `$cads-external-link-domain: "https://www.citizensadvice.org.uk/" !default;`

davidsauntson (approved):
Lush! 🔗

PR #1693, opened 2021-12-08
Feat/sidebar

marianayovcheva:
I will be writing comments where I have questions, please have a look at those


engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
+%aside.cads-sidebar
+  .cads-sidebar__content
+    %p.cads-sidebar__title= title
+    - if section_title.url.present?
```
marianayovcheva, line 4: I think this section (line 4 to 9) needs to be changed a bit. It also threw me an error (which is why the test failed) but I couldn't figure out how to re-write it using &. (especially the `else` statement). I also saw that some of the other haml files also use if statements so not sure how to fix this one

`Run bundle exec haml-lint
app/components/citizens_advice_components/sidebar.html.haml:10 [W] RuboCop: Style/SafeNavigation: Use safe navigation (&.) instead of checking if an object exists before calling the method.`

engine/previews/citizens_advice_components/sidebar_preview/with_additional_content.html.haml
```diff
@@ -0,0 +1,12 @@
+= render CitizensAdviceComponents::Sidebar.new(title: "Related Content") do |c|
+  - c.section_title(section_title)
+  - c.sidebar_links(sidebar_links)
+  .example-links-block= link_to "Adviser resources - EU Settlement Scheme", "/advisernet/immigration/staying-in-the-uk/settled-status/adviser-resources-eu-settlement-scheme"
```
marianayovcheva, line 4: Not sure if this should have a special class like `cads-sidebar__additional-content`? I suppose it can just use the stying of the additional content used (e.g advisor link block)

engine/spec/components/citizens_advice_components/sidebar_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
```
marianayovcheva, line 1: I've made quite a lot of changes to this file, please let me know if the code is as expected (I've tested locally) or if something needs adjusting/adding

scss/6-components/_sidebar.scss
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  &__title {
+    color: $cads-language__secondary-text-colour;
+  }
+
+  &__section-title {
+    color: $cads-language__secondary-text-colour;
+
+    &--small {
```
marianayovcheva, line 15: Can I use a modifier here? public-website has a separate class

testing/visual-regression/backstop-common.js
```diff
@@ -184,6 +184,10 @@ module.exports = function backstopCommon(baseUrl) {
         focusSelector: '.cads-search__button',
         selectors: ['.cads-search'],
       },
+      {
```
marianayovcheva, line 187: Not sure if I should cover anything else here

engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
+%aside.cads-sidebar
```
marianayovcheva, line 1: In the end I've decided not to use the `.cads-grid-col-md-4.layout__sidebar` wrapper because there is a case in the public-website where the component is used within the 8-col layout (within the mobile toggle) https://github.com/citizensadvice/public-website/blob/develop/app/views/advice/_advice_collection.html.haml#L10, and I also saw there can be other content in that container ([preview placeholder and expandable sidebar](https://github.com/citizensadvice/public-website/blob/develop/app/views/advice/_advice_collection.html.haml#L39)) Please do let me know however, if I should use the wrapper classes here

davidrapson: That makes sense, those wrapper classes really site outside this component anyway

scss/6-components/_sidebar.scss
```diff
@@ -0,0 +1,46 @@
+.cads-sidebar {
```
davidrapson, line 1: Mind adding a `/** @define sidebar */` comment at the top of this file. That will tell the linter to check for BEM conventions.

davidrapson (commented):
This is looking flippin' amazing so far. I have a question about the naming. Does "Sidebar" as a name feel too layout specific? Wondering if there's a more content-focussed name for this component?

scss/6-components/_sidebar.scss
```diff
@@ -0,0 +1,46 @@
...
+    margin-bottom: 0;
+    padding-left: 0;
+  }
+
+  &__item {
+    margin-bottom: $cads-spacing-4;
+  }
+}
+
+.cads-targeted-content .cads-sidebar {
```
davidrapson, line 34: Is this meant to be here? It'd be surprising if this existed within targeted content?

engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
...
+  .cads-sidebar__content
+    %p.cads-sidebar__title= title
+    - if section_title.url.present?
+      %h3.cads-sidebar__section-title.cads-sidebar__section-title--small
+        = link_to section_title.title, section_title.url
+    - else
+      %h2.cads-sidebar__section-title
+        = section_title.title
```
davidrapson, line 9: Kinda curious about this part. We make the title a h3 and smaller if it's linked or a h2 if not? Am I misreading?

engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
...
+    %nav{ aria: { label: section_title&.title } }
+      %ul.cads-sidebar__links
+        - sidebar_links.each do |link|
+          %li.cads-sidebar__item= link_to link.title, link.url
+      = content
```
davidsauntson, line 14: I think it's fine for the `content` to be rendered into the same nav (it's what happens in public website), we'll need to make sure that this is explicitly mentioned in the docs though, so people know.

engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
+%aside.cads-sidebar
+  .cads-sidebar__content
+    %p.cads-sidebar__title= title
+    - if section_title.url.present?
+      %h3.cads-sidebar__section-title.cads-sidebar__section-title--small
+        = link_to section_title.title, section_title.url
+    - else
+      %h2.cads-sidebar__section-title
+        = section_title.title
+    %nav{ aria: { label: section_title&.title } }
```
davidsauntson, line 10: I wonder if we should make the title mandatory and part of .new's parameters, because we will always want to have an aria-label for the nav.

Then you could get rid of the `&`. As it stands, if there is no title we won't get an `aria-label`.

engine/previews/citizens_advice_components/sidebar_preview/with_additional_content.html.haml
davidsauntson: Yeah I think people can style whatever they want to put in, so I don't think we need it.  If they have a specificity issue they can always use `.cads-sidebar` instead.

engine/spec/components/citizens_advice_components/sidebar_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  it "renders the title" do
+    expect(component.text.strip).to include "Related Content"
+  end
+
+  it "renders the section title" do
+    expect(component.text.strip).to include "Applying to the EU settlement scheme"
+  end
+
+  it "renders a link for each sibling" do
+    expect(component.css(".cads-sidebar__item").count).to eq sidebar_links.count
```
davidsauntson, line 33: You could add a `[data-testid]` attribute to each link, and count those instead.  Testing against a styling CSS class tends to make the test more fragile.

scss/6-components/_sidebar.scss
davidsauntson: Is there another variant? If not you can just put it all into the same selector ruleset.

scss/6-components/_sidebar.scss
davidsauntson, line 34: I think this is probably only needed in public website because of the mobile version of the sidebar collapses into a targeted content.  I think that's a choice that public website has made so we shouldn't cater for it in the design system, what do you think?

testing/visual-regression/backstop-common.js
davidsauntson: This is fine for now I think

engine/app/components/citizens_advice_components/sidebar.html.haml
```diff
@@ -0,0 +1,14 @@
...
+      %ul.cads-sidebar__links
+        - sidebar_links.each do |link|
+          %li.cads-sidebar__item= link_to link.title, link.url
```
davidsauntson, line 13: The links should have the `.cads-hyperlink` class (they styles are currently applied via the `a` selector but it's probably worth being specific)

davidsauntson (commented):
Really great! 👏🏻 

A few thoughts in the comments but nothing major.  Please could you double check the line height of the links?  They look a squished to me on the mobile version.

davidsauntson:
> This is looking flippin' amazing so far. I have a question about the naming. Does "Sidebar" as a name feel too layout specific? Wondering if there's a more content-focussed name for this component?

That's what I went with on the issue, and I am famously bad at naming.  Some other thoughts I had include:
- `secondary nav`
- `section nav`
- `sibling nav`

These are more content focussed but less obvious, to me.

Any suggestions welcome!

styleguide/components/sidebar/sidebar-docs.mdx
```diff
@@ -0,0 +1,40 @@
...
+</Preview>
+
+## Using with Rails
+
+If you are using the `citizens_advice_components` gem, you can call the component from within a template using:
+
+```rb
+= render CitizensAdviceComponents::Sidebar.new(title: "Related Content") do |c|
+    c.section_title(title: "Applying to the EU settlement scheme", url: "/immigration#h-applying-to-the-eu-settlement-scheme")
+    c.sidebar_links([
```
davidsauntson, line 22: It might make the docs a bit easier to read if the links array was a bit shorter and the link text inside was also a bit shorter / more illustrative (eg `{ title: "Link 1", url: "/link-1" }` or smth,

scss/6-components/_sidebar.scss
davidsauntson: Ahhh I just had a look in public website - I think this has come from the fact the `AdviceCollectionLinks` component is also used to render the advice list pages.  We'll probably change that when this component is done - the sidebar shouldn't concern itself with the styles/markup for the advice collection lists pages.  So we can just have one ruleset here.

engine/app/components/citizens_advice_components/sidebar.html.haml
davidsauntson: See: https://github.com/citizensadvice/design-system/pull/1693/files#r765567643

I think we can just have the `h3` regardless of whether there it contains a link or not

styleguide/components/sidebar/sidebar-docs.mdx
marianayovcheva: Good point. Should I also change the links array in other places it's used (spec, preview and example files) or is it okey to only have it illustrative in the docs?

engine/app/components/citizens_advice_components/sidebar.html.haml
marianayovcheva: This makes sense. I'm not sure though how to properly write it - in this case should i get rid of the slot as it will only contains the section_title url? A side question (because I just don't know the answer) - if something is required in a slot class (`SectionTitle`), won't it be required on the main object as well (`Sidebar`)?

engine/app/components/citizens_advice_components/sidebar.html.haml
marianayovcheva: I have just pushed a change to add it but I'm not sure if the wording is ok

davidrapson:
Chewing on this a bit I think the component should have the word "links" in it as it's ultimately a collection of links with a title and prefix. One name that comes to mind is `SectionLinks`. We use section title to describe what the title is doing so perhaps aligning it with that makes sense.

davidrapson:
Another option if we don't want to block on coming up with a name is that we _could_ use this as an excuse to try out an experimental namespace. e.g. under `CitizensAdviceComponents::Experimental::Sidebar` and maybe don't include the styles in the default export. That'd let us merge this early, use it in the docs site, and give us the opportunity to rename it if we want when we actually promote it to the main namespace. Just a 💭 as an option available to us.

styleguide/components/sidebar/sidebar-docs.mdx
davidsauntson: It's okay for it to be illustrative in the docs, I don't think you need to change the others

engine/app/components/citizens_advice_components/sidebar.html.haml
davidsauntson: Yeah maybe have title as required and url as optional:

```rb
def initialize(section_title:, section_title_url: nil)
  ...
end
```

(or whatever you want to call the params)

davidsauntson: On the second point - I don't know for sure but I think if you try and instantiate a slot class without a required param then it'll fail

engine/app/components/citizens_advice_components/sidebar.html.haml
davidsauntson: Ah OK - yeah I think we need to emphasise the fact that `content` will be rendered into a `nav` element so they'll need to bear that in mind when using it.

styleguide/components/sidebar/sidebar-docs.mdx
```diff
@@ -0,0 +1,40 @@
...
+<Preview>
+  <Story id="components-sidebar--example" />
+</Preview>
+
+## With Additional Content
+
+<Preview>
+  <Story id="components-sidebar--with-additional-content" />
+</Preview>
+
```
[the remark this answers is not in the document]
davidsauntson: This is fine for now, apart from the other comment about `content` being inside `nav`

engine/spec/components/citizens_advice_components/sidebar_spec.rb
davidsauntson: These look good.

I think you should also test for when:
- no section title url is passed in, and assert that no link is rendered in the title
- if no links are passed, nothing is rendered

engine/app/components/citizens_advice_components/sidebar.html.haml
davidsauntson, line 4: You can have this `if` inside the `h3`, that way you don't have to duplicate the `h3` element

davidsauntson (approved):
🚄 📊 

Great!

PR #1695, opened 2021-12-10
fix: Changed the error summary docs to be consistent with public-website

marianayovcheva:
Using plural `c.errors` array in the docs and preview file to be consistent with the public-website markup (see [PR2155](https://github.com/citizensadvice/public-website/pull/2155))


davidrapson (approved):
Makes sense. The `renders_many` slots let you call both plural and singular methods, but I definitely prefer the plural version for things like this too 👍

Issue #1671, opened 2021-11-30
On this page


davidsauntson:
# On this page component

The 'On this page' component is currently used by public website and advisernet to show a list of shortcuts to headers on a page (for users that are signed in):
- https://www.citizensadvice.org.uk/immigration/staying-in-the-uk-if-youre-from-the-eu/
- https://www.citizensadvice.org.uk/advisernet/immigration/advising-on-immigration-and-nationality1/what-you-can-advise-on/

The new design system docs will also make use of this component, so let's extract the component from public-website and bring it into the design system.

The time has come to extract a generic component that allows us to:
 - provide a list of links to be displayed 
 - optionally nest links one level deep to provide the expand/collapse function you see in advisernet

Figma: https://www.figma.com/file/Im9r6ZM1YWiJK488OWx8a6/AdviserNet-Template-Redesign?node-id=5197%3A35582
---

- does your component need to be used by other products or services?
> yes - public website, advisernet, design-system docs

- does something similar already exist in the design system
> no

- is there a product team that would perform the research, design, development and QA work associate with building this component?
> content-platform has done this (there will be some refactoring to do)

- do you have any user stories / scenarios that describe the component?
> https://github.com/citizensadvice/public-website/blob/develop/features/advice/combined/on_this_page.feature

### Code

- provide a link to your component working in a dev or prod environment

> https://www.citizensadvice.org.uk/immigration/staying-in-the-uk-if-youre-from-the-eu/
> https://www.citizensadvice.org.uk/advisernet/immigration/advising-on-immigration-and-nationality1/what-you-can-advise-on/

- provide links to the code for this component
> https://github.com/citizensadvice/public-website/blob/develop/app/components/on_this_page_component.rb
> https://github.com/citizensadvice/public-website/blob/develop/app/components/on_this_page_component.html.haml

### Usability

- have you tested the component with the intended user group? Please include any links to research documentation in your proposal

> yes, tbc with Kostas and UR

### Accessibility

- have you ensured your component meets WCAG 2.X AA guidelines? Include test protocols from external audits or notes from internal tests
- have you tested your component with a screen reader?
- have you tested your component by using only the keyboard to interact with it?
> yes it was part of the DAC audit done for the new immigration section

### Quality

- provide links to any appropriate automated tests you have written
  - unit tests https://github.com/citizensadvice/public-website/blob/develop/spec/components/on_this_page_component_spec.rb 
[...]

PR #1707, opened 2021-12-17
Feat/on this page

marianayovcheva:
Moving the _On this page_ component from public-website into the design-system. Details outlined here: https://github.com/citizensadvice/design-system/issues/1671


engine/app/components/citizens_advice_components/on_this_page.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def columns?
+      links.size > 3
+    end
+
+    def show_nested_links?
+      @show_nested_links
+    end
+
+    def children_exist?
```
marianayovcheva, line 40: I'm not sure I've done the children_exist logic properly so it's worth checking

styleguide/components/on-this-page/on-this-page.stories.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const Example = () => templateExample;
+Example.parameters = { docs: { source: { code: templateExample } } };
+
+export const WithNestedLinks = () => templateWithNestedLinks;
+WithNestedLinks.parameters = {
+  docs: { source: { code: templateWithNestedLinks } },
+};
+WithNestedLinks.decorators = [decorateWithBehaviour];
```
marianayovcheva, line 24: In this file I had to add the decorator only on `WithNestedLinks`. It was running the js on the `Example` as well which caused problems - the nested links wouldn't open as the toggle was getting confused.

engine/app/components/citizens_advice_components/on_this_page.rb
davidsauntson: The logic looks OK to me.

You can simplify it a bit by not passing the link into its own `children?` method, and use the instance variable `@children` to see if they are present.

So you would have
```rb 
def children_exist?
  links.any? { |link| link.children? }
end
```
and in `class Link` have
```rb
def children?
   @children.present?
end
```

engine/spec/components/citizens_advice_components/on_this_page_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+          "Link 5"
+        ]
+      end
+
+      it "renders nested links" do
+        expect(component.at("[data-testid='on-this-page-children']")).not_to be_nil
+      end
+
+      it "renders toggle buttons" do
+        expect(component.at("[data-testid='on-this-page-toggle']")).not_to be_nil
```
davidsauntson, line 102: You could make this a bit more specific and assert that the number of toggles present is correct maybe?

davidsauntson (approved):
Looking great!

Small question about one of the rspecs and a possible simplification of `children_exist` but nothing major 👍🏻

davidsauntson:
OH please can you update the changelog too, thanks 🙏🏻

PR #1714, opened 2021-12-20
fix: on this page config

davidsauntson:
I made a branch from your branch with a working config setup.  Your initial directory structure for `src/tsc/` was 👌🏻 - here I've ended up with:

```
- src/tc/
-- on-this-page
--- on-this-page.ts
-- utils
--- show-hide.ts
``` 

It was a good shout to use project references - this is the correct way to handle these dependencies using `tsc`.

To fix the webpack issue, we need to make sure the built directory structure in `lib` ends up being what webpack is expecting.  It was complaining that it couldn't find `../utils/show-hide.js` from `lib/on-this-page.js`, which is fair enough because resolving that relative location would be outside of `lib` and therefore 💥.

To fix it we just need to put the built `on-this-page.js` into its own folder, like we do with `greedy-nav`.  It's not ideal because it's a bit inconsistent to have some built modules in folders and some not, but we already have that problem with greedy nav.  It's something we'll need to address in the future (looks like we'll end up with `module/module.js` being the pattern).

Check this out, see if it works on your local set up and, if you're happy with it, merge it into your `feat/on-this-page` branch.


davidsauntson (commented):
This should probably be
`/lib/on-this-page/index`

demo/app/frontend/packs/application.js
```diff
@@ -1,7 +1,7 @@
 import initHeader from '@citizensadvice/design-system/lib/header';
 import greedyNav from '@citizensadvice/design-system/lib/greedy-nav/index';
 import initTargetedContent from '@citizensadvice/design-system/lib/targeted-content';
-import initOnThisPage from '@citizensadvice/design-system/lib/on-this-page';
+import initOnThisPage from '@citizensadvice/design-system/lib/on-this-page/on-this-page';
```
davidsauntson, line 4: This should probably be `/lib/on-this-page/index`

marianayovcheva (approved):
🙏

PR #1742, opened 2022-01-26
Fix/error summary group ids

### LINKED ISSUE #1729: Radio button group input ids
# Radio button group & Checkbox group
Raising this as more of a discussion point first.

## Summary of problem

Radio button inputs are given an id using the following method:

https://github.com/citizensadvice/design-system/blob/c90af45f6f53570f1bded935e1e841ad767c5f8c/engine/app/components/citizens_advice_components/checkable/base.rb#L46

Which results in `id` attributes like this:
- `how-would-you-like-to-be-contacted-Please contact me by email`
- `how-would-you-like-to-be-contacted-Please contact me by telephone`
- `how-would-you-like-to-be-contacted-Please send me a letter in the post`

This is fine until there is an error - then the consuming application needs to know both the `name` of the input and its `value` so that it can construct the anchor link correctly for the error summary, so we can give focus to the invalid input.

All other inputs only require the consuming application to know the `name` of the input to construct the error summary url.  I've worked through the problem in public website where it occurs, but it's a bit grim and inconsistent with other form controls.

## Possible solution

I think the design-system could be a bit nicer by using an index to generate the `id` attribute instead, eg:

- `how-would-you-like-to-be-contacted-0`
- `how-would-you-like-to-be-contacted-1`
- `how-would-you-like-to-be-contacted-2`

Then we can rely on `<blah>-0` being the input we need to focus on when the radio button group fails validation.

davidsauntson:
Updates the checkbox and radio button components so that they construct their input ids using the index of the input rather than the value.

Fixes #1729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 **Bugfixes**
 - 🔗 Links: Style links with `rel=external` in `.cads-prose` as a more reliable way of determining external links.
+- 🛂 Ids: Radio buttons and checkbox `id` attributes are now generated using the input's index instead of value
 
 ## <sub>v5.2.0-alpha.1</sub>
 
--- a/engine/app/components/citizens_advice_components/checkable/base.rb
+++ b/engine/app/components/citizens_advice_components/checkable/base.rb
@@ -13,8 +13,9 @@ module CitizensAdviceComponents
         @additional_attributes = additional_attributes
       end
 
-      def attributes(name, error: false)
+      def attributes(name, index, error: false)
         @name = name
+        @index = index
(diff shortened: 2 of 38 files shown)


marianayovcheva (approved):

davidrapson (approved):
👍 Looks great!

Small non-blocking nit pick from me: I starting with 1 might make more sense for a HTML ID, rather than zero-indexed. But I don't feel strongly about that.

davidsauntson:
> 

Cool - fairly happy with it as it is, since it'll be simpler for a back-end to work with zero based.

PR #1752, opened 2022-01-26
Feat section links (previously sidebar)

marianayovcheva:
This is modifying existing design-system component.
After initially importing the section links component in the [public-website](https://github.com/citizensadvice/public-website/pull/2247), I realised there are a couple of changes needed:

- added `additional_attributes` to the section_links ( allows adding things like aria-current) and modified the tests and documentation accordingly
- extended the render condition to allow rendering if only content, links or section title are present


davidsauntson (approved):
Lovely stuff! 🍰

PR #1768, opened 2022-02-09
NP-2452 remove browserstack code and refactor some scenarios

celonicolosi:
This PR removes all Browserstack code related from the Design System repository and do a  refactor in some scenarios removing tags that are not necessary, like "@not_mobile", "@not_tablet" and updating some "@failing", "@future_release".


davidrapson (approved):
🗑️

PR #1778, opened 2022-02-14
feat: doc workflow

davidsauntson:
- moves control of preview deployment to our github actions workflow
- adds publishing to production url when a new release is created (need to test this once it's in `main`)


marianayovcheva:
@davidsauntson, I'd suggest @davidrapson will be better in reviewing this one as I'm not so familiar with workflow files.
But do give me a shout if it's more of a case of approving it so you can test on `main` and I'll hit the button.

davidrapson (approved):
🚀 

Will be interesting to see how the comments feel as there's quite a lot on this PR, but we should just try it out for a bit. Looks great.

davidrapson:
It works! Ship it 🚀

Issue #1718, opened 2022-01-05
Disclosure View Component


marianayovcheva:
# Disclosure
The disclosure component already exists as part of the public-website and is used to show/hide search filters. It will also be used in the template letters. 

The time has come to extract a generic component that allows us to:
- provide button titles:
   - required closed title
   - optional open title (if not provided, the closed title will be used)
- provide content that will appear when the disclosure is open
___
- does your component need to be used by other products or services?
> it is currently used by the [search filters](https://www.citizensadvice.org.uk/s/?q=benefits) on the public-website and by [template letters](https://web.develop.content.citizensadvice.org.uk/new-platform/a-template-letter/)
- does something similar already exist in the design system
> No, but it can reuse the show-hide functionality with additional js (similar to on this page component)
- is there a product team that would perform the research, design, development and QA work associate with building this component?
> Probably content-platform?
- do you have any user stories / scenarios that describe the component?
> not currently

### Code

- provide a link to your component working in a dev or prod environment
> https://www.citizensadvice.org.uk/s/?q=benefits&filter_by=advice
- provide links to the code for this component
> https://github.com/citizensadvice/public-website/blob/develop/app/components/disclosure_component.rb
> https://github.com/citizensadvice/public-website/blob/develop/app/components/disclosure_component.html.haml

### Usability

- have you tested the component with the intended user group? Please include any links to research documentation in your proposal
> It was initially tested on equivalent for search filters on the Intranet. It was not used in the end, however, once prompted the users were able to use the toggle so the behaviour test passed.

### Accessibility

- have you ensured your component meets WCAG 2.X AA guidelines? Include test protocols from external audits or notes from internal tests
- have you tested your component with a screen reader?
- have you tested your component by using only the keyboard to interact with it?
> not yet tested

### Quality

- provide links to any appropriate automated tests you have written
  - unit tests
  > https://github.com/citizensadvice/public-website/blob/develop/spec/components/disclosure_component_spec.rb
  - behavioural tests
  > can be extracted from here: https://github.com/citizensadvice/public-website/blob/develop/features/advice/public/advice_only_search.feature#L5
  - visual regression tests
  > this test can be modified to be more general and not search specific: https://github.com/citizensadvice/public-website/blob/develop/testing/visual-regression/scenarios/public-advice.js#L76

marianayovcheva:
@davidsauntson, here are the bits I'm not sure about:
[...]

PR #1781, opened 2022-02-16
Feat/disclosure

marianayovcheva:
Added the disclosure component which already exists in public-website. This addresses Issue #1718.
I have added a couple of question I have about BDD - and that section will generally require more attention when reviewing.


testing/features/support/components/disclosure/example.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Disclosure
+  class Example < ::Base
+    set_url "/iframe.html?id=components-disclosure--example&args=&viewMode=story"
+
+    element :toggle, ".cads-disclosure__toggle"
```
marianayovcheva, line 7: Not sure if it's better for me to add `data-test-id` instead

testing/features/components/step_definitions/disclosure_steps.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Given("the Example Disclosure component is on the page") do
+  @component = Disclosure::Example.new.tap(&:load)
+end
+
+When("I open/close the disclosure component") do
```
marianayovcheva, line 7: This works but I don't fully get how it works - i.e how open/close can then be translated to two expressions later (i.e `I open the disclosure component` and `I close the disclosure component`. Links to docs would be greatly appreciated - I've tried to look myself, but I'm probably not wording the search term properly.

testing/features/components/step_definitions/disclosure_steps.rb
```diff
@@ -0,0 +1,17 @@
...
+  @component.toggle.click
+end
+
+Then("The disclosure component displays expanded content") do
```
marianayovcheva, line 11: For those two I've tried to match what's in [public-website](https://github.com/citizensadvice/public-website/blob/develop/features/advice/public/step_definitions/search_steps.rb#L96). The problem I was having was that when I was using `.not_to be_visible` was throwing an error:
image
What I have now works, however I want to understand why `.not_to be_visible` was not working in this case

testing/features/components/step_definitions/disclosure_steps.rb
davidrapson: These are cucumber expressions. The only docs I could find are the Java one but the syntax is the same https://github.com/cucumber/cucumber-expressions#readme

testing/features/support/components/disclosure/example.rb
davidrapson: I don't mind the classes for design system components as class names are ultimately part of the public API. `data-testid` would be fine too.

davidrapson (approved):
💯 This looks fantastic

PR #1786, opened 2022-02-23
Remove unused testing gems

davidrapson:
## Background

This pull request makes some changes to simplify our selenium testing set up with the longer term goal of being able to decouple these tests from storybook and open ourselves up to explore alternative options for this kind of testing. For now, I'm mostly deleting code.

Quite a lot going on here so will try and call out the major changes.

## Changes

**Remove dependency on ca_testing gem**

This is an unmaintained library which consists solely of patches that don't apply to the context design system tests are run in so we can remove this. Replaced this entire gem with a handful of lines of standard webdriver code so at least we know it's not needed for anything.

**Remove unused support code and browser specific logic**

Lots of this was mostly to handle browser edge cases in grid tests copied over from public-website. We only use a single headless runner in CI so this code never applied.

**Remove custom logging**

Our selenium tests were the only check that used file based logging rather than using stdout. GitHub actions already has decent log management based on stdout so we don't need to reimplement our own here.

There was a huge amount of code here to support this when we don't need it. Remove all custom file based logging code and rely on default behaviour. Tested this in the following action run: https://github.com/citizensadvice/design-system/runs/5307795336?check_suite_focus=true

**Cleanup**

The remaining changes are around cleanup up any unused code paths and scoping the test code to the `testing/` sub-directory to keep the top-level project simpler.

## Next steps

The next steps after this are to:

- Strip the feature scenarios right back to just those that have interaction (e.g. header, navigation, targeted content)
- Use this as a baseline to first migrate over to run against the demo app, or at least not storybook, and then to give us a much simpler set of scenarios to use to compare alternative approaches for behavioural testing.


testing/features/components/asset_hyperlink.feature
```diff
@@ -7,6 +7,5 @@ Feature: Asset Hyperlink component
     Given an Asset Hyperlink component is on the page
     Then a link to the PDF is present
     And the label includes a name at the beginning
-    And the label includes an icon
```
davidrapson, line 10: Removed this because it relied on some hyper-specific test code and it's a pretty visual test covered by backstop.

testing/features/business_rules/automated_release_process.feature
```diff
@@ -1,27 +0,0 @@
-Feature: Automated Release Process
```
davidrapson, line 1: I feel we should avoid using this "business rules" directory for design system stuff, but do we want to keep this elsewhere?

testing/features/components/navigation.feature
```diff
@@ -29,12 +27,6 @@ Feature: Navigation component
       Then the Close button is present
       And the dropdown menu is open
 
-    @failing @NP-1755
```
davidsauntson, line 32: This does fail and this is one of the few components that need behavioural tests so I'd be reluctant to delete this

testing/features/business_rules/automated_release_process.feature
davidsauntson: I just closed the ticket for this work in public website since we'll be reviewing all this once the new docs site is done - so happy for it to go.  We can always come back here if we want it.

davidsauntson (commented):
❤️ so great to get rid of all the testing code we don't need anymore. 🙌🏻 

There is one test I think we should keep - I've added a comment.

testing/features/components/navigation.feature
davidrapson: The `@failing` tag means it's never run so will see if I can't wrangle things to get this running again.

PR #1817, opened 2022-02-28
Upgrade citizens-advice-style-ruby in engine

davidrapson:
- Upgrade citizens-advice-style-ruby in engine to v7
- Fix warnings from version upgrade
- Add engine to dependabot rules

### DIFF
--- a/engine/Gemfile
+++ b/engine/Gemfile
@@ -6,7 +6,7 @@ gemspec
 
 group :development, :test do
   gem "capybara"
-  gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby", tag: "v2.1.0"
+  gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby", tag: "v7.0.0"
   gem "haml_lint"
   gem "i18n-tasks"
   gem "pry", "~> 0.14.1"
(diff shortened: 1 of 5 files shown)


davidrapson:
@mrdaniellewis FYI that you reminded me that the `citizens-advice-style-ruby` version used in the design system engine code was a bit out of date. Updating to the latest now so we have another project tracking the newest version.

engine/Gemfile
```diff
@@ -6,7 +6,7 @@ gemspec
 
 group :development, :test do
   gem "capybara"
-  gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby", tag: "v2.1.0"
+  gem "citizens-advice-style", github: "citizensadvice/citizens-advice-style-ruby", tag: "v7.0.0"
```
davidrapson, line 9: This big jump was fine because we're already using `rubocop-rspec` and the engine ruby code is pretty simple.

davidrapson:
> I am amazed those are the only changes you had to make, though. We often have to change hundreds of files with every new rule.

This is a comparatively tiny codebase and we were already doing some of the shared practice stuff like using `rubocop-rspec` so there's not much that _could_ change tbh.

PR #1818, opened 2022-02-28
fix: release script does not create tag

davidsauntson:
Prevents the release script from adding a tag for the release by using [`--no-git-tag-version`](https://docs.npmjs.com/cli/v8/commands/npm-version#description), as the docs say to do this manually after publishing.  

Duplicate tags are potentially causing [issues](https://github.com/citizensadvice/public-website/pull/2384) in products that have the gem as a github dependency.


davidsauntson:
Actually thinking about this we can probably get the workflow to create the tag?

davidrapson:
> Actually thinking about this we can probably get the workflow to create the tag?

The release process will do this for you. If you create a release through the UI and type in a non-existent tag name it will create a tag at the same time as creating the release.

davidrapson (approved):

PR #1822, opened 2022-03-03
Upgrade engine dependencies

davidrapson:
Upgrade engine dependencies. Includes https://github.com/citizensadvice/design-system/pull/1819


engine/lib/citizens_advice_components/engine.rb
```diff
@@ -3,7 +3,7 @@
 require "haml-rails"
 require "rails-i18n"
 require "active_support/core_ext"
-require "view_component/engine"
+require "view_component"
```
davidrapson, line 6: This fixes a deprecation warning with a recent version of view component

Issue #1398, opened 2021-08-05
Migrate Design System component tests from Storybook to demo app


davidsauntson:
With the introduction of the citizens_advice_components gem we have a demo Rails app which will eventually contain the vast majority of the examples used in our docs generated using view component previews.

We could reasonably use this demo app to back our selenium, backstop, and pa11y tests. It’s a simple Rails app with no other behaviour to interfere with the tests.

By comparison the current Storybook environment is not representative as it’s a heavy client-side app with limited browser support which leads to a lot of false-negatives and workarounds when running the current tests.

Goals

- Tests still run and pass as before (No regressions)
- The components being tested are roughly speaking the same as before (Previously tested in Storybook iFrames)

OG Jira ticket for conversation history here: https://citizensadvice.atlassian.net/browse/DS-19

PR #1757, opened 2022-01-31
Saucelabs experimental branch

celonicolosi:
I would not ask for a review (for now) from you guys if I could, **so please ignore this PR!!**

Just an experimental branch, **do not merge!**
I need to see how the tests are running today, experimenting with Browserstack x Saucelabs, GitHub action, etc.

PR #1841, opened 2022-03-15
Test basic cypress workflow

davidrapson:
## Background

Bit of an experiment into https://github.com/citizensadvice/design-system/issues/1398 to move more of our tests to run against the demo app using Cypress.

## Changes

- Adds Cypress testing support to the demo app using https://github.com/testdouble/cypress-rails
- Adds an accessibility smoke test for every component preview using https://github.com/component-driven/cypress-axe
- Adds reference examples of component tests for: navigation, skip links, disclosure, targeted content, and pagination. Basically everything that has an interaction to it with the exception of the header which I can add if this looks good.

I've added explanatory comments to each step to call out the main changes.

## How to run the tests

From within the `demo/` directory you can run the following commands.

**Running using the cypress UI**

```
./bin/rails cypress:open
```

This command will open the Cypress UI for working with tests.

image

**Running headless tests**

```
./bin/rails cypress:run
```

This command will run all tests headlessly

image

# TODO

- [x] Update README with instructions
- [x] Add engine-lint job to required status checks after merging (https://github.com/citizensadvice/design-system/pull/1843)

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -64,12 +64,9 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
+          bundler-cache: true
           ruby-version: '2.7.4'
           working-directory: engine
-      - run: |
-          bundle config path vendor/bundle
-          bundle update
-          bundle install --jobs 4 --retry 3
       - run: bundle exec rubocop --format github
       - run: bundle exec haml-lint
 
@@ -97,13 +94,34 @@ jobs:
           bundle update
           bundle install --jobs 4 --retry 3
         working-directory: demo
+
+      # Cache cypress binary
+      - uses: actions/cache@v2
+        with:
+          path: /home/runner/.cache/Cypress
+          key: cypress-binary-${{ hashFiles('package-lock.json') }}
+
+      # Npm install
       - uses: bahmutov/npm-install@v1
         with:
           working-directory: demo
 
-      # Run a webpack build as a smoke test
+      # Run a webpack build
       - run: ./bin/webpack
         working-directory: demo
+
+      # Run cypress tests
+      - run: ./bin/rails cypress:run
+        working-directory: demo
+        env:
+          CYPRESS_RAILS_CYPRESS_OPTS: '--browser chrome'
+
+      # Archive test screenshots
+      - uses: actions/upload-artifact@v2
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+        if: ${{ failure() }}
 
   # Actions can run in parallel at the job or the workflow level.
   # In order to run our build-dependent tests in parallel we need a way
--- a/demo/cypress/integration/axe.spec.js
+++ b/demo/cypress/integration/axe.spec.js
@@ -1,0 +1,13 @@
+context('Accessibility', () => {
+  it('all component previews pass a basic accessibility check', () => {
+    cy.visit('/rails/view_components');
+    cy.get('li a').then((tags) => {
+      Array.from(tags)
+        .map((t) => t.href)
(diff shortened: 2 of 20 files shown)


demo/cypress/integration/axe.spec.js
```diff
@@ -0,0 +1,13 @@
...
+    cy.visit('/rails/view_components');
+    cy.get('li a').then((tags) => {
+      Array.from(tags)
+        .map((t) => t.href)
+        .forEach((href) => {
+          cy.visit(href);
+          cy.runAxe();
+        });
+    });
+  });
```
davidrapson, line 12: This test essentially allows us to replace pa11y with an automated check within the cypress tests using `cypress-axe`. This list is against every preview example so we don't need to keep any lists in sync.

We can also run additional checks during scenarios if we want, e.g. to check again when targeted content is open. For now I've left it to this smoke check.

demo/cypress/integration/disclosure.spec.js
```diff
@@ -0,0 +1,27 @@
+describe('Disclosure', () => {
+  it('should allow toggling disclosure', () => {
+    cy.visitComponentUrl('disclosure/example');
+
+    assertClosed();
+
+    cy.findByRole('button', { name: /View/i })
+      .click()
+      .should('have.attr', 'aria-expanded', 'true');
```
davidrapson, line 9: This uses https://testing-library.com/docs/cypress-testing-library/intro/ to do things like find elements by role.

demo/cypress/support/commands.js
```diff
@@ -0,0 +1,65 @@
...
+Cypress.Commands.add('runAxe', () => {
+  cy.injectAxe();
+
+  cy.configureAxe({
+    rules: [
+      // Skip links are missing a unique landmark name
+      { id: 'landmark-unique', enabled: false },
+
+      // Table overflow needs to be focusable
+      { id: 'scrollable-region-focusable', enabled: false },
```
davidrapson, line 54: These are two extra accessibility warnings that axe flagged up that the previous pa11y tests missed.

.github/workflows/ci-workflow.yml
```diff
@@ -81,14 +94,35 @@ jobs:
...
+        working-directory: demo
+        env:
+          CYPRESS_RAILS_CYPRESS_OPTS: '--browser chrome'
+
+      # Archive test screenshots
+      - uses: actions/upload-artifact@v2
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+        if: ${{ failure() }}
```
davidrapson, line 124: Cypress has automatic screenshot capturing (and video if we want it) so we can upload those on failure.

.github/workflows/ci-workflow.yml
```diff
@@ -52,12 +52,25 @@ jobs:
...
+      run:
+        working-directory: engine
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: '2.7.4'
+          working-directory: engine
       - run: bundle exec rubocop --format github
```
davidrapson, line 70: This is to a) allow us to run the demo app job earlier as it's doing more work now b) avoid running lint checks for very single version of ruby/rails. We only care about rspec tests for different versions. We'll need to add this job to the required checks after merging this.

marianayovcheva (approved):
Nice! 👏 Tested locally, worked well.

davidsauntson (approved):
This is great!  Works fine for me locally.

I'm guessing we won't use saucelabs after this is in?

davidrapson:
Great!

> I'm guessing we won't use saucelabs after this is in?

Yep we opted not to proceed with adding saucelabs a while back in https://github.com/citizensadvice/design-system/pull/1757

PR #1855, opened 2022-03-22
Remove pa11y

davidrapson:
As of https://github.com/citizensadvice/design-system/pull/1841 we now run accessibility smoke tests using cypress-axe as part of our Cypress tests. We can remove pa11y to avoid duplication.

TODO:

- [ ] Remove `pa11y` as a required status check


davidrapson:
Merged into https://github.com/citizensadvice/design-system/pull/1870

PR #1862, opened 2022-03-22
Add ADRs for recent testing decisions

davidrapson:
We have made a number of architecturally significant decisions around how and when we test our components. We should document these decisions.

Using this as an excuse to set up a basic ADR structure using adr-tools. As we manage our work through github projects storing our ADRs alongside the code makes most sense.

I've also include an architecture overview as suggested by @marianayovcheva to hopefully make it a little clearer what the boundaries are between parts of the design system.

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+contributing/adr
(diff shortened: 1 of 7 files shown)


marianayovcheva (approved):
Nice and useful. Makes sense to me, thanks!

Issue #1418, opened 2021-08-11
Rationalise colour var names


davidsauntson:
Currently the colour variable names are a bit all over the place and the design system colour palette includes product specific colours for advisernet.  We need to:

 - use a consistent approach to naming colour language variables so they can be more easily understood and overridden by products (suggest smth along the lines of `${component}__{element}__{property}--{state}`)
   - eg `$cads-header__link-hover-colour` should be smth like `$cads-header__link__color--hover` 
- remove advisernet product colours from palette and add into product specific css

PR #1863, opened 2022-03-23
Tech debt/rationalise colour names

marianayovcheva:
Originated from Issue #1418 

- Renamed the colours using the following structure: `${component}__{element}__{property}--{state}`
- Added a simple cads-colours functions that allows for legacy compatibility, useful for cases where colours of the design-system components are overriden by the product. Eg: [advisernet colour-language.scss](https://github.com/citizensadvice/public-website/blob/develop/app/javascript/stylesheets/intranet/settings/_colour-language.scss). Similar approach is used by [govuk](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_colours-applied.scss#L33) (Thanks to David R for the reference)
- Kept the header and footer colour variables but moved them directly in the components as they are local variables, not global. Inspired by [govuk](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/header/_index.scss)
- Updated the Storybook colour section with the new colour names
- Tested the colour updates with the current state of the public-website (linked my local repo in package.json) - things looked good and VRT didn't return errors. 
I wanted to get your opinion on the colour names first and then proceed with the following:

- [x] Update the Changelog
- [x] Add a file that contains the old colour names mapped to the new ones (this will ensure we avoid the breaking change)
- [ ] Add information for the users about how to rename the colours using the new naming conventions


davidrapson:
Couple of thoughts here:

**Backwards compatibility:**

This is likely a pretty big breaking change as projects use variable names directly. It's worth doing but we might want to look at a backwards compatible way of changing things, so we can still keep aliases of the old names around.

**Naming conventions:**

I wonder if we should use a simpler naming convention here rather than a BEM inspired one. From the Sass docs:

> 💡 Fun fact:
> Sass variables, like all Sass identifiers, treat hyphens and underscores as identical. This means that $font-size and  $font_size both refer to the same variable. This is a historical holdover from the very early days of Sass, when it only allowed > underscores in identifier names. Once Sass added support for hyphens to match CSS’s syntax, the two were made ? equivalent to make migration easier.

— https://sass-lang.com/documentation/variables

This means that the following are all equivalent:

```
$cads-language__background-colour--hover
$cads_language--background-colour__hover
$cads_language_-background-colour-_hover
$cads_language__background-colour__hover
$cads-language--background-colour--hover
```

Based on this I think a simple hyphen based naming convention would make sense e.g. `$cads-language-background-colour-hover`. Would minimise the number of possible variations.

davidrapson:
Just realised this was still draft 🙈 carry on.

scss/1-settings/_colour-language.scss
```diff
@@ -1,62 +1,29 @@
 @use '../1-settings/colour-palette' as palette;
+// ${component}__{element}__{property}--{state}
+$cads-language__brand--primary: palette.$cads-palette__heritage-blue !default;
+$cads-language__brand--primary-dark: palette.$cads-palette__heritage-blue-dark !default;
+$cads-language__brand--secondary: palette.$cads-palette__heritage-yellow !default;
+$cads-language__brand--primary-contrast: palette.$cads-palette__white !default;
+$cads-language__brand--adviser: palette.$cads-palette__blue-adviser !default;
+$cads-language__brand--focus: palette.$cads-palette__yellow !default;
```
marianayovcheva, line 8: I'm not sure I want to call it `brand--focus` because I don't know if fits `brand` naming. However, naming the colour just `$cads-language__focus` also doesn't feel quite right as focus is not an element. Ideas are welcome

styleguide/foundations/colours.stories.mdx
```diff
@@ -69,16 +69,18 @@ Use the ‘Input border’ for interactive elements like input boxes so it meets
...
 </Description>
 
 ### Adviser
 
 Only use ‘Adviser’ colours for adviser-related components, like ‘Adviser callout’. This is also the brand colour for AdviserNet.
 
-<Description>{colourLanguageTable([['Adviser', 'brand-adviser']])}</Description>
+<Description>
+  {colourLanguageTable([['Adviser', 'brand--adviser']])}
+</Description>
```
marianayovcheva, line 84: I'm tempted to add `$cads-language__error__colour` and `$cads-language__success__colour` in the documentation but I'm not sure if this needs to be agreed with designers first.

davidrapson:
> naming conventions - I see your point, tbh I wouldn't normally bemify variable names. In my opinion thought this can be a good case for using BEM as it makes the variables easier to read and follow. It also makes it easier to keep colours consistent in the future - if a dev needs to add a new colour, they would need to follow the naming convention. Happy to have a further chat

Yeah, we also technically use a BEM-ish style now. So maybe the difference I mentioned isn't a big deal? I think as long as we have a consistent pattern that's the most important thing. Another route might be to do something with maps (see https://github.com/biglotteryfund/blf-alpha/blob/master/assets/sass/_vars.scss#L70-L104) and write a helper function to access colours (https://github.com/biglotteryfund/blf-alpha/blob/master/assets/sass/_functions.scss#L12-L14) 

Using a wrapping function might help with backwards compatibility as this is how GDS do things too https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_colours-applied.scss#L33

scss/6-components/_footer.scss
```diff
@@ -75,7 +75,7 @@
   .cads-footer__company-info {
     @include cads-restrict-content-width();
 
-    color: $cads-footer__text-colour;
+    color: $cads-language__text__colour;
```
davidrapson, line 78: This probably _does_ need to be specific to the footer. By using the shared `$cads-language__text__colour` I'm not sure how we'd handle themeing like with AdviserNet/Intranet as this colour needs to be reversed out in those cases.

marianayovcheva: I see, good shout! I think if we keep them, we can perhaps put the header and footer vars into different files - that was the bit that made me remove them in first place

marianayovcheva:
@davidrapson those links are super useful, thank you! I'll have a further think about how to structure the colours, I'm beginning to lean towards the blf approach as it seems quite clean to me. I'll also think about how to simplify the header/footer colours without wiping the theming functionality. For now I'll move the PR back to draft but will let you know when I reopen it.

scss/1-settings/_colour-language.scss
```diff
@@ -1,62 +1,110 @@
...
-$cads-footer__text-secondary-colour: $cads-language__brand-primary !default;
-$cads-footer__link-colour: $cads-language__link-colour !default;
-$cads-footer__link-hover-colour: $cads-language__link-hover-colour !default;
-$cads-footer__hover-background-colour: $cads-language__hover-background-colour !default;
-$cads-footer__link-visited-colour: $cads-language__link-visited-colour !default;
-$cads-footer__link-active-colour: $cads-language__link-active-colour !default;
-$cads-footer__link-focus-colour: palette.$cads-palette__black !default;
-$cads-footer__link-focus-background-colour: $cads-language__focus-colour !default;
-$cads-footer__link-focus-border-colour: palette.$cads-palette__black !default;
+@import './legacy-colours';
```
marianayovcheva, line 1: Very tempted to take the function into its own file but not sure if it makes more sense for it to be here

marianayovcheva:
Reopening this PR - changed my approach and edited the description to reflect. Happy to have a discussion about it once we are all back from holidays etc. but also feel free to post any immediate thoughts and I will have a look.

scss/4-elements/_buttons.scss
```diff
@@ -111,20 +113,20 @@ $button-shadow-size: 4px;
...
+  color: $cads-language__button__colour--primary;
   background-color: $cads-palette__white;
-  border: solid $cads-border-width-small $cads-language__primary-button-colour;
+  border: solid $cads-border-width-small $cads-language__button__colour--primary;
   box-shadow: none;
 
   &:focus {
     box-shadow: 0 ($button-shadow-size - $cads-border-width-small) 0 0
-      $cads-language__focus-border-colour;
+      $cads-language__text__colour--secondary;
```
davidsauntson, line 123: Not sure about using a variable called `$cads-language__text__colour--secondary` to colour a box shadow - what if we want to theme them differently? Will the box shadow text colour always be the text colour or is it coincidence?

scss/6-components/_footer.scss
```diff
@@ -1,6 +1,43 @@
 // ============================================================================
 // Footer
 // ============================================================================
+// Colours
+$cads-footer__divider__colour: cads-colours(
```
davidsauntson, line 5: How does this get overridden when we theme for advisernet or the intranet?

davidsauntson (commented):
Loving the function for letting consumers swap out variable names really easily.  Agree we need better consistency in the variable names too.

A couple of thoughts on the new variable names:
- we should use this opportunity to drop or replace `language` - it's never felt particularly intuitive to me
- this one might be just me, but if we're using the property in the variable name, can we use `color` instead of `colour` because `color` is the property (eg `$cads-language__text__colour` &rarr; `$cads-language__text__color`) and I find it confusing
- using BEM results in some interesting scenarios, I wonder if it is the best method to use for colour vars:
  - `$cads-language__button__colour--primary-hover` - should this be `$cads-language__button--primary__color--hover`?
  - `$cads-language__brand--primary-dark` these feel like generic settings and this isn't really using BEM or even the format decided on in the comments
 
Some other comments / thoughts inline too

marianayovcheva:
UPDATE
Based on our discussion I've:
- replaced `$cads-language` with `$cads-color`
- decided it makes more sense to unbemify the colours. The fact that we have `$cads-color` as a base of the var name made the rest of the variable cleaner
- added `!default` to the header and footer colours

If you are happy with the colour names, I'll also update them in the new `design-system-docs` colour page. I'm also planning to add the red warning/important colour and the green success colour to the colour page as well, based on our post recent monthly meeting as designers mentioned it would be useful.

We also discussed ubemifying the `$cads-palette`. I had a check in public website and we do use `$cads-palette` vars. I'll wait to see what is the outcome of the current PR before starting on working on unbemifying the palette

marianayovcheva:
Closing this as the PR is very stale (pre Storybook removal). I'll raise a new PR with the updated files

PR #1870, opened 2022-03-24
Migrate remaining feature tests to Cypress

davidrapson:
Did a bit of an audit of the outstanding tests that needed to be migrated over to Cypress to allow us to remove the previous selenium testing support code and it turns out it's a small list.

- Migrates a viewport based breadcrumb test to Cypress
- Migrates one interaction based test from the header feature to check that search can be toggled

Everything else I _think_ is either:

- Already covered in the existing Cypress tests
- Already covered by faster view component specs
- And/or are better covered by visual regression tests

These assumptions need checking though so please shout if it looks like I'm removing something that isn't covered elsewhere.

For the forms tests I think I'd like to update the demo app to have a full working form that we could write a Cypress test against. The previous feature scenarios are _technically_ interactive but as they only type into a field but don't do anything else I've not migrated them over.

Also merged https://github.com/citizensadvice/design-system/pull/1855 into this pull request as they touch related areas.


davidsauntson (approved):
🚮

PR #1874, opened 2022-03-28
Experimental: New docs component examples

davidrapson:
Opening for visibility

PR #1876, opened 2022-03-28
Footer: Simplify how year is calculated

davidrapson:

Use simpler `Time.current.year` rather than `Time.zone.today.year` for footer year. Allows component to be more portable and usable outside of a Rails context.


marianayovcheva (approved):
⌛

Issue #1761, opened 2022-02-04
Internal links treated as external when using the design system locally


marianayovcheva:
## Summary of problem
When using the design system on local sites (e.g on public-website), internal links get external links style. This happens because the link doesn't start with `https://www.citizensadvice.org.uk` as specified here: https://github.com/citizensadvice/design-system/blob/main/scss/1-settings/_prose.scss. 

This happens on pages that use .cads-prose (e.g: http://localhost:3000/immigration/bringing-family-to-live-in-the-uk-after-brexit/).

## Expected behaviour
First link is internal and has no external icon
![image](https://user-images.githubusercontent.com/94383569/152568575-aa7c0d46-9d36-49cb-b801-e485daad7ec8.png)

## Actual behaviour
First link is internal but gets external icon
![image](https://user-images.githubusercontent.com/94383569/152570714-96d3052a-ffc0-4548-8722-ff0d118566ca.png)

## Proposed solution
Most of the internal urls actually begin with `/` and not `http` a possible solution to the problem could be:
```
  a:not([href^="https://www.citizensadvice.org.uk"]):not([href^="/"]):not([href^="#"])::after
  {
    @media not print {
      @include cads-hyperlink--external-icon();
    }
  }
```
- `[href^="https://www.citizensadvice.org.uk"]` - in case there are urls that are internal but begin with the full url
- `[href^="/"]` - most of the internal links 
- `[href^="#"]` - possible hashes

davidsauntson:
This bug presumably also causes internal links on non-prod domains to be rendered with an external link icon too (if they aren't relative links).

Another option would be to change the current `$cads-internal-link-domain` SASS variable to a [list](https://sass-lang.com/documentation/values/lists), and create a rule based on each value in the map.  This would allow users of the design system to have more control

PR #1877, opened 2022-03-28
Fix external link styles

davidrapson:
Applies @marianayovcheva's fix for external links when used outside of the production domain.

Fixes https://github.com/citizensadvice/design-system/issues/1761

Feeling like we should definitely open up the conversation about removing these icons in the long term but let's at least get them working reliably again after my previous half-fix 🙈

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Bugfixes**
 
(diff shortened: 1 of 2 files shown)


davidrapson:
> I'm approving this one. I think David S also had a suggestion on that issue (about having the domains in a list to allow more flexibility) which I think would be interesting to try in the future

Also noticed some useful feedback about them in the latest referrals accessibility report that definitely tie into this.

PR #1878, opened 2022-03-28
Migrate backstop tests to demo app

davidrapson:
Migrates backstop tests over to the demo app. This is the last step needed to allow us to remove the previous `testing/` directory.

@davidsauntson @marianayovcheva mind testing this by:

1. Run `./bin/setup` from with the demo directory
1. Starting the demo Rails app (you might need to do the normal dance of running `./bin/webpack-dev-server` if running in development)
2. Running `npm run backstop` from within the `demo` directory


demo/backstop-common.js
```diff
@@ -269,168 +266,156 @@ module.exports = function backstopCommon(baseUrl) {
...
-      html_report: 'backstop_data/html_report',
-      ci_report: 'backstop_data/ci_report',
+      engine_scripts: 'visual-regression/engine_scripts',
+      bitmaps_reference: 'visual-regression/backstop_data/bitmaps_reference',
+      bitmaps_test: 'visual-regression/backstop_data/bitmaps_test',
+      html_report: 'visual-regression/backstop_data/html_report',
+      ci_report: 'visual-regression/backstop_data/ci_report',
     },
-    report: ['browser'],
+    report: ['html'],
```
marianayovcheva, line 418: I liked it backstop would open the browser right after the tests were complete. I do realise that's more of a personal preference thing, so I'm happy to run `npm run backstop:reference` instead

marianayovcheva (approved):
Nice! I've not checked every single file in the pr but I've had an overall look at them and more importantly I've tried to run it locally (which worked). Happy to approve, I think it would be useful if David S also reviews the set up

demo/backstop-common.js
davidrapson: Oh yeah! I think I needed to tweak this to run against the demo app, but let me see if I can't maintain that behaviour locally.

davidsauntson (commented):
Nice one!

Config etc looks fine to me.

Interestingly, I get a failure in the advisernet targeted content test:
image

I thought it might be around the `dockerman/backstopjs 5.4.4` M1 image I'm using but actually localhost does render slightly differently from the reference images that are checked in - this is 800px wide and the column title doesn't wrap
image

PR #1660, opened 2021-11-25
Feat/de yarn design system docs

davidsauntson:
Replaces `yarn` with `npm` in the design system docs project.

PR #1880, opened 2022-03-28
Prepare docs build

davidrapson:

## Improve yarn support

In https://github.com/citizensadvice/design-system/pull/1778 we introduced the dependency on yarn as bridgetown makes some assumptions about this being available.

Tidy things up by:

- Removing the old package-lock.json file (if we're using yarn we should only have the `yarn.lock` file)
- Updating internal rake tasks to use yarn
- Adding a bin/setup script with a yarn install step

## Update docs workflow

Allows for building full docs site including assets within the new CI workflow. Uses the same general approach as for the demo app.

## Basic docs homepage

Adds a minimal home page and navigation structure.

Includes a fix to the builder to make sure local files are loaded so we can use the footer component.

image

### DIFF
--- a/design-system-docs/README.md
+++ b/design-system-docs/README.md
@@ -1,52 +1,29 @@
-# Bridgetown Website README
+# Citizens Advice Design System documentation website
 
-Welcome to your new Bridgetown website! You can update this README file to provide additional context and setup information for yourself or other contributors.
-
-## Table of Contents
-
-- [Prerequisites](#prerequisites)
-- [Install](#install)
-- [Development](#development)
-- [Contributing](#contributing)
+This is the documentation website for the Citizens Advice Design System.
 
 ## Prerequisites
 
-- [GCC](https://gcc.gnu.org/install/)
-- [Make](https://www.gnu.org/software/make/)
-- [Ruby](https://www.ruby-lang.org/en/downloads/)
-  - `>= 2.7`
-- [Bridgetown Gem](https://rubygems.org/gems/bridgetown)
-  - `gem install bridgetown -N`
-- [Node](https://nodejs.org)
-  - `>= 12`
-- [npm](https://npmjs.com)
+- [Ruby 2.7](https://www.ruby-lang.org/en/downloads/)
+- [Node 14](https://nodejs.org)
+- [yarn](https://yarnpkg.com/getting-started/install)
 
-## Install
+## Setup
+
+All commands should be run within this directory, `design-system-docs/`, from within your terminal of choice.
+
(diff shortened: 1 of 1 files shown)


design-system-docs/src/_layouts/default.erb
```diff
@@ -1,22 +1,26 @@
 <!doctype html>
 <html lang="<%= site.locale %>">
   <head>
-    <title>Design system docs</title>
+    <meta charset="utf-8" />
+    <%= seo %>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="robots" content="noindex, nofollow">
     <link rel="stylesheet" href="<%= webpack_path "css" %>" />
+    <style><%= Rouge::Theme.find('github').render(scope: '.highlight') %></style>
```
marianayovcheva, line 9: Curiosity question - I'm not really sure what the highlight styles do?

davidrapson: Good question. It is syntax highlighting for code samples. Bridgetown includes this library https://github.com/rouge-ruby/rouge which has some built in themes. Picked one to get us started, but there's loads available or you can write your own.

marianayovcheva: I see, thanks!

marianayovcheva (approved):
Nice! I've had a look, and tried it on my machine and it worked nicely. I think it may be worth for David S to also have a look as I think he's the one who initially set up Bridgetown

davidsauntson (commented):
Sooo, I'm a bit confused as I deliberately de-yarned at your request (https://github.com/citizensadvice/design-system/pull/1660)

Although do the bridgetown bin commands require us to use yarn?

davidrapson:
> Sooo, I'm a bit confused as I deliberately de-yarned at your request (https://github.com/citizensadvice/design-system/pull/1660)

Yeah I was confused myself, but I checked and most of the `bridgetown configure` commands assume the existence of yarn. Rails is a bit like this too. Also noticed that you added a yarn.lock file https://github.com/citizensadvice/design-system/blob/main/design-system-docs/yarn.lock and used yarn in the build https://github.com/citizensadvice/design-system/blob/main/.github/workflows/docs-workflow.yml#L31 so I thought it worth following through.

Given how much the ruby framework world seems to prefer yarn it's probably for the best to stick with it.

davidsauntson:
Yeah that makes sense - just wanted to check.  I guess the yarn files came from running build etc and I didn't even notice at the time! 👁️

davidsauntson (approved):

PR #1881, opened 2022-03-29
Footer: Support new feedback_link slot

davidrapson:
## Background

We are exploring some new feedback routes for intranet content. Currently the feedback link in the footer has a) global text b) always assumes an external link and c) always opens in a new tab.

### Add more flexible `feedback_link` slot and deprecate `feedback_url` argument.

In order to allow more flexibility in configuring the footer feedback link I've added a new `feedback_link` slot used like this:

```rb
render CitizensAdviceComponents::Footer.new do |c|
  c.columns(columns)
  c.feedback_link(url: "https://example.com/", title: "Example text", external: true)
end
```
Maintains backwards support for the previous `feedback_url` and makes use of `ActiveSupport::Deprecation` to log a deprecation warning to tell people to upgrade.

### Fix accessibility issue with external link icons using old icon font and reading out content. Use new SVG icons.

This was flagged by a recent referrals accessibility audit. The footer was using the old icon fonts still which are read out to some screen readers, so this would read "let us know e" where "e" is the character the icon is mapped to. Swap this out with the equivalent SVG icon.

### Support `external` icon for column links and deprecate `icon` property

Previous option was too generic and supported any icon within the icon font. Limit this to just external links and swap to an SVG icon, fixing the same accessibility issue described above.


styleguide/components/footer/footer-docs.mdx
```diff
@@ -19,9 +19,8 @@ import { Meta, Story, Canvas } from '@storybook/addon-docs';
 If you are using the `citizens_advice_components` gem, you can call the component from within a template using:
 
 ```rb
-render CitizensAdviceComponents::Footer.new(
-  feedback_url: "https://www.research.net/r/J8PLH2H"
-) do |c|
+render(CitizensAdviceComponents::Footer.new) do |c|
+  c.feedback_url(url: "https://example.com", external: true)
```
marianayovcheva, line 23: Should this be `feedback_link`?

PR #1885, opened 2022-03-31
Add support for component examples

davidrapson:

Adds a basic component template and index page along with a ComponentExample view component which allows rendering a component example along with source code and generated HTML.

Opted not to extract the arguments table for now as I think that needs a little bit more thought.

localhost_4000_components_targeted-content_

### DIFF
--- a/design-system-docs/src/_component_docs/targeted-content.md
+++ b/design-system-docs/src/_component_docs/targeted-content.md
@@ -1,0 +1,62 @@
+---
+layout: component
+title: Targeted Content
+---
+
+The targeted content component allows users to show and hide sections of related content on a page. It can help users scan and process content on a page more effectively.
+
+However, it does hide content from users and it can be easy for users to miss. When using it, you should aim to use **no more than 6** and you shouldn’t have a targeted content on its own.
+
+Because targeted contents hide content, they should always have a clear label/title that helps users understand what is in the section. You can use `h3` within them to break up the content.
+
+## When to use
+
+When content is only relevant to some particular users visiting the page.
+
+## When not to use
+
+They shouldn’t be used to make a page seem shorter, for example: using them to hide large chunks of content.
+If you find yourself using targeted content to do this, think about:
+
+- If you’re repeating content on the page and could consolidate it.
+- Using headings on the page instead.
+- If you need to move some of the content onto another page.
+
+## Examples
+
+### Default variant
+
+<%= render(Shared::ComponentExample.new("targeted_content_default")) %>
+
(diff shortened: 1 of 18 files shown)


design-system-docs/src/_components/shared/component_example.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def iframe?
+      @iframe.present?
+    end
+
+    def highlighted_source
+      # @FIXME: There's probably some helper method to do this for us
+      lexer = Rouge::Lexers::ERB.new
+      formatter.format(lexer.lex(example.untransformed_content))
```
marianayovcheva, line 41: Learning question - is `untransformed_content` something that comes directly from the Bridgetown gem? I've tried looking for it but could only find this: https://www.rubydoc.info/gems/bridgetown-core/Bridgetown/Resource/Base#untransformed_content-instance_method

design-system-docs/src/_component_docs/targeted-content.md
```diff
@@ -0,0 +1,62 @@
...
+### Fallback behaviour
+
+If JavaScript either fails or is disabled in the users browser the fallback is to show the content as open. Disable JavaScript for this page to see the fallback behaviour in action.
+
+## Using with Rails
+
+If you are using the `citizens_advice_components` gem you call the component from within a template using:
+
+```rb
+<%%= render(CitizensAdviceComponents::TargetedContent.new(
```
marianayovcheva, line 55: Learning question - I'm used to seeing `%` but I'm unsure what `%%` does?

marianayovcheva (approved):
I'm not a Rubyist but I've had a look through the code and things mostly make sense to me. I've added a couple of questions that are for my personal learning. 

Nice and exciting to see it progressing! 👏

design-system-docs/src/_components/shared/component_example.rb
davidrapson: Yep that's the exact method! Each resource in bridgetown terms has a few methods we can call. This one as the name suggests is for the untransformed content i.e. the source code for the example.

design-system-docs/src/_component_docs/targeted-content.md
davidrapson: In this context it escapes the code sample, otherwise the code will get executed as we're inside a file that understands erb. From the docs here https://www.bridgetownrb.com/docs/template-engines/erb-and-beyond

> If you need to escape an ERB tag (to use it in a code sample for example), use two percent signs:

PR #1907, opened 2022-04-06
Proposal: Remove external link icon

davidrapson:
This is speculative and needs discussion with the working group, but raising this as a starter for ten before I suggest it through the contribution process, no expectation of this being merged.

---

Following the precedent set by: https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/

We've also not evidence that external link icons improve the user experience but we have seen cases where they confuse things.

We have examples within products of the icon being used to mean simple external link and to mean opens in a new tab making the icon ambiguous.

We also have a related issue with our current implementation which uses an icon-font for the icon which gets read out to screen readers.

Rationale is that as we know the icon is ambiguous **and** we have a current accessibility issue with the implementation, rather than investing time in fixing it we could explore removing the icon completely.


davidrapson:
Closing this but going to write up an actual proposal for it and circulate to the working group.

PR #1908, opened 2022-04-06
Remove changelog scripts

davidrapson:
For better or worse I don't think we've ever run these and have got on reasonably well hand-writing the changelog. Given this and the fact there's quite a lot of supporting code and dependencies for it how do we feel about removing it?


marianayovcheva:
🙋‍♀️ Question - I've never done a release but I thought that the version information and date were automatically added to the changelog. Is it all manually written?

marianayovcheva (approved):

PR #1924, opened 2022-04-13
fix: closing greedy nav on tab out if there are no header links

marianayovcheva:
Fixes: Greedy Nav doesn't close when tabbing out if no header links are present [NP-1755](https://citizensadvice.atlassian.net/jira/software/c/projects/NP/boards/208?modal=detail&selectedIssue=NP-1755&search=greedy)

### DIFF
--- a/src/ts/greedy-nav/GreedyNav.ts
+++ b/src/ts/greedy-nav/GreedyNav.ts
@@ -604,20 +604,23 @@ export class GreedyNavMenu {
       navDropdownToggle.addEventListener(
         blurEventName,
         (e: FocusEvent): void => {
-          let lastItem: Nullable<HTMLElement> | undefined = null;
-          const headerLinksInNav: Nullable<HTMLElement> =
-            document.querySelector(
-              `${this.navDropdownSelector} .js-cads-copy-into-nav`
-            );
+          let lastItem: HTMLElement | null | undefined;
+          const headerLinksInNav: HTMLElement | null = document.querySelector(
+            `${this.navDropdownSelector} .js-cads-copy-into-nav`
+          );
 
-          if (headerLinksInNav?.offsetParent !== null) {
-            lastItem = headerLinksInNav?.querySelector(
+          if (headerLinksInNav?.offsetParent) {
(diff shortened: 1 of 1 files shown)


src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -604,20 +604,23 @@ export class GreedyNavMenu {
...
-          const headerLinksInNav: Nullable<HTMLElement> =
-            document.querySelector(
-              `${this.navDropdownSelector} .js-cads-copy-into-nav`
-            );
+          let lastItem: HTMLElement | null | undefined;
+          const headerLinksInNav: HTMLElement | null = document.querySelector(
+            `${this.navDropdownSelector} .js-cads-copy-into-nav`
+          );
 
-          if (headerLinksInNav?.offsetParent !== null) {
```
marianayovcheva, line 613: This was the initial reason for the issue - the case in which there were no headerLinksInNav would also fall into this if statement as headerLinksInNav?.offsetParent was returning undefined. So I've added an extra condition to cover the possible scenarios

davidsauntson (approved):
Great! 

It would be great if you could add a feature test for this - ATM it is a skipped `it` in the cypress tests 

https://github.com/citizensadvice/design-system/blob/f033ab0430b50acd7f394d8fc93e551f117c3be5/demo/cypress/integration/navigation.spec.js#L20

marianayovcheva:
I have tried to add testing scripts as follows:
``` 
 it('tabbing out of the dropdown menu automatically closes it', ()=>{
    tabIntoNavigation();
    assertNavigationOpen();
    cy.get('.cads-greedy-nav__dropdown').findByText('Sign in').focus().tab();
    assertNavigationClosed();
  });
```
When tabbing in a browser the dropdown closes when the last item is reached, however in Cypress the dropdown doesn't close as expected. This may be related to the fact that [Cypress doesn't support tabbing](https://docs.cypress.io/api/commands/type#Typing-tab-key-does-not-work) and we are testing it using an experimental plugin.

Issue #1902, opened 2022-04-06
Icon fonts elements are announced to screen readers


davidrapson:
## Summary of problem

A recent accessibility report highlighted that places where we still use the old icon font can cause issues with screen readers

## Steps to replicate

This happens wherever there is an icon font element without `aria-hidden="true"` or used as a background image but a particularly egregious example is the search toggle.

1. Load the sample content page at a small viewport size
2. Use voice over and toggle the search open and closed
3. It will announce "K" or "H" before then reading the correct descriptive label. 

## Expected behaviour

No additional text announced

## Actual behaviour

The letter that the icon font element is mapped to is announced.

PR #1925, opened 2022-04-14
Asset hyperlink icons work with screen readers

tahb:
We were migrating a corporate page[1] and when we ran an accessibility check we found that the icons used in asset download links don't read meaningfully to screenreaders[2]. And would be a possible fail at level A.

[1] https://cdn.develop.content.citizensadvice.org.uk/about-us/our-work/our-prevention-work/BESN
[2] https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 **New**
+
+- 🔗 Links: Document links no longer present their appended file icon to screen
+  readers
 - 🔗 Links: Relflow long links so they do not add scrollbars to the viewport
 - 🔓 Disclosure: Adds new `Disclosure` view component
 - 🥾 Footer: Add more flexible `feedback_link` slot and deprecate `feedback_url` argument.
--- a/engine/app/components/citizens_advice_components/asset_hyperlink.html.haml
+++ b/engine/app/components/citizens_advice_components/asset_hyperlink.html.haml
@@ -1,5 +1,5 @@
 %a{ href: href }
   = description
   %span.cads-asset-type<
-    %span.cads-icon_file
+    %span.cads-icon_file{ "aria-hidden": "true" }
     \ #{number_to_human_size(size)}
--- a/styleguide/examples/asset_hyperlink/example.html
+++ b/styleguide/examples/asset_hyperlink/example.html
@@ -1,6 +1,6 @@
 <a href="https://example.com">
   Test PDF
   <span class="cads-asset-type"
-    ><span class="cads-icon_file"></span> 6.15 MB</span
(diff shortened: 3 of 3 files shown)


tahb:
I have no idea why CI fails with `Error: Input required and not supplied: netlify-auth-token`. Help please! 😬

davidsauntson:
> I have no idea why CI fails with `Error: Input required and not supplied: netlify-auth-token`. Help please! 😬

Ooooh, interesting. We're using Netlify to create a preview of the docs site when a qualifying PR is changed.  You're our first external contributor (❤️ ) so we haven't tested this scenario yet.  You're getting the error because you don't have access to the secrets needed by the workflow to deploy the preview version.  I'll see if I can update the workflow so that it only triggers if the PR comes from a non-contributor and let you know.

davidrapson (approved):
Made a suggestion to use `aria-hidden="true"` but can go either way on it as the explicit label is definitely better than something misleading 👍 

We've still got a few places where we use this old icon-font style so we'll hopefully come back to this soon and swap things out for an SVG version (see https://github.com/citizensadvice/design-system/issues/1902)

davidrapson:
Not sure why the other jobs don't run either. We only have the workflow set to trigger on `push` so I wonder if that limits things. Going to make a commit to update the changelog to see if that triggers things.

davidrapson:
@tahb Would you mind updating your branch? Just made a change to the CI workflow so that it should hopefully trigger the required status checks now.

tahb:
Looks like I do in fact have a test to fix 🙌

davidrapson:
Ah yeah, it's complaining that `aria-label` isn't appropriate for spans. I think in that case let's go with `aria-hidden="true"` rather than a named label for the icon. It's ultimately decorative so that feels right.

davidrapson:
Unfortunately axe is being picky and needs "true" to be quoted (it's expecting `aria-hidden=true` in the markup but the boolean just outputs `aria-hidden`). Mind tweaking that, then we should be green and I can merge and do a release.

tahb:
Oh!? That's fair enough, I'll change it over. I can't see that feedback.

davidrapson:
I was going by https://github.com/citizensadvice/design-system/runs/6113336693?check_suite_focus=true#step:10:56 and realised I'd seen that error before so had a hunch what it was asking for.

davidrapson:
Oh that's interesting, hopefully that problem goes away once we merge https://github.com/citizensadvice/design-system/pull/1934. It's the main CI workflow statuses we want.

PR #1931, opened 2022-04-19
Run CI on pull_request event

davidrapson:
Pull requests from forks like https://github.com/citizensadvice/design-system/pull/1925 are not running any jobs. This is probably because we only trigger on the push event. It's likely we should be running on the `pull_request` event too.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -1,6 +1,7 @@
 name: Continuous Integration
 on:
   push:
+  pull_request:
   schedule:
     - cron: '0 2 * * *'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -1,6 +1,7 @@
 name: Continuous Integration
 on:
   push:
```
davidrapson, line 3: If we care about the double-jobs for PRs we could maybe do something like https://github.com/primer/view_components/blob/main/.github/workflows/ci.yml#L7-L9 to limit the push events to the main branch.

davidrapson: Ended up doing this

davidsauntson (approved):

PR #1932, opened 2022-04-19
(feature) Reflow long strings (accessibility)

mec:
During work to migrate this page: 

https://www.citizensadvice.org.uk/about-us/our-work/our-prevention-work/financial-skills-for-life/financial-skills-for-life-what-we-do/financial-skills-for-life-past-projects/save-xmas/

It was noted in the accessibility audit that the `mailto` link was potentially causing reflow issues for users with zooming enabled. This often happens when a url or email address is shown as content rather than the `href` target.

Whilst this particular link does not appear to be longer than 320px, we want to be sure about any others

## Background reading

Read the W3 docs: https://www.w3.org/WAI/WCAG21/Understanding/reflow
Read the W3 recommended approach: https://www.w3.org/WAI/WCAG21/Techniques/css/C33
Read more about the `overflow-wrap` attribute: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

## Screenshot of an example where this would be an issue

164042652-50636597-85b5-488c-936c-2e4a1f34b4df

## Screenshot of an example with the fix applied

164043246-6a1afedc-f87d-46df-b09b-cfb15cd1de7b

## Screenshot of testing the 'save xmas' mailto link

Screenshot 2022-04-19 at 10 34 05


contributing/testing.md
```diff
@@ -18,6 +32,28 @@ You can install backstop and set up the demo app by running the following from w
 ./bin/setup
 ```
 
+#### Compile packs
+
+You will need to compile any changes to the assets for the test environment:
+
+```sh
+RAILS_ENV=test bin/webpacker:compile
```
tahb, line 40: I think this might need to be `… bin/rails webpacker…`.

Running it for me fails though which is frustrating. 

I first got:

```
ModuleNotFoundError: Module not found: Error: Can't resolve '@citizensadvice/design-system/lib/header' in '/Users/tomhipkin/sites/citizens-advice/tahb/design-system/demo/app/frontend/packs'
    at /Users//sites/citizens-advice/tahb/design-system/demo/node_modules/webpack/lib/Compilation.js:925:10
```

I also saw a warning `Warning: you are using an unstable release of Node.js (v17.7.2)` so tried to bump my node to 18 but got a new error:

```
Compiling...
Compilation failed:
node:internal/crypto/hash:67
  this[kHandle] = new _Hash(algorithm, xofLen);
                  ^
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:67:19)
```

tahb: `bin/webpack-dev-server` doesn't work either. I hate webpacker.

mec: The first error smells like you have not run `npm run build` from the root before going into `demo` and running `RAILS_ENV=test bin/webpacker:compile`. To be honest, I am not in love with how this works, but it's not ours so I decided not to wade in with opinions (too much!)

In terms of Node version, I treat it like ruby, the (docs say 14)[https://github.com/citizensadvice/design-system/blob/main/contributing/setup.md#prerequisites], so I set that up,

tahb: Everything passes now. Though I did have to change it to `RAILS_ENV=test bin/rails webpacker:compile`

Yeah that's right had to go back up and run npm build. Despite the `**note**` I was trying to earlier `compile` step since I knew I had never run it. I wonder if that would be read first?

"If you have not done so, run npm run build from the root and then come back to /demo" or similar?

tahb:
I've followed the steps from a fresh install and it eventually worked for me 👍

contributing/testing.md
mec: 🤦 I've just seen what you mean!! Yep `bin/rails webpacker:compile` sorry!

davidrapson (approved):
👍 Ran this change using `npm link` against the public website code and there were no surprising side-effects so this looks good to me.

Really appreciate you taking the time to update the setup guide too.

As we discussed if you want the change in quicker it's totally fine to patch the corporate website styles in the meantime.

For the CI build, the docs step should be fixed by https://github.com/citizensadvice/design-system/pull/1934 but we can ignore that for now as it's not a required status check. The only required check that's failing is the lint step, I think it just wants you to run prettier on the readme via `npx prettier --write demo/README.md`.

If you don't mind adding an item to the CHANGELOG too that'd be fab/

davidrapson:
Going to merge this as it's only the docs preview build failing which we know about and isn't a required check.

PR #1933, opened 2022-04-20
Avoid double adding previews

davidrapson:
Component previews are already added to preview_paths by the engine code. We occasionally see odd load errors for some component previews and I wonder if it's because we were adding the preview paths twice. Either way, it's redundant so let's remove it.


davidrapson:
Merged this as this config is definitely redundant.

PR #1934, opened 2022-04-20
Run docs build only when a label is added

davidrapson:
We currently run docs preview builds on all pull requests when they are opened or updated. This has a couple of trade-offs:

- Previews are generated for every PR which can add noise if you don't want one, or are not ready for one yet.
- Causes an error on pull requests from forks as the workflow doesn't have access to secrets

This pull request makes docs site previews opt-in by only running them when a label is added. Included the full context in an ADR added to this pull request.

### DIFF
--- a/.github/workflows/docs-workflow.yml
+++ b/.github/workflows/docs-workflow.yml
@@ -3,9 +3,12 @@ name: Design System Docs
 on:
   pull_request:
     types:
-      - opened # a PR is created
-      - synchronize # existing PR is updated
+      # Trigger when the pull request is labelled.
+      # Used to only deploy previews when the "docs preview" label is added.
       - labeled
+      # Trigger when an existing PR is updated.
+      # Used to make sure new previews are deployed for changes after the label is added.
+      - synchronize
   release:
     types:
       - created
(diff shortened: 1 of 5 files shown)


contributing/documentation-site.md
```diff
@@ -0,0 +1,9 @@
+# Documentation site
+
+We have a new documentation site under `design-system-docs`.
+
+This docs site is deployed via Netlify which allows us to access previews of work in progress changes. You can request a preview deploy for any pull request by adding the "docs preview" label.
```
davidsauntson, line 5: It's worth adding that external collaborators / forked repos will not be able to deploy a preview to Netlify.

.github/workflows/docs-workflow.yml
```diff
@@ -5,18 +5,19 @@ on:
     types:
       - opened # a PR is created
       - synchronize # existing PR is updated
```
davidsauntson, line 7: Are you not going to remove these two?

davidrapson: The synchronize event is still needed as it's what triggers a new preview if you make changes after adding a label (otherwise it's just a one time thing). You're right though, the opened event probably isn't needed anymore. Will remove that and add a comment to the synchronize event

davidsauntson (approved):

PR #1963, opened 2022-04-22
Compile assets in demo app test environment

davidrapson:
This should improve the new setup workflow a bit as it stops cypress tests from relying on a specific test asset build.

We don't really care about different environments for the demo app so we can stand to have more parity between them. Allows the test environment to automatically compile assets if it needs to and bundles them in the same directory, no need for a separate packs-test directory.

Simplifies some of the documentation now there should be fewer gotchas.

### DIFF
--- a/contributing/testing.md
+++ b/contributing/testing.md
@@ -1,99 +1,32 @@
 # Testing
 
-## JS and Typescript
+## Linting
 
-We use Jest:
+We run `eslint`, `stylelint`, and `prettier` as lint checks.
+
+From the project root run:
+
+```sh
+npm run lint
+```
+
+## Unit tests
+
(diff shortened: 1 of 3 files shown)


davidsauntson (approved):

davidrapson:
@mec This is based on one of the changes you suggested last week. If you have time would you mind doing a fresh clone of the design system then from this branch try running `./bin/setup` from the top-level and then `./bin/rails cypress:open` from the `demo` app directory. Hoping this removes at least a bit of setup faff.

davidrapson:
Meyric is off, given this is all working in CI and locally i'm going to merge this and see if we can do a sense check after the fact.

mec:
Sorry I missed this - looks good though 👍

PR #1964, opened 2022-04-22
Move component previews into the demo app

davidrapson:
We currently bundle component previews as part of the engine gem. With the recent change to use the previews for all our demo app tests this changes the function these previews serve. They are now really internal test cases and don't make sense to bundle as part of the gem.

Moving them into the demo app alongside the tests that use them make this boundary clearer and hopefully simplifies some of the engine config.


davidsauntson (approved):
Nice - all looks ok to me.

PR #1977, opened 2022-04-25
Don't consider mailto: or tel: links as external

davidrapson:
Small change to make sure that mailto: and tel: links aren't counted as external links. We don't typically recommend using either of these but we have some mailto: links in migrated corporate content and they currently get styled as external links.

Also adds a preview with an example of different types of links so we can run some explicit visual regression tests against things (at least whilst we still support external link icons https://github.com/citizensadvice/design-system/pull/1907)


scss/5-objects/_prose.scss
```diff
@@ -61,7 +61,7 @@
   // Allow external links to be determined based on $cads-internal-link-domain value
   // Favour using rel=external as a styling hook.
   // https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-external
-  a:not([href^="#{prose-config.$cads-internal-link-domain}"]):not([href^="/"]):not([href^="#"])::after {
+  a:not([href^="#{prose-config.$cads-internal-link-domain}"]):not([href^="/"]):not([href^="#"]):not([href^="mailto:"]):not([href^="tel:"])::after {
```
davidsauntson, line 64: chef's kiss

davidsauntson (approved):

PR #1988, opened 2022-04-27
Move button backstop scenarios down.

davidrapson:
We occasionally see an error with the button backstop scenarios where they error with a `ActionNotFound` error or otherwise timeout. Curious to test if this is specific to the button scenarios or if it's because they come first. Test this assumption by moving the button scenarios down a bit (greater than the `asyncCaptureLimit`)


davidrapson:
Validated that it is in fact a start up issue as some different tests failed. Trying a few things here: https://github.com/citizensadvice/design-system/pull/1989

PR #1992, opened 2022-04-29
Add basic foundations structure

davidrapson:
Extracts just the structural work for the foundations templates from work I've started on the icons page. Allows @marianayovcheva and I to use the same base to work from.

- Config
- Layout
- Index page
- Default frontmatter
- Minimal icons document


marianayovcheva (approved):
🙏

Issue #1975, opened 2022-04-25
Migrate foundations page: Breakpoints


davidrapson:
Migrate Breakpoints foundations page to the new documentation site: https://citizensadvice.github.io/design-system/?path=/docs/foundations-breakpoints--page

No special behaviour on this page. 

There is this line which I think used to apply but no longer applies for either storybook or the new documentation site so we should remove it when we migrate this:

> To see how any tokens or components respond to different screen sizes just resize this window and look out for the breakpoint indicator.

PR #1996, opened 2022-04-29
feat: adding breakpoints page

marianayovcheva:
Copying the breakpoints page from Storybook to Bridgetown. Removed the line about resizing the screen as noted in [the issue](https://github.com/citizensadvice/design-system/issues/1975)


davidrapson (approved):
Nice!

PR #1997, opened 2022-04-29
Add missing file icon component

davidrapson:
Noticed whilst migrating the icons page to the new docs site that we're missing the file icon from our SVG components.

Resurrected it via https://github.com/citizensadvice/design-system/blob/ce10696b4639a14129f0bde9e2fc5a97dd3f7431/assets/icons/06%20icon_file.svg?short_path=628ae58
and added it to our icon set.

Updates the asset hyperlink component to use the SVG version in place of the old icon font.

The icon is slightly larger because I updated it to use our standard 16px icon grid and use the standard `small` size defined by our icon component.

image


styleguide/examples/links/example_links.html
```diff
@@ -0,0 +1,13 @@
+<div class="cads-prose">
```
davidrapson, line 1: This got auto-formatted when running `./bin/generate-examples`. Left it in this commit as it's small.

marianayovcheva (approved):

PR #2011, opened 2022-05-03
feat: adding spacing & sizing page

marianayovcheva:
Adding Spacing & Sizing page. Using the existing cads table component. Moved and renamed already existing css from the styleguide to the docs (i.e `_spacing-block.scss` and `_sizing-border.scss`)


davidrapson (approved):
Approving this as it can definitely go in as is. But Mariana and I are also going to pair on seeing if we can extract the tables into shared components to keep them out of the markdown 🤞

davidrapson:
🚀

PR #2013, opened 2022-05-04
Add on this page to component templates

davidrapson:
Kramdown—the default markdown parser used by bridgetown—has support for auto-generated tables of contents but they are very limited in their output so can't be used to render our on this page component.

Thankfully we can write a simple component which accepts the generated markup for the page and passes it through nokogiri to extract the headings for our components.

- Shown by default but can be disabled by setting `show_on_this_page: false` in component front matter
- Extracts any h2s from the main page content excluding any used within component examples

image


design-system-docs/Gemfile
```diff
@@ -20,6 +20,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 gem "bridgetown", "~> 1.0"
 gem "htmlbeautifier"
+gem "nokogiri"
```
davidrapson, line 23: This is already available as part of actiontext, I think, but adding it here makes it clear we're using it as an explicit dependency.

design-system-docs/src/_component_docs/_defaults.yml
```diff
@@ -0,0 +1 @@
+show_on_this_page: true
```
davidrapson, line 1: This _defaults.yml file is super useful, it lets you set common front-matter for any files within this directory.

PR #2014, opened 2022-05-05
Migrate Grid foundations

### LINKED ISSUE #1973: Migrate foundations page: Grid
Migrate Grid foundations page to the new documentation site: https://citizensadvice.github.io/design-system/?path=/docs/foundations-grid--basic-grid

Code examples here are HTML only so maybe we need a way to disable ERB examples for specific examples.

davidrapson:
Migrates the grid foundations page to the new docs site. Closes #1973

Includes some notable changes to the component examples to allow us to decide whether to show html or erb source samples:

- Allow component examples to be grouped by category
- Allow option to show just HTML or just ERB source
- Update styles to improve boundary between example UI and the example itself
- Add a minimal component example layout

image

### DIFF
--- a/design-system-docs/frontend/styles/_foundations.scss
+++ b/design-system-docs/frontend/styles/_foundations.scss
@@ -1,6 +1,15 @@
 // ============================================================================
-// Icons
+// Foundations
 // ============================================================================
+
+// Grid examples
+
+.grid-example {
+  padding: $cads-spacing-4;
+  background-color: $cads-palette__light-grey;
+}
+
+// Icons
 
 .icon-grid {
   display: grid;
--- a/design-system-docs/src/_component_examples/_targeted_content/adviser.erb
+++ b/design-system-docs/src/_component_examples/_targeted_content/adviser.erb
@@ -1,10 +1,8 @@
 ---
-title: Default
 ---
-<%= render(CitizensAdviceComponents::TargetedContent.new(
-  title: "If you are a citizen of a country outside the EU, EEA or Switzerland",
-  id: "targeted-content-123")
-) do %>
+<%= render(CitizensAdviceComponents::TargetedContent.new(type: :adviser,
+  title: "Students or self-sufficient people",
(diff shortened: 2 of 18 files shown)


design-system-docs/frontend/styles/_foundations.scss
```diff
@@ -1,7 +1,16 @@
 // ============================================================================
-// Icons
+// Foundations
```
marianayovcheva, line 2: Do you think it's worth me moving the contents of `sizing-border.scss` and `spacing-block.scss` here as they are part of the foundation? (as part of a separate PR)

design-system-docs/src/_component_examples/_targeted_content/adviser.erb
```diff
@@ -1,5 +1,4 @@
 ---
-title: Adviser
```
marianayovcheva, line 2: I'm slightly confused about why these don't have titles and the grid examples do. Is it because the grid ones are using iframes?

marianayovcheva (approved):

design-system-docs/src/_component_examples/_targeted_content/adviser.erb
davidrapson: Oh good catch, I don't think the grid ones need titles anymore either.

design-system-docs/frontend/styles/_foundations.scss
davidrapson: Yeah possibly. I originally did what you did an had separate files, but it started to feel simpler to just have a foundations.scss file so it's clear(er) these styles are one-offs.

design-system-docs/src/_component_examples/_targeted_content/adviser.erb
davidrapson: Oh wait I worked it out. If you don't add a `title` it uses the slug as the title. I removed these because the titles were just repeating the slug. It's probably clearer though to always have them.

PR #2018, opened 2022-05-12
Update dependabot rules

davidrapson:
- Removes old top-level bundler rule
- Removes old testing rule
- Adds missing npm rule to demo app
- Adds rules fro design system docs site


davidsauntson (approved):
🤖

davidrapson:
Apologies in advance for the dependabot carnage this will probably cause.

PR #2060, opened 2022-05-23
chore: updating codeowners

marianayovcheva:
As part of the spanner work, we have discussed Kath and Clair also doing design system dependabot updates. I have updated the codeowners file to give them the required access

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,10 @@
 # Global owners - all PRs will be assigned to these users
 *       @davidrapson @davidsauntson @marianayovcheva
+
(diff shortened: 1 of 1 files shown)


davidrapson (commented):
I think it might be better to leave this as an informal thing. Reason being: With the codeowners set github will automatically request Kath and Clair as reviewers which I imagine would be _very_ noisy on the weeks where they don't need to care.

clairgriffiths:
Is there a way of giving us permission to do this but not be tagged on every PR?

davidrapson:
Going to close this based on the decision on Slack to probably exclude design system from the rota. Plus because making this change would notify kath and clair every time there's a dependency update.

PR #2071, opened 2022-05-27
Feat/form group page heading

marianayovcheva:
Added `legend_heading` to the form group options. Addresses a semantic issue of having no `h1` on 'one question per page' type of forms following gov.uk's approach described here: https://design-system.service.gov.uk/get-started/labels-legends-headings/
Set up a page for radio buttons in the `design-system-docs`.

### DIFF
--- a/design-system-docs/src/_component_docs/radio-group.md
+++ b/design-system-docs/src/_component_docs/radio-group.md
@@ -84,7 +84,7 @@ Any additional key/value pairs (beyond `label` and `value`) in your radio button
 
 ### With heading in the legend
 
-Used for one question per page forms.
+Used for one question per page forms. Using similar approach to the one described in gov.uk - [Making labels and legends headings](https://design-system.service.gov.uk/get-started/labels-legends-headings/#legends-as-page-headings).
 
 <%= render(Shared::ComponentExample.new(:radio_group, :legend_heading)) %>
 
--- a/design-system-docs/src/_component_docs/radio-group.md
+++ b/design-system-docs/src/_component_docs/radio-group.md
@@ -10,8 +10,8 @@ A radio allows a user to select only one option from a group of exclusive choice
 
 Do not use a radio group if the userL
 
-- should be able to select one or more choices from a list of options (use a [checkbox group](/docs/forms-checkbox-group--basic) for this)
-- there is only one checkbox in the list of options, for example agreeing to terms and conditions (use [checkbox](/docs/forms-checkbox--basic) for this)
+- should be able to select one or more choices from a list of options (use a checkbox group for this)
+- there is only one checkbox in the list of options, for example agreeing to terms and conditions (use checkbox for this)
 
 ## How it works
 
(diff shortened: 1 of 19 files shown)


design-system-docs/src/_component_examples/_radio_group/additional_attributes.erb
```diff
@@ -0,0 +1,12 @@
+---
```
marianayovcheva, line 1: I'm not sure if/how to lint those files or if we should add a linting command later on

design-system-docs/src/_component_examples/_radio_group/default.erb
```diff
@@ -0,0 +1,12 @@
+---
+title: Default
+---
+<%= render CitizensAdviceComponents::RadioGroup.new(
+  legend: "Example radio group",
+  name: "radio-buttons"
+) do |c|
+  c.inputs([
```
marianayovcheva, line 8: Is there a way to reuse a radio group variable the same way it's done in `radio_group_preview.rb` (i.e `c.inputs(radios)`)? I found myself copy pasting the same inputs quite a lot.

design-system-docs/src/_data/component_arguments.yml
```diff
@@ -9,3 +9,22 @@ targeted_content:
...
   - argument: close_button
     description: Optional, include close button at the end of content. Defaults to true
+radio_group:
+  - argument: name
+    description: Required, field name
+  - argument: label
+    description: Required, the text for the label associated with the input
+  - argument: options
+    description: Optional, a hash with one or more of the following values
+  - argument: hint
```
marianayovcheva, line 19: This is probably a question for a future PR but I'm not sure how to indent the arguments that are part of the options, in the old docs it is done with '&rarr; hint' but this throws an error here

design-system-docs/src/_component_examples/_radio_group/default.erb
davidrapson: Possibly: You could define the options as front matter in `_component_examples/_radio_group/_defaults.yml` and use the data from each example. But I'm also fine with it being explicit for example. Makes copy-pasting the example from the docs less error prone.

design-system-docs/src/_component_examples/_radio_group/additional_attributes.erb
davidrapson: We could maybe use something like [erb-lint](https://github.com/Shopify/erb-lint) but I don't know if the front matter would confuse things. I'm inclined to not worry about linting for the docs site, especially as we want designers to be able to easily edit things.

design-system-docs/src/_component_docs/radio-group.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Additional attributes (View component only)
+
+Any additional key/value pairs (beyond `label` and `value`) in your radio button definition will be added to the input.
+
+<%= render(Shared::ComponentExample.new(:radio_group, :additional_attributes)) %>
+
+### With heading in the legend
+
+Used for one question per page forms.
```
davidrapson, line 87: Would you mind sticking a link somewhere to https://design-system.service.gov.uk/get-started/labels-legends-headings/#legends-as-page-headings to remind us why we do this?

davidrapson (changes requested):
Looks great! Couple of small suggestions but fab so far. It looks like you've got a bunch of unexpected changes to the demo package lock file, mind fixing that up first?

marianayovcheva:
@davidrapson good should about `package-lock.json` in `demo`. I've reverted it locally and tried to see at what point it started to differ. What I found was that when I was in `/demo` and running `./bin/setup` it would alter my `package-lock.json` but if I ran `npm ci` separately, it wouldn't change the lock file. Not sure why this happens though and how to prevent it from happening in the future

davidrapson (approved):
👍 Looks great. Mind updating the changelog too?

PR #2099, opened 2022-06-10
Add HTML validation check in Cypress tests

davidrapson:
Prompted by https://github.com/citizensadvice/design-system/pull/2098 I thought I'd try adding an offline HTML validation check to all our component previews to see if it flagged up any issues. Uses `cypress-html-validate` to run the tests.

A decent few things here, but thankfully nothing major. I'll keep this as a draft and then extract some of the fixes out of here. Some are uncontroversial, others need some review.

PR #2105, opened 2022-06-13
Header: Fix issues with search toggle in header

davidrapson:
Prompted by https://github.com/citizensadvice/design-system/pull/2099 this PR attempts to fix a few accesibility and markup quality issues with the header search toggle:

1. Was using title rather than aria-label for the description
2. Was missing a type attribute on the button element
3. Was missing aria-controls to associate the button with the search
4. Both the toggle and the search form were using icon fonts which causes additional text to be read out when interacting with either element. Switch to use SVG icons. See https://github.com/citizensadvice/design-system/issues/1902

### DIFF
--- a/engine/app/components/citizens_advice_components/header.html.haml
+++ b/engine/app/components/citizens_advice_components/header.html.haml
@@ -9,12 +9,16 @@
       .cads-grid-col-md-2.cads-header__logo-row
         = logo
         - if search_form.present?
-          = tag.button(class: "cads-header__search-reveal js-cads-search-reveal cads-icon_search",
-            title: t(".open_search"),
+          = tag.button(class: "cads-header__search-reveal js-cads-search-reveal",
+            type: "button",
+            "aria-label": t(".open_search"),
             "aria-expanded": "false",
+            "aria-controls": "header-search-form",
             "data-testid": "expand-button",
             "data-descriptive-label-show": t(".open_search"),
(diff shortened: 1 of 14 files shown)


davidrapson:
There's a small github actions issue at the moment so this will probably fail right now: https://www.githubstatus.com/incidents/6r4sms3hcj3x

engine/app/components/citizens_advice_components/header.html.haml
```diff
@@ -9,12 +9,15 @@
       .cads-grid-col-md-2.cads-header__logo-row
         = logo
         - if search_form.present?
-          = tag.button(class: "cads-header__search-reveal js-cads-search-reveal cads-icon_search",
-            title: t(".open_search"),
-            "aria-expanded": "false",
```
davidsauntson, line 14: I think you need to keep `aria-expanded`, looking at the docs: https://w3c.github.io/aria/#aria-expanded

davidsauntson (commented):
I think you'll also need to update the example HTML in the `styleguide` folder

engine/app/components/citizens_advice_components/header.html.haml
davidrapson: Although saying that I'm explicitly setting the other aria attributes in the template so you're right I should explicitly leave this for clarity.

PR #2108, opened 2022-06-15
Fix invalid demo app markup

davidrapson:
Extracted out from https://github.com/citizensadvice/design-system/pull/2099. Fixes a few small markup issues within the demo app.


davidrapson:
Merging this one as it's just cleaning up the demo app

PR #2111, opened 2022-06-16
Add full colour foundations page

davidrapson:
Uses sass-export to export a json representation of the colour language to src/_data which is made available within bridgetown. This lets us port the equivalent colour tables from storybook.

This data file will need syncing with `npm run update-colour-language` if we change the values but this happens infrequently enough that this is fine to do manually when that happens.


marianayovcheva (approved):
🎨

PR #2113, opened 2022-06-17
Pagination: Remove redundant role

davidrapson:
A role of navigation is implied when using the nav element. Flagged by cypress-html-validate as part of https://github.com/citizensadvice/design-system/pull/2099


davidsauntson (commented):
Also nice - can you update the `styleguide` example html pls

PR #2115, opened 2022-06-17
Targeted content: Fix validation issues

davidrapson:

1. Wrapper elements were divs which get wrapped by a button so they should be spans (invalid for a button to contain a div)
2. Generated buttons were missing an explicit type


davidsauntson (commented):
🎯  nice - can you update the example html in `styleguide` too pls?

PR #2116, opened 2022-06-17
On this page: Add missing button type

davidrapson:

Buttons should have an explicit type

### DIFF
--- a/engine/app/components/citizens_advice_components/on_this_page.html.haml
+++ b/engine/app/components/citizens_advice_components/on_this_page.html.haml
@@ -4,7 +4,7 @@
     - links.each do |link|
       %li.cads-on-this-page__list-item
         - if show_children_for?(link)
-          %button.cads-on-this-page__show.cads-icon_plus.cads-linkbutton.js-cads-on-this-page__toggle{aria: { expanded: false }, data: { testid: "on-this-page-toggle", toggle_target_id: "#{link.id}-list", label_when_hiding: "Show sub-headings for #{link.label}", label_when_showing: "Hide sub-headings for #{link.label}" } }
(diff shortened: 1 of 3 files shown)


engine/app/components/citizens_advice_components/on_this_page.html.haml
```diff
@@ -4,7 +4,7 @@
     - links.each do |link|
       %li.cads-on-this-page__list-item
         - if show_children_for?(link)
-          %button.cads-on-this-page__show.cads-icon_plus.cads-linkbutton.js-cads-on-this-page__toggle{aria: { expanded: false }, data: { testid: "on-this-page-toggle", toggle_target_id: "#{link.id}-list", label_when_hiding: "Show sub-headings for #{link.label}", label_when_showing: "Hide sub-headings for #{link.label}" } }
+          %button.cads-on-this-page__show.cads-icon_plus.cads-linkbutton.js-cads-on-this-page__toggle{ type: "button", aria: { expanded: false }, data: { testid: "on-this-page-toggle", toggle_target_id: "#{link.id}-list", label_when_hiding: "Show sub-headings for #{link.label}", label_when_showing: "Hide sub-headings for #{link.label}" } }
```
davidrapson, line 7: Lotta attributes here, I don't know if we want to extract these into a method?

marianayovcheva: I think you're right, it's becoming a bit tricky to follow so would be good if those are separate

marianayovcheva (approved):

engine/app/components/citizens_advice_components/on_this_page.html.haml
davidrapson: Might come back to this one as we need to update the template to swap out the icon font anyway soon

PR #2137, opened 2022-06-22
Greedy nav: Fix HTML validation issues

davidrapson:

TODO: 

- [x] Wrap nav in div not span
- [x] Add aria-controls to toggle button
- [x] Replace `aria-labelledby="cadsGreedyNavLabel"` with a plain `aria-label`
- [x] Make `aria-label` contextual to state ([NP-2772](https://citizensadvice.atlassian.net/browse/NP-2772))


src/ts/greedy-nav/GreedyNav.ts
```diff
@@ -184,6 +150,7 @@ export const updateLabel = (
   }
 
   toggle.innerHTML = label;
+  toggle.setAttribute('aria-label', `${label} navigation options`);
```
marianayovcheva, line 153: I must admit, there are a couple of instances where `navDropdownBreakpointLabel` is used for the label, however, when I was manually testing it, the labels were only 'More' or 'Close'. I feel like we should completely refactor the greedy nav at some point in the future.

davidsauntson (approved):
🙌🏻

src/ts/greedy-nav/GreedyNav.ts
davidsauntson: ➕ it is totally in need of an overhaul

Issue #1981, opened 2022-04-25
Migrate foundations page: Visibility


davidrapson:
Migrate Visibility foundations page to the new documentation site: https://citizensadvice.github.io/design-system/?path=/docs/foundations-visibility--page

There's a few inline visibility examples here that might warrant wrapping in the example UI to make them clearer, but otherwise this could be migrated over as-is.

PR #2150, opened 2022-06-27
Feat/new docs visibility

marianayovcheva:
https://github.com/citizensadvice/design-system/issues/1981

While tidying up the issues, I've realised I haven't finished off the visibility page. 
It works similar to the other foundation pages. A thing to mention is that I've overridden the default `iframe: true` in order to display how the visibility changes when the user resizes the screen.


davidrapson (approved):
Nice!

Issue #1980, opened 2022-04-25
Migrate foundations page: Typography


davidrapson:
Migrate Typography foundations page to the new documentation site: https://citizensadvice.github.io/design-system/?path=/docs/foundations-typography--paragraph

Similar to the grid page, code examples here are HTML only so maybe we need a way to disable ERB examples for specific examples.

PR #2152, opened 2022-06-28
Feat/migrate typography

davidsauntson:
Migrates the ol' typography page into the new docs format.  `iframe: false` added to these so that the examples display with desktop styles rather than mobile ones.

#1980 

image

### DIFF
--- a/design-system-docs/src/_component_examples/_foundations/typography_external_link.erb
+++ b/design-system-docs/src/_component_examples/_foundations/typography_external_link.erb
@@ -1,0 +1,13 @@
+---
+title: External link
+iframe: false
+---
+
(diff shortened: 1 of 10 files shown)


design-system-docs/src/_component_examples/_foundations/typography_external_link.erb
```diff
@@ -0,0 +1,13 @@
+---
+title: External link
+iframe: false
+---
+
+<div class="cads-prose">
```
davidsauntson, line 6: Decided to be explicit about the `.cads-prose` element in the examples as well as the body copy, which is a change from the existing docs site.

PR #2153, opened 2022-06-28
Feat/code example styles

davidsauntson:
Brings the `code` blocks and `ComponentExample`... component closer to the Figma designs.

Adds a custom theme in site builder dir, registers it with `Rouge` and renders the theme in the default layout.  I'm not sure how much value the brand colours add, but it's better than it was!

image

### DIFF
--- a/design-system-docs/frontend/styles/_code.scss
+++ b/design-system-docs/frontend/styles/_code.scss
@@ -3,20 +3,23 @@
 // ============================================================================
 
 code {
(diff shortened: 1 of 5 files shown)


davidrapson (approved):
Nice!

Did wonder if you could define a custom syntax theme. Great find. I wonder how much we'll need to tweak it for different code samples but fun to try so let's go for it!

PR #2154, opened 2022-06-28
Feat/code copy

davidsauntson:
Adds the `Copy code` button to code blocks rendered both in markup and in the `ComponentExample` component.  The functional part of the feature requires JS anyway, so the button and styles are added using JS too.

I'll add wider support for non-js in a different PR - but for this the button just doesn't appear (which is what we want).

There is a slight change from the figma designs - wrapping the code isn't very helpful most of the time, so I had to move the button.  I'll check it with @JamesR-CA 

image

### DIFF
--- a/design-system-docs/frontend/styles/_code.scss
+++ b/design-system-docs/frontend/styles/_code.scss
@@ -5,7 +5,6 @@
 code {
   line-height: 1rem;
   margin: 0 2px;
-  padding: $cads-spacing-1 $cads-spacing-2;
   border-radius: $cads-border-radius;
   background-color: #f8f8f8;
   font-size: 1rem;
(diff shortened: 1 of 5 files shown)


design-system-docs/frontend/styles/_code.scss
```diff
@@ -18,8 +17,23 @@ pre code {
...
+  padding: $cads-spacing-4;
+
+  &--with-button {
+    margin: 0;
+  }
+
+  button {
+    font-family: $cads-font-family;
+  }
+  > .highlight {
```
davidsauntson, line 35: I can't work out why, but Rouge and/or Kramdown is creating the following structure when converting a ``` code block in markdown into a `code` element:
`div.highlight > pre.highlight > code`

Not sure where the additional wrapping `div` is coming from, so I've had to add this into the SCSS

marianayovcheva (approved):
LGTM! Only question I have is more about UX and not code - do we need to add an indication  that content has been copied. In the reference component on public website a success message appears. Here maybe something like the text changing for the text for a few seconds to 'Copied'?

PR #2155, opened 2022-06-29
feat: add no-js support

davidsauntson:
Adds no-js support to the docs site, based on the same model as public website:
- add `html.no-js` to the default layout
- remove it again in a `script` block in the `head`
- add some error handling in the main js file

### DIFF
--- a/design-system-docs/frontend/javascript/index.js
+++ b/design-system-docs/frontend/javascript/index.js
@@ -4,6 +4,11 @@ import initTargetedContent from '@citizensadvice/design-system/lib/targeted-cont
 import initDisclosure from '@citizensadvice/design-system/lib/disclosure/disclosure';
 import initCodeCopy from './code-copy';
 
-initTargetedContent();
-initDisclosure();
-initCodeCopy();
+try {
+  initTargetedContent();
+  initDisclosure();
+  initCodeCopy();
+} catch (error) {
+  document.querySelector('html').classList.add('no-js');
+  throw error;
+}
--- a/design-system-docs/src/_layouts/component_example.erb
+++ b/design-system-docs/src/_layouts/component_example.erb
@@ -1,11 +1,14 @@
 <!doctype html>
-<html lang="<%= site.locale %>">
+<html class="no-js" lang="<%= site.locale %>">
   <head>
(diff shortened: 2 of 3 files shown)


marianayovcheva (approved):
Nice! 
A side question - does the component_example.erb also need no-js support (not sure how this works when it's an iframe)

davidsauntson:
> Nice! A side question - does the component_example.erb also need no-js support (not sure how this works when it's an iframe)

I think we're ok, because:
- the components are rendered in the iframe view and they should be build to handle the absence of JS anyway, if they need to (eg target content and disclosure just appear open
- the `ComponentExample` component is OK because it uses components (🤯) that are already fine without JS

LMK if I've misunderstood tho and we can have a chat - I'll wait till you get back to me 🙂

marianayovcheva:
I think we need to check this for every component that uses js - I just found that `on-this-page` with nested links doesn't show the nested links if js is disabled. Targeted content is a nice example as it doesn't rely on `.no-js` in the scss. The ones that do, if added as an iframe (which I don't know how much we'll do) or when open in a new tab will not work so smoothly without `.no-js`.

davidrapson:
For the component iframe / new window I think mirroring the same no-js scripts in https://github.com/citizensadvice/design-system/blob/72188ce0f0c5194b7096252621a6604d73e34c10/design-system-docs/src/_layouts/component_example.erb would cover us there.

davidsauntson:
> I think we need to check this for every component that uses js - I just found that `on-this-page` with nested links doesn't show the nested links if js is disabled. Targeted content is a nice example as it doesn't rely on `.no-js` in the scss. The ones that do, if added as an iframe (which I don't know how much we'll do) or when open in a new tab will not work so smoothly without `.no-js`.

Ah - gotcha! Yup I'll add `no-js` to the layout for the iframe view as @davidrapson suggests.

For on this page, my recollection is that not showing the nested links was a conscious decision since:
- they are displayed hidden by default
- showing all of them at the same time would be too much
- the core behaviour is seeing the main links and the show/hide is a nice sprinkle of prog enhancement on top

marianayovcheva (approved):
🙌

Issue #1694, opened 2021-12-09
Error summary links scrolling past the form legend


marianayovcheva:
## Summary of problem
Error summary links scroll past the form legend, to see the legend, the users need to scroll up

## Steps to replicate
1. Scroll to the bottom of a [content page](https://www.citizensadvice.org.uk/immigration/staying-in-the-uk-if-youre-from-the-eu/) and from the advice feedback section ('Did this advice help?') select **No**
2. This will open a form - submit the form without filling in the form (we want to see the error message appear)
3. When the error summary appears, click on a link (in the example case 'Select an option')

## Expected behaviour
When clicking on a link from the error summary, this should scroll the user to the legend before the input so that the user is aware what needs to be adjusted in their input

## Actual behavior
When clicking on a link from the error summary, the page scrolls to the input, past the legend which means that the user would need to scroll up to check what they need to input
## Context

- url - content pages, example: https://www.citizensadvice.org.uk/immigration/staying-in-the-uk-if-youre-from-the-eu/
- screenshot
  - actual
![image](https://user-images.githubusercontent.com/94383569/145395625-bf0bea52-9d6f-499b-a0b8-8e5beaeda6c5.png)
  - expected
![image](https://user-images.githubusercontent.com/94383569/145395702-d6d9d251-01f6-4525-a486-cf1597a09c99.png)

- browser / os / device - any

marianayovcheva:
This actually exists in Jira: https://citizensadvice.atlassian.net/jira/software/c/projects/NP/boards/208?modal=detail&selectedIssue=NP-2423&search=error%20summary

davidrapson:
This has been fixed

PR #2156, opened 2022-06-29
Fix/error summary scroll

marianayovcheva:
Original Issue:
https://github.com/citizensadvice/design-system/issues/1694
https://citizensadvice.atlassian.net/browse/NP-2423
> The links in the advice feedback form error summary link to the inputs so they get focus.  This means when you click on a link in the error summary you are scrolled to the top of the input and can see neither the error message or the field label.

Approach:
Using a [gov.uk's approach](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/error-summary/error-summary.mjs) as a base.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - 🎯 Targeted content: Fix validation issues with HTML
 - 🔗 Pagination links: Remove redundant `navigation` role
 - 📜 On this page: Add missing type to toggle buttons
+- 🛑️ Error summary: Fixed scrolling to the input label/legend
 
 ## <sub>v5.3.0-alpha.1</sub>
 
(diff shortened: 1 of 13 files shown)


davidrapson:
> For other components where we use a polyfill, we don't include the polyfill in the design system itself, we make a note on the component page and ask consumers to make sure they include it if they want to support older browsers (eg https://citizensadvice.github.io/design-system/?path=/docs/components-targeted-content--default#javascript-behaviour)

Not for this PR but I noticed working on https://github.com/citizensadvice/design-system/pull/2137 that we have another inline element.closest polyfill there too; so I should tidy that up too.

davidrapson:
Sorry didn't mean to close this 🙈

marianayovcheva:
Thanks! I've now removed the polyfill from the error summary. I think I'll keep the function as it is for now as when I've tried to adjust it locally I managed to make it less readable 😬

davidsauntson (approved):
🙌🏻

PR #2157, opened 2022-06-29
Adding Asset hyperlink to the new docs site

marianayovcheva:
Two questions that will potentially go into separate PRs:

- Should we include 'On this page' for the foundation pages as well?
- Should I remove the external icon after the hyperlink? 
image

### DIFF
--- a/design-system-docs/src/_component_docs/asset-hyperlink.md
+++ b/design-system-docs/src/_component_docs/asset-hyperlink.md
@@ -1,0 +1,24 @@
+---
+title: Asset hyperlink
+---
+
+<%= render(Shared::ComponentExample.new(:asset_hyperlink, :default)) %>
+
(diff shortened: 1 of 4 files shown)


davidsauntson (approved):
Nice!

> Should we include 'On this page' for the foundation pages as well?

I was thinking the same when I did the typography page, I think it might be useful so let's try it and see


> Should I remove the external icon after the hyperlink?

Did the design-system group ever make a decision on this? If they decided they to remove it, you might as well do it now.  Otherwise, let's leave it.

davidrapson:
> Did the design-system group ever make a decision on this? If they decided they to remove it, you might as well do it now. Otherwise, let's leave it.

We didn't yet but thank you for reminding me we need to bring it up. Let's discuss that next time.

PR #2158, opened 2022-06-30
Upgrade to stylelint v14

davidrapson:
Restructures stylelint dependencies and to use stylelint-config-standard-scss in favour of manually configuring scss syntax as recommended in the docs. This automatically handles the new customSyntax requirements from the new
stylelint version and reduces the number of direct packages we manage.

Updates rules to reflect the new config and any new rules in stylelint v14. Removes any rules that are defined by newer versions of the standard config.

### DIFF
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -1,6 +1,6 @@
 module.exports = {
-  extends: ['stylelint-config-standard', 'stylelint-config-prettier'],
-  plugins: ['stylelint-scss', 'stylelint-selector-bem-pattern'],
+  extends: ['stylelint-config-standard-scss', 'stylelint-config-prettier'],
+  plugins: ['stylelint-selector-bem-pattern'],
   ignoreFiles: ['**/_normalize.scss', '**/_cads-iconfont.scss'],
   rules: {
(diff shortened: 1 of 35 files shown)


stylelint.config.js
```diff
@@ -32,6 +35,9 @@ module.exports = {
     'at-rule-disallowed-list': ['debug'],
     'scss/at-rule-no-unknown': true,
 
+    // Disable unwanted scss rules
+    'scss/comment-no-empty': null,
```
marianayovcheva, line 39: I was initially surprised that we had empty comments but then pulled and saw where those are. Tbh I have a personal preference for `/* Long comment on multiple lines */` pattern but not sure how others feel about it and even if we decide to change the comments feels like a new PR

marianayovcheva (approved):

stylelint.config.js
davidrapson: Yeah I think we prefer `//` style comments because they are not output in the final CSS and we're largely documenting that Sass code here, where `/* */` is a real CSS comment and is output in the built file. Not super precious about comments though hence disabling the rule to let us do what makes sense.

PR #2160, opened 2022-07-01
Replacing paragraph with heading for error summary (DAC audit) A

marianayovcheva:
https://citizensadvice.atlassian.net/browse/NP-2770

This was flagged in the DAC audit (p. 18):

> The text of ‘There is a problem’ contained within the error summary is not marked up as a heading. As this text is used to introduce the same page skip link, this should be marked up as a heading to ensure screen reader users are given the same context as a sighted user gets from visual cues.

This PR implements the recommended fix.


davidsauntson (commented):
Interesting one this, since this doesn't always appear underneath an `h1`, eg in decision tree questions: 
image


~Although I have lost track of what valid heading orders are 🙈~ Yeah I think this will be an issue.  Maybe we could default to an `h2` and allow users to specify what they want the title to be rendered as?  Otherwise we could sacrifice control of the error summary title and have a slot for users to add whatever title they wish?

davidrapson:
Might be one to confirm with DAC as part of the current report, or in the next drop-in. 

AFAIK a h2 before a h1 is OK e.g. to introduce navigation or other secondary content https://www.w3.org/WAI/tutorials/page-structure/headings/#main-heading-after-navigation. I think the main requirement is for heading to not skip around levels.

I'd expect it to be worse for the error summary title to be the `h1` in this case, but I could easily be wrong.

davidsauntson:
> I'd expect it to be worse for the error summary title to be the `h1` in this case, but I could easily be wrong.

I think you're definitely right!

The gov.uk example has the `h2` at the top before an `h1`, which is the opposite of what I thought was OK on reading mdn's docs.  https://design-system.service.gov.uk/components/error-summary/full-page-example/index.html. But it has convinced me so I'm fine.  I might ask at the next drop in anyway 👍🏻

davidrapson:
Yeah we definitely should check 👍 WCAG is really fuzzy on this. Sounds like we're happy to ship this as is based on the recommendation in the report and that it matches GOV.UK here.

marianayovcheva:
Thanks both! Now that I think about this more, I wonder if we should go back to David S' original suggestion and allow users to specify what element the title to be rendered as. This is not so much for the instances when the error summary is before h1 (because after reading the documentation I think it's fine but I'll pop a note in the ability net drop in doc). It's for  'Email this advice to yourself':
image
'Email this advice to yourself' is `h2` and if 'There is a problem' is part of that section, should it not be `h3`?

davidsauntson (approved):
🙌🏻 Can you update the example HTML in the styleguide section too pls.

PR #2161, opened 2022-07-01
Renaming skip to content link (DAC audit) AAA

marianayovcheva:
https://citizensadvice.atlassian.net/browse/NP-2785

This issue was flagged by the DAC audit (p.51):

> The skip link text was not fully descriptive of its function.

This PR includes the recommended solution


davidrapson (approved):
👍  Mind updating the changelog?

Not a blocker, but this text is translated into welsh so we should get "main" added there too.

marianayovcheva:
> Not a blocker, but this text is translated into welsh so we should get "main" added there too.

Good shout, thanks! I've requested a translation, I'll wait for it before I merge and will update the Changelog as well.

PR #2170, opened 2022-07-04
Allow opting-out of external link styles

davidrapson:
A recent accessibility report flagged that our external link icon can be interpreted as a new tab icon causing confusion. This mirrors the issues mentioned in https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/

We also know that we have an outstanding accessibility issue with our icon fonts which affects this icon too (see: https://github.com/citizensadvice/design-system/issues/1902)

To allow us to test removing the external link icon, add a new config option to opt-out of the external link styles: `$cads-enable-external-link-icon`. This is enabled by default to respect the current defaults.

NP-2782


scss/1-settings/_config.scss
```diff
@@ -0,0 +1,5 @@
+// ============================================================================
+// Config flags
+// ============================================================================
+
+$cads-enable-external-link-icon: true !default;
```
davidsauntson, line 5: should be `$cads-enable-opens-in-new-tab-icon`

LOL JK

davidsauntson (approved):
🔗 🚮

scss/1-settings/_config.scss
davidrapson: 👏

PR #2179, opened 2022-07-11
Feat/new docs buttons

marianayovcheva:
Adding the button component to the new docs site.
While working on this, I had issues with Bridgetown rendering the content, so I had to change `button.html.haml` to use the haml syntax instead of the rails `tag`. Before changing the file I have tried adding `.raw` and `.html_safe` to the [component_example](https://github.com/citizensadvice/design-system/blob/main/design-system-docs/src/_components/shared/component_example.erb#L11) but this didn't resolve the issues:
- duplicated text content
image

- duplicated text issue plus the svg rendering the element 
image


davidsauntson:
Hmmm, there is deffo some problem with how this is rendering using the `tag.button` helpers.  I think they might not be being included in the bridgetown build perhaps?  We will need to work it out because there are a few components that use them.

If I just try and render the button component in markdown, like this:

```
<%= render CitizensAdviceComponents::Button.new(variant: :primary, attributes: { "data-testid": "example" }).with_content("Hello") %>
```
I get this
image

Which to my mind means one or more of:
- whatever is parsing the markdown isn't creating html correctly
- the helpers (`tag.button`) isn't understood outside the context of the rails app - it is a rails helper but the engine does have rails as a dependency

marianayovcheva:
Update: The issue in the description has been successfully resolved by merging in `main` which contains https://github.com/citizensadvice/design-system/pull/2215

design-system-docs/src/_component_examples/_button/previous_next.erb
```diff
@@ -0,0 +1,19 @@
...
+  end %>
+</div>
+<div>
+  <%= render CitizensAdviceComponents::Button.new(variant: :secondary) do |c| %>
+    Previous
+    <% c.icon_left do
+      render CitizensAdviceComponents::Icons::ArrowLeft.new
+    end
+  end %>
+</div>
```
davidrapson, line 19: I'd be tempted to make these two separate examples but I don't think it matters too much

davidrapson (approved):
🚀

design-system-docs/src/_component_examples/_button/previous_next.erb
marianayovcheva: Thanks! I'll leave it as it is for now and merge. Only reason is that I want to see how the content will be changed and I'll need some help with the wording if we split the examples - probably one for @danielnissenbaum

PR #2206, opened 2022-07-25
Docs: Speed up yarn installs

davidrapson:
Currently running `yarn install` in the docs package take a long time to resolve the design-system using the `file:` install method. I also tried using `link:` to install the package which should use a simple symlink. In both cases the install stalls for a long time, presumably because the mode module is at the root of the project so it's getting confused with some circular behaviour.

Either way linking using relative paths has exactly the same effect locally, we don't actually need to install the package. Switching to relative paths fixes any install issues.

### DIFF
--- a/design-system-docs/frontend/styles/index.scss
+++ b/design-system-docs/frontend/styles/index.scss
@@ -1,4 +1,4 @@
-@import '~@citizensadvice/design-system/scss/lib';
+@import '../scss/lib';
 
 @import 'layout';
 @import 'code';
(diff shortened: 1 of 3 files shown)


davidrapson:
This speed issue can be seen even in github actions, where the `yarn install` step took a few seconds for this build https://github.com/citizensadvice/design-system/runs/7495535045#step:7:18 where the last build to a minute https://github.com/citizensadvice/design-system/runs/7284168530?check_suite_focus=true#step:7:24

Locally it's much worse and I've seen the install hanging for several minutes.

design-system-docs/frontend/styles/index.scss
```diff
@@ -1,4 +1,4 @@
-@import '~@citizensadvice/design-system/scss/lib';
+@import '../scss/lib';
```
davidrapson, line 1: This feels like too few `..` but the sass-loader seems to resolve from the root of the project, so this works as expected.

marianayovcheva (approved):
That's a good find and it works faster for me! 🏃

PR #2207, opened 2022-07-25
Demonstrate output buffer bug

davidrapson:
As shown in https://github.com/citizensadvice/design-system/pull/2179 there is an issue rendering components in bridgetown where blocks are used.

This causes an output buffer issue where the contents of the block is also rendered on the page.

This only seems to happen with haml templates not erb templates. It's possibly a bridgetown bug but more likely down to the fact we are having to approximate haml support in our builder by having to patch in a new template handler.

image

image

This pull request acts as a reduced test case to demonstrate the issue. I'll raise a pull request based on this showing how switching to ERB bypasses any underlying issues.

### DIFF
--- a/design-system-docs/src/_component_docs/output-buffer-bug.erb
+++ b/design-system-docs/src/_component_docs/output-buffer-bug.erb
@@ -1,0 +1,10 @@
+---
+title: Output buffer bug
+---
+
+<%= render CitizensAdviceComponents::Button.new(variant: :secondary) do |c| %>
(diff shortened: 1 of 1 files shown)


davidrapson:
Page is here showing the issue: https://62de6067e705d1131da56e5b--ca-design-system.netlify.app/components/output-buffer-bug/

davidrapson:
Updated the demo to include an input example too which really highlights the issue.

PR #2208, opened 2022-07-25
Fix output buffer issue by switching to ERB for affected components

davidrapson:
This change demonstrates how switching to ERB for internal component templates bypasses the output buffer issues seen in https://github.com/citizensadvice/design-system/pull/2207 when using design system components within bridgetown.

It's possible to workaround some of the issues by adjusting the haml templates with delicate use of `capture_haml` or by adjusting other parts of the templates as shown in https://github.com/citizensadvice/design-system/pull/2179 but this feels fraught.

By switching to ERB for internal component templates we'd be able to avoid the need for any custom templating logic in the builder as ERB is a core part of ruby so has a higher likelihood of working without caveats and be officially supported.

image

Because view component exists to wrap the underlying templates in an interface we don't need to change all the templates in one go—only the affected ones—and it doesn't affect our ability to use haml in the main apps.

I'd appreciate critical review on this, and thoughts on if there's another way to approach this. I'm biased as I don't particularly care for haml but it feels like if we can avoid issues by using something built-in we should consider that.


davidsauntson (approved):
It's a shame to lose HAML (this makes me ill: `<div class="cads-form-field<% if error? %> cads-form-field--has-error<% end %>">`) but I agree we should take the simplest approach to fixing the issue.

I wonder if it's only worth doing for the ones that are broken (where we use `tag`, right?) and leaving the rest or porting everything over to erb. Having both might be a bit confusing long term and looks like there are some tools to help (https://haml2erb.org/api-reference), so my vote would be for switching everything over.

I wonder also if switching to erb would encourage us to keep logic out of templates?  Eg the snippet below should probs be replaced with a `render_content` method or smth in the class file? (Not for now, just an example).  I will be much less inclined to write this sort of thing in erb
```
<% if content.present? %>
  <%= content %>
<% else %>
  <%= tag.input(class: class_names.join(" "), **input_attributes) %>
<% end %>
```
So I'm fine with this - it would be interesting to know what the other solutions that allow us to keep haml look like but we can catch up about that separately if you like.

davidrapson:
> this makes me ill: <div class="cads-form-field<% if error? %> cads-form-field--has-error<% end %>">

I can empathise with that 😅. I went with the smallest change here and whilst I don't _mind_ this style there's no reason we shouldn't either a) push class name logic into the component and/or b) port over something like https://github.com/primer/view_components/blob/main/app/lib/primer/class_name_helper.rb to make things nicer.

Totally agree about the `render_content` suggestion too. I'd be inclined to do that kind of thing too when we do this for real.

> so my vote would be for switching everything over.

I'd agree. That said we only need to start with affected components and target v6 to ditch Haml completely. We can cope with mixing and matching for a bit but agree it doesn't make sense long term.

> So I'm fine with this - it would be interesting to know what the other solutions that allow us to keep haml look like but we can catch up about that separately if you like.

Sounds good. Rather than merging this as is I'll raise a new pull request with an ADR and migrating a couple of affected components over to ERB and happy to run you through what else I tried.

davidrapson:
ADR with initial changes here https://github.com/citizensadvice/design-system/pull/2215

PR #2209, opened 2022-07-25
DAC - Label in name (A)

marianayovcheva:
https://citizensadvice.atlassian.net/browse/NP-2771

Changed the `aria-label` of the close button from `hide` to `close` as suggested by DAC


davidrapson (approved):
👍 Nice

PR #2210, opened 2022-07-25
Improve Backstop reliabilty

davidrapson:
We've been seeing occasional reliability issues with backstop tests for one of two reasons:

1. Tests failing because the app hasn't started fully (early previews failing with errors)
2. Fonts not being fully loaded (the far less frequent issue)

Tries to resolve that by adjusting how we run the tests:

- Runs the app in production mode to try and give the most stable base
- Runs using [start-server-and-test](https://www.npmjs.com/package/start-server-and-test), the main benefit being that we can wait for a specific URL to be ready not just a port. Set this to be the first preview that we use for the backstop scenarios to make sure that the app is actually running OK as best as we can
- Preloads fonts to reduce likelihood of font-loading issues


.github/workflows/ci-workflow.yml
```diff
@@ -117,8 +118,13 @@ jobs:
 
   demo-app-backstop:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        test: [1, 2, 3, 4, 5, 6, 7, 8]
```
davidrapson, line 122: This is set so that backstop runs multiple times on this branch to make it easier to flush out any remaining reliability issues.

demo/config/application.rb
```diff
@@ -24,6 +24,7 @@
 module Demo
   class Application < Rails::Application
     config.load_defaults 6.1
+    config.view_component.show_previews = true
```
davidrapson, line 27: Previews are only loaded in development and test normally

demo/package.json
```diff
@@ -6,7 +6,10 @@
     "backstop": "npx backstop test --docker --config=backstop-config.js",
     "backstop:reference": "npx backstop reference --docker --config=backstop-config.js",
     "backstop:approve": "npx backstop approve --config=backstop-config.js",
-    "backstop:report": "npx backstop openReport --config=backstop-config.js"
+    "backstop:report": "npx backstop openReport --config=backstop-config.js",
+    "backstop:ci": "start-server-and-test start-test-server http://localhost:3001/rails/view_components/button/primary backstop:actions",
```
davidrapson, line 10: Using `http://localhost:3001/rails/view_components/button/primary` as the start URL seems to make the most difference as we're waiting for an actual preview to have loaded not just a port.

davidsauntson (approved):
This looks good and thanks for finding a fix!

Can you add a bit to the docs about why we specify the first preview url as the one used in the `backstop-ci` command? 

I can see the ordering change in the future and it breaking again and no-one knowing why.

davidrapson:
> Can you add a bit to the docs about why we specify the first preview url as the one used in the backstop-ci command?
> I can see the ordering change in the future and it breaking again and no-one knowing why.

I don't think it _has_ to be the first one but that seems most reliable to pick. I added a script to grab the first URL instead of hardcoding it now to avoid confusing our future selves. Good shout.

PR #2212, opened 2022-07-25
Ref 427 DAC_Pseudo_content_02

elceebee:
https://citizensadvice.atlassian.net/browse/REF-427

This ticket addresses an issue raised in the referrals accessibility report. The fix was made to the referrals codebase, but this ticket fixes the issue in the design system components and styleguide so it is no repeated elsewhere. 

It will be a breaking change for anyone who was relying on the classname `cads-form-field__optional` to add the brackets around the word 'optional'. It shouldn't be a breaking change for anyone using the rails components.

[DAC Accessibility Report WCAG 2.1 Citizens Advice referrals tool 16 03 2022 - FINAL .pdf](https://github.com/citizensadvice/design-system/files/9182911/DAC.Accessibility.Report.WCAG.2.1.Citizens.Advice.referrals.tool.16.03.2022.-.FINAL.pdf)

### DIFF
--- a/engine/app/components/citizens_advice_components/form_group.html.haml
+++ b/engine/app/components/citizens_advice_components/form_group.html.haml
@@ -8,7 +8,7 @@
       - else
         = legend
       - if optional?
-        %span.cads-form-field__optional= t("citizens_advice_components.input.optional")
+        %span.cads-form-field__optional= "(#{t("citizens_advice_components.input.optional")})"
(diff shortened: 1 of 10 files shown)


davidrapson (approved):
I know this is still in draft but this looks great to me! One tiny thing before merge this, mind adding something to the CHANGELOG?

PR #2215, opened 2022-07-26
Add ADR for ERB decision and convert initial templates to ERB

davidrapson:
See https://github.com/citizensadvice/design-system/pull/2208 for original context.

- Adds ADR for ERB decision
- Migrates two components over to ERB. The button component which prompted the decision and the header component because it's also affected when using the full component and is complex enough to give a real feel for the change.
- Adds erblint to make sure we have equivalent static analysis for ERB templates.


contributing/adr/0005-use-erb-for-component-templates.md
```diff
@@ -0,0 +1,32 @@
...
+**Pros:**
+
+- Makes it possible to use all components in our docs without trade-offs and simplify our Bridgetown integration
+- ERB is a core part of ruby so has a higher likelihood of working without caveats and be officially supported
+- Allows us to eventually remove `haml-rails` from our dependencies
+
+**Cons:**
+
+- Limits us to only using ERB internally
+- When migrating components over from consumers into the design system
```
davidrapson, line 32: @davidsauntson Does this seem representative for the cons. The first one is the immediate trade-off but felt worth capturing the fact that there'd be an extra step when porting components from public-website.

contributing/adr/0005-use-erb-for-component-templates.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+We've seen output buffer issues when using out components outside of Rails within our Bridgetown docs. This happens where the components using a Haml template and where there is any kind of nested content block (e.g. tag helpers or composition). See https://github.com/citizensadvice/design-system/pull/2207 for an example of the issue and https://github.com/citizensadvice/design-system/pull/2207 demonstrating that using ERB for the template resolves any issues.
+
+It's possible to work around some of the issues by adjusting the Haml templates with delicate use of `capture_haml` or by adjusting other parts of the templates on a case by case basis but this feels fraught.
```
davidsauntson, line 13: It would be nice to expand on this slightly - not War and Peace but just what the pitfalls of it are / why it feels `fraught`.

contributing/adr/0005-use-erb-for-component-templates.md
davidsauntson, line 32: Should this be smth like
`Additional work to convert HAML to ERB when migrating components over from consumers into the design system`

davidrapson: I didn't write an actual sentence here did I 🙈

styleguide/examples/sample_pages/form-sample.html
```diff
@@ -96,6 +96,7 @@
               id="search-query"
               aria-label="Search through site content"
               class="cads-search__input cads-input"
+              autocomplete="off"
```
davidrapson, line 99: Adding `autocomplete="off"` to the search input was recommended by erblint which seems legit so have added this into the template. Nice work linter.

styleguide/examples/sample_pages/form-sample.html
```diff
@@ -6,7 +6,7 @@
           >Skip to navigation</a
         >
         <a class="cads-skip-links__link" href="#cads-main-content"
-          >Skip to content</a
+          >Skip to main content</a
```
davidrapson, line 9: This is just picking up the wording change from https://github.com/citizensadvice/design-system/pull/2161

davidsauntson (approved):
🌿 Nice!

PR #2219, opened 2022-07-26
Ref 427 dac hint usability 01

elceebee:
This is a recommendation coming from the referrals accessibility report. 

https://citizensadvice.atlassian.net/browse/REF-427

### DIFF
--- a/engine/app/components/citizens_advice_components/input.rb
+++ b/engine/app/components/citizens_advice_components/input.rb
@@ -98,7 +98,7 @@ module CitizensAdviceComponents
       classes = []
       classes << error_id if error?
       classes << hint_id if hint?
-      classes.join(' ')
+      classes.join(" ")
     end
 
     def input_attributes
--- a/engine/app/components/citizens_advice_components/input.rb
+++ b/engine/app/components/citizens_advice_components/input.rb
@@ -95,10 +95,10 @@ module CitizensAdviceComponents
     end
 
     def described_by
-      classes = []
-      classes << error_id if error?
-      classes << hint_id if hint?
-      classes.join(" ")
+      ids = []
+      ids << error_id if error?
+      ids << hint_id if hint?
+      ids.present? ? ids.join(" ") : nil
     end
 
     def input_attributes
--- a/engine/app/components/citizens_advice_components/input.rb
+++ b/engine/app/components/citizens_advice_components/input.rb
@@ -78,6 +78,10 @@ module CitizensAdviceComponents
       "#{name}-error"
     end
 
+    def hint_id
+      "#{name}-hint"
+    end
+
     def base_input_attributes
       {
         type: (@type.to_s.dasherize if @type.present?),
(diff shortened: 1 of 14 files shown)


engine/spec/components/citizens_advice_components/input_spec.rb
```diff
@@ -112,6 +112,18 @@
...
+    end
+  end
+
+  context "when there is hint text and an error" do
+    let(:options) { { hint: "this is the hint text", error_message: "Enter your name" } }
+
+    it "adds aria-describedby to the input" do
+      expect(component.css("input").attribute("aria-describedby").value).to eq("example-input-error example-input-hint")
+    end
   end
```
davidrapson, line 127: This is great. Could we add one more test to check that there's no `aria-describedby` when there's no error or hint? I reckon right now it'll output an empty attribute which is probably not what we want. Or maybe Rails is smart enough to work that out?

davidrapson (changes requested):
This is really clear, great work. I reckon we could do with one more test: checking that the attribute is not present by default. And then this looks good to go!

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -86,10 +90,17 @@ def base_input_attributes
         value: value,
         required: required?,
         "aria-invalid": error?,
-        "aria-describedby": (error_id.to_s if error?)
+        "aria-describedby": described_by
       }
     end
 
+    def described_by
+      classes = []
```
davidrapson, line 98: 💅 Nitpick: These are IDs not class names

engine/spec/components/citizens_advice_components/input_spec.rb
elceebee: Good call, I'll add one.

engine/app/components/citizens_advice_components/input.rb
elceebee: I'm so embarrassed.

styleguide/examples/textarea/hint.html
```diff
@@ -8,8 +8,12 @@
     >
       Example input with hint
     </label>
-    <p class="cads-form-field__hint">This is the hint for the input</p>
+    <p class="cads-form-field__hint" id="example-input-with-hint-hint">
+      This is the hint for the input
+    </p>
     <textarea
+      aria-describedby="example-input-with-hint-hint"
+      aria-disabled="false"
```
davidrapson, line 16: One more tiny thing sorry, where is this `aria-disabled` coming from? I know it's just in the examples but we shouldn't expect this to be here right?

davidrapson (approved):
One query about the `aria-disabled` that's turned up in the examples, but once that's removed merge away! 🚀

styleguide/examples/textarea/hint.html
elceebee: I was perhaps being too literal in interpreting the recommendations in the report. Here is the sample they give. It sounds like it isn't necessary. I'll remove it!

image

PR #2220, opened 2022-07-27
Allow sample form to be submitted

davidrapson:
As the sample form now exists in the Rails demo app we can make it actually work.

- Adds a basic form model and controller for the form
- Shows error messages when submitting the form
- Simplifies the form logic to keep things simple enough to model

Really bare-bones but just enough to let us test error summary behaviour in the demo app.

image


marianayovcheva (approved):
It's a very nice addition! ✨
A couple of questions but they are fairly minor so nothing blocking:
- when I first looked at the method name `new` in `example_fom_controller` and the form file name `new.html.erb` I expected that it would be linked to the user creating a new form, but it may just be me
- do the storybook pages need to also be updated?

davidrapson:
> when I first looked at the method name new in example_fom_controller and the form file name new.html.erb I expected that it would be linked to the user creating a new form, but it may just be me

The `new` action is meant to map to the blank form and it's the `create` action that handles the actual submission. This always catches me out too, but this page has a good summary of what the different actions map to in Rails conventions https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

PR #2249, opened 2022-08-03
Upgrade Backstop

davidrapson:
Upgrades to the latest version of Backstop. This version includes playwright as an engine options so we include a couple of options in CI to avoid installing browser engines we don't need.

Updates snapshots to account for a slight change in the engines.


marianayovcheva (approved):
Nice! I guess it's expected that some of the references needed to regenerate?

davidrapson:
> I guess it's expected that some of the references needed to regenerate?

Yeah, it'll use a newer version of chromium behind the scenes and the font rendering seems to have changed _very_ slightly (although I can't really see the difference, backstop seems to notice)

PR #2251, opened 2022-08-05
Documentation website theme

davidrapson:
image


scss/6-components/_navigation.scss
```diff
@@ -6,7 +6,7 @@
 .cads-navigation {
   display: flex;
   justify-content: space-between;
-  background-color: $cads-language__brand-primary;
```
davidrapson, line 9: This was using too broad a variable, so needed to make a dedicated language var for this to avoid changing too much in the docs site context

scss/1-settings/_colour-language.scss
```diff
@@ -29,6 +29,7 @@ $cads-language__warning-colour: palette.$cads-palette__red !default;
 $cads-language__success-colour: palette.$cads-palette__green !default;
 
 // Navigation colours
+$cads-language__nav-background-colour: $cads-language__brand-primary !default;
```
davidrapson, line 32: @marianayovcheva Just an FYI that I've needed to add this and we'll probably need to reconcile things whenever you pick https://github.com/citizensadvice/design-system/pull/1863 back up again

design-system-docs/frontend/styles/_header.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  &:hover,
+  &:focus,
+  &:focus:hover,
+  &:active  {
+    border: none;
+    background-color: transparent;
+    margin-bottom: -3px;
+    border-bottom: 3px solid white;
+  }
```
davidrapson, line 39: This seems like a less messy way of treating the logo when linked compared to what we usually do. Sets a thick border on hover rather than messing with colours.

design-system-docs/frontend/styles/_header.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.cads-header {
+  border-bottom: $cads-border-width-small solid $cads-palette__white;
+
+  // Allow logo row to use auto-width
+  // @TODO: Currently uses grid styles and must always be 2-cols, which seems wrong?
+  .cads-header__logo-row {
+    width: auto;
+    max-width: none;
+  }
```
davidrapson, line 13: Noticed this is a problem on products, where the logos actually overflow technically. The header might make sense using more forgiving grid styles here.

design-system-docs/frontend/styles/_footer.scss
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }
```
davidrapson, line 36: This is more complex than it needs to be a) because the footer logo is a link, should it be? and b) because we don't provide a standard way of either customising this logo or reversing it out.

marianayovcheva: Personally I don't mind it being a link but I now that I think about it, I have more questions than answers 😅 
- if we use the main CA logo, should it not link to the public website site instead?
- can we insert the svg, similarly to how it's done in the header, and then replace every white fill with currentColour ( this also depends a lot on the logo itself)?
- should the header logo and the footer logo have the same focus state?

design-system-docs/frontend/styles/_footer.scss
```diff
@@ -0,0 +1,37 @@
...
+    width: 99px;
+    height: 107px;
+    margin-top: 0;
+
+    &:focus {
```
marianayovcheva, line 18: This is nitpicky so feel free to ignore - should we also use a different colour border bottom on focus (currently black)

marianayovcheva (approved):
This is very nice! Adding the theme really brings the docs site to life! 👏 
I've added a couple of comments but nothing blocking

design-system-docs/frontend/styles/_footer.scss
davidrapson: Yep, I had the same thoughts. I agree making it an SVG would be useful and probably should have the same focus states. I'd be surprised if anyone knew it was a link or clicked on it. We could probably find out from heatmaps of the public site.

design-system-docs/frontend/styles/_footer.scss
davidrapson: Probably yep. What I'll do is come back to the footer logo in a separate PR as I think there's some useful questions there. This code is a bit gross but mirrors what intranet/advisernet do now so we should probably try and fix it up in a more general way.

PR #2252, opened 2022-08-05
Use static placeholder image

davidrapson:
The placeholder service we were using for the sample image in the demo site is down. Rather than rely on a service we can use a simple static image instead. Add a sample image from our future of advice strategy (a simple graphic, no photography)

image


davidrapson:
Going to merge this straight in as it's a little fix and gets the build green again.

PR #2275, opened 2022-08-15
Header: Only render right column when needed

davidrapson:
Extracted from https://github.com/citizensadvice/design-system/pull/2251

Only applies to the Rails engine header component. When rendering the header with only the logo slot the markup for the secondary column is still output. Only render the extra markup when needed.

Best viewed with whitespace ignored: https://github.com/citizensadvice/design-system/pull/2275/files?w=1


davidsauntson (approved):
&rarr;

PR #2279, opened 2022-08-19
Fix/disclosure improvements

marianayovcheva:
(description coming soon)


davidrapson (approved):
👏 This is such excellent work, as always.

> I initially wanted to generate unique ids automatically in Rails, but I couldn't find a way to count the number of disclosures added to a page and generate the ids based on it.

I like the way you've handled it with an optional id: it's clear and explicit. That said, if you ever want to generate random IDs something like `SecureRandom.uuid` might work rather than counting elements, but I don't think that's necessary here.

styleguide/components/disclosure/disclosure-docs.mdx
```diff
@@ -25,6 +25,22 @@ If you have functionality that a lot of users will need, the options should not
   <Story id="components-disclosure--without-open-summary" />
 </Preview>
 
+## With custom id
+
+By default, the disclosure details id, is generated based on the closed label. Adding a custom id is useful in the cases where there are more than one disclosure per page and it is possible that multiple disclosures would have the same label
```
davidsauntson, line 30: Nitpick - no need for the second comma after `id`

davidsauntson (approved):
🥏 Lush, really great! 

Please could you add a line to the changelog too? 🙏🏻

PR #2292, opened 2022-08-24
feat: add breadcrumbs component doc

davidsauntson:
Adds the breadcrumbs component to the new docs site 🍞 🥖 🥯


marianayovcheva (approved):
Breadcrumbs LGTM! 🍞 
Unrelated to the breadcrumbs but the only thing I'm uncertain about is whether we should have a `package-lock.json` as the docs site is using yarn and in my head it should only be `yarn.lock` (also the `yarn.lock` file has a slightly different format but I'm not sure if that's an issue)

davidrapson:
Yep, we should remove the `package-lock.json` file before this goes in; but what Mariana said: LGTM 🚀

davidsauntson:
> Breadcrumbs LGTM! 🍞 
> Unrelated to the breadcrumbs but the only thing I'm uncertain about is whether we should have a `package-lock.json` as the docs site is using yarn and in my head it should only be `yarn.lock` (also the `yarn.lock` file has a slightly different format but I'm not sure if that's an issue)

Whoopsies! This is the second time I've done this today ☀️ thanks for spotting

PR #2296, opened 2022-08-25
Introduce the concept of scoped prose styles

davidrapson:
Often we use the cads-prose class for content where we also embed lots of rich components inside. By default styles within cads-prose cascade down which can subtly affect any components inside. We often work around this in our styles by resetting styles if we are inside a cads-prose element.

I'm proposing that we introduce a new cads-prose-scoped class which uses a child combinator selector to only apply typographic styles to immediate children. Any components within a scoped prose block would not be effected as they would start a new context.

I think the way our standard cads-prose class behaves is a sensible default but it is useful to have an equivalent class that is more targeted for cases where we need a bit more control.

---

Notably using the child combinator in combination with an extend logs a warning that extending this kind of selector will be a warning Dart Sass 2.0.0. Using a mixin here has no major impact as our minifier already dedupes selectors (doing what extend does but automatically).

### DIFF
--- a/design-system-docs/src/_foundations/typography.md
+++ b/design-system-docs/src/_foundations/typography.md
@@ -85,10 +85,10 @@ If you need to apply these styles to heading elements that you cannot add a clas
 </main>
 ```
 
-If you need to apply these styles to heading elements that you cannot add a class to but want more tightly scoped styles than the default `.cads-prose` class you can add a `.cads-prose-scoped` class instead. This is useful when you are embedding lots of rich components within a prose block and you want to more narrowly scope the effects of the cascade.
+If you need to apply these styles to heading elements that you cannot add a class to but want more tightly scoped styles than the default `.cads-prose` class you can add a `.cads-prose-direct-descendants` class instead which will target direct descendants only using the `>` child combinator selector. This is useful when you are embedding lots of rich components within a prose block and you want to more narrowly scope the effects of the cascade.
 
 ```html
-<main class="cads-prose-scoped">
+<main class="cads-prose-direct-descendants">
   <h1>Your options if you're in the UK illegally</h1>
   <h2>Get immigration advice</h2>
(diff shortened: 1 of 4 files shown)


design-system-docs/src/_foundations/typography.md
```diff
@@ -85,6 +85,25 @@ If you need to apply these styles to heading elements that you cannot add a clas
 </main>
 ```
 
+If you need to apply these styles to heading elements that you cannot add a class to but want more tightly scoped styles than the default `.cads-prose` class you can add a `.cads-prose-scoped` class instead. This is useful when you are embedding lots of rich components within a prose block and you want to more narrowly scope the effects of the cascade.
```
davidsauntson, line 88: I think we should be specific here and say it will be scoped using `>`, ie to the direct children of the element it's applied to.  It's implied in the example below but would be nice to be super clear.

scss/5-objects/_prose.scss
```diff
@@ -70,6 +66,39 @@
...
+
+/**
+ * Prose styles (scoped)
+ *
+ * This applies the same styles as the default cads-prose class but scoped using
+ * a child combinator selector to only apply the styles to immediate children.
+ * This is useful when you are embedding lots of rich components within a prose
+ * block and you want to more narrowly scope the effects of the cascade.
+ */
+.cads-prose-scoped {
```
davidsauntson, line 96: Nitpick so feel free to ignore - wonder if it's worth having a more descriptive name like one of these terrible options, but better:
- `cads-prose-on-children`
- `cads-prose-scoped-to-children`
or smth

davidsauntson (approved):
Nice! 🔭

Some minor thoughts but nothing blocking.

scss/5-objects/_prose.scss
davidsauntson: LOL 👏🏻

PR #2297, opened 2022-08-25
Rationalise colour names amends

davidrapson:
Had a go at using https://github.com/citizensadvice/design-system/pull/1863 to re-theme the intranet. It worked really well and allowed me to remove a bunch of previous override and configure fewer variables. Spotted two small details along the way:

1. The footer was still using an old colour variable name which broke when switching to the new variables
2. Link state ordering in the header caused some undesirable conflicts when theming. See the commit message in https://github.com/citizensadvice/design-system/commit/f5ecdee7d254102bfd31431c4a5c8272d06d90d0 for more detail


marianayovcheva (approved):
Thanks! 🙌

PR #2298, opened 2022-08-25
Rationalise colour names docs theme (amends / docs theme)

davidrapson:
A couple more suggested amends for https://github.com/citizensadvice/design-system/pull/1863

1. Prefixes the `$use-legacy-color-names` variable with `cads-` for consistency
2. Updates the docs theme settings to use all new variable names


scss/1-settings/_color.scss
```diff
@@ -11,7 +11,7 @@
   @return $color;
 }
 
-$use-legacy-color-names: true !default;
+$cads-use-legacy-color-names: true !default;
```
davidrapson, line 14: @marianayovcheva I wasn't 100% sure what this setting was meant to do. Might be useful to write a sass test along the lines of https://github.com/citizensadvice/design-system/blob/main/scss/1-settings/__tests__/settings.test.js to document what the expected behaviour of this flag is.

davidsauntson (commented):
LGTM but would like @marianayovcheva  to have the final say!

scss/1-settings/_color.scss
marianayovcheva: > I wasn't 100% sure what this setting was meant to do. 

I guess it's worth me adding a comment explaining what the variable does.

> Wonder if we should write a sass test

I think adding tests makes sense but I need advice what to test in our case - that the colour of a component is the same in both cases?

marianayovcheva: I'll probably address those as part of the main PR

marianayovcheva:
LGMT as well! 👨‍🎨

marianayovcheva (approved):

PR #2299, opened 2022-08-26
feat: adding callout component to the new docs site

marianayovcheva:
Added the callout component to the new docs site. I had to convert the haml file to erb


davidrapson (approved):
Nice!

PR #2304, opened 2022-08-30
Add new ComponentExampleSource component

davidrapson:
1. **Adds a new ComponentExampleSource component**. Serves a similar purpose to ComponentExample but just renders the ERB source code. Useful for showing inline code samples that reuse our existing code samples.
2. **Share code sample code between components**. Both the source and full example components use the same underlying data. Provide a presenter-ish object which can also find an example based on the site, category, and slug. Provides a shared interface for accessing an example with pre-transformed and formatted source.

The upshot is you can call this from your component docs:

```
<%= render(Shared::ComponentExampleSource.new(:asset_hyperlink, :default)) %>
```

And it will render just the inline source for the provided example

### DIFF
--- a/design-system-docs/src/_component_docs/button.md
+++ b/design-system-docs/src/_component_docs/button.md
@@ -58,10 +58,7 @@ Do not use anchor tags with an empty href when you want a button `<a href=â€#â€
 
 If you are using the `citizens_advice_components` gem, you can call the component from within a template using:
 
-```haml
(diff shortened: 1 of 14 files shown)


design-system-docs/src/_component_docs/button.md
```diff
@@ -71,20 +71,4 @@ If you are using the `citizens_advice_components` gem, you can call the componen
```
marianayovcheva, line 71: Should the button component use the new example source in the 'Using with Rails' section? (L61)

marianayovcheva (approved):
This is super useful!

PR #2305, opened 2022-08-30
Migrate more component docs pages

davidrapson:
Using the new ComponentExampleSource component from https://github.com/citizensadvice/design-system/pull/2304 this PR migrates over a few of the simpler component docs pages.

1. Notice banner (converts component to ERB in the process)
2. Success message
3. Page review
4. Pagination

### DIFF
--- a/design-system-docs/frontend/styles/_code.scss
+++ b/design-system-docs/frontend/styles/_code.scss
@@ -21,21 +21,12 @@ pre code {
   background-color: $cads-palette__white;
 }
 
-.highlight {
+pre.highlight {
   margin-bottom: $cads-spacing-6;
   border: $cads-border-width-small $cads-language__border-colour solid;
   padding: $cads-spacing-4;
 
-  &--with-button {
-    margin: 0;
-  }
-
   button {
     font-family: $cads-font-family;
-  }
-
-  > .highlight {
-    border: none;
-    padding: 0;
   }
 }
--- a/design-system-docs/src/_component_docs/pagination.md
+++ b/design-system-docs/src/_component_docs/pagination.md
@@ -1,0 +1,21 @@
+---
+title: Pagination
+---
+
+<%= render(Shared::ComponentExample.new(:pagination, :default)) %>
+
+## Accessibility requirements
+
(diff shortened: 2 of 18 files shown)


design-system-docs/frontend/styles/_code.scss
```diff
@@ -26,6 +26,10 @@ pre code {
   border: $cads-border-width-small $cads-language__border-colour solid;
   padding: $cads-spacing-4;
 
+  &--with-button {
```
marianayovcheva, line 29: Maybe we need to scope those to the component-example. When looking at the 'Using with Rails' sections, there is no margin above 'View Component Option' header that follows (https://630dc2463174f015e06e24be--ca-design-system.netlify.app/components/breadcrumbs/). E.g:
image

design-system-docs/src/_component_docs/pagination.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+<%= render(Shared::ComponentExample.new(:pagination, :default)) %>
+
+## Accessibility requirements
+
+- Pagination is wrapped in a `<nav>` with `role="navigation"` and an appropriate `aria-label` e.g. `aria-label="Pagination navigation"`. You may wish to customise this based on the context it is used in, e.g. "Search pagination"
+- Each link has an appropriate `aria-label` e.g. "Go to first page", or "Go to page 3".
+- The current page has an `aria-label` of `aria-label="Current page, page 5"` with `aria-current="page"`
+
```
marianayovcheva, line 12: I noticed when comparing the docs that there used to be a paging-info cads class (I think it's not really a component): https://citizensadvice.github.io/design-system/?path=/docs/components-pagination--example#paging-info. I think it makes sense not to add it here but wondering if we should mention it elsewhere?

design-system-docs/frontend/styles/_code.scss
davidrapson: Oh yeah, that's why I removed this originally 🙈. Let's try that!

marianayovcheva (approved):
Nice! I've added a question and a comment but fairly minor - other than that good to go!

design-system-docs/src/_component_docs/pagination.md
davidrapson: I removed it for the time being as @davidsauntson and I realised a while back that it's kinda unused and we completely override it in public-website. So maybe even something we can remove. It's literally just a paragraph which is kinda confusing.

PR #2308, opened 2022-08-30
Migrate SVG Icons page to new component docs

davidrapson:
As the foundations icons page now uses our SVG icon components I thought it made sense to merge the two icons pages together. Previously the component version referenced the foundations page quite a bit anyway.


davidsauntson (approved):

davidrapson:
Just fixed up a small reference to the old storybook docs.

PR #2309, opened 2022-08-30
Convert breadcrumbs to ERB

davidrapson:
Convert breadcrumbs component template to ERB, pushes some of the rendering logic into the component.


davidsauntson (approved):
Nice to have the rendering logic in the class 🙌🏻

PR #2310, opened 2022-08-30
Add new cads-grid-col-width function

davidrapson:
Adds a new cads-grid-col-width function which allows us to get the
percentage width for a given number of grid columns.

```scss
.my-component {
  max-width: cads-grid-col-width(6);
}
```

Throws an error if you request a column count greater than the maximum.

### DIFF
--- a/scss/2-tools/__tests__/grid.test.js
+++ b/scss/2-tools/__tests__/grid.test.js
@@ -1,0 +1,44 @@
+/* eslint-env jest */
+const sass = require('sass');
+
+describe('cads-grid-col-width', () => {
+  test.each([
+    [1, '8.3333333333%'],
+    [2, '16.6666666667%'],
+    [3, '25%'],
+    [4, '33.3333333333%'],
+    [5, '41.6666666667%'],
+    [6, '50%'],
+    [7, '58.3333333333%'],
+    [8, '66.6666666667%'],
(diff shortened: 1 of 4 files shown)


scss/2-tools/_grid.scss
```diff
@@ -0,0 +1,13 @@
+@use 'sass:math';
```
davidrapson, line 1: As nice as the ITCSS structure is generally I don't _love_ having the settings / helpers / and styles for grids in three different files. Following the existing path here though.

scss/2-tools/__tests__/grid.test.js
```diff
@@ -0,0 +1,44 @@
...
+  ])('outputs width for %i columns', (cols, expected) => {
+    const data = `
+      @use 'scss/2-tools/grid';
+
+      .foo {
+        width: grid.cads-grid-col-width(${cols});
+      }
+    `;
+
+    const result = sass.renderSync({ data });
```
davidrapson, line 27: `renderSync` is technically deprecated in favour of a new `compileString` API, couldn't work out how to get it to respect `@use` statements so have kept this the same as our existing tests for now and then raised https://github.com/citizensadvice/design-system/issues/2311 to look at this properly.

davidsauntson (approved):
Nice! 🥨

PR #2313, opened 2022-08-31
Migrate search component docs

davidrapson:
Not to toot my own 🎺 but the new docs generator makes this process _so_ much quicker.


marianayovcheva (approved):
🎺

PR #2314, opened 2022-08-31
Migrate on this page component docs

davidrapson:
Migrate on this page component docs. Split the component arguments out into two tables to better reflect the fact some are top-level the rest are for a slot.


marianayovcheva:
> Split the component arguments out into two tables to better reflect the fact some are top-level the rest are for a slot.

I quite like this approach. Reckon we should go back and do this for the other components that use slots. I'm happy to do it

marianayovcheva (approved):

PR #2315, opened 2022-08-31
Migrate header docs

davidrapson:
Tried a bigger one as I'm having a productive spell on migrating docs pages. Actual docs page wasn't too bad but spotted a couple of changes along the way.

**Fix a small bug with the header javascript**

If you initialise the header javascript when there is no search present it causes an error (as that's all it's used for), add a guard clause to the script.

**Split docs assets into core and theme**

When rendering the header examples in the design system docs the theme styles were overriding the example header styles.

Split the assets into two so we have:

1. A core set of assets that only load the design system, used on isolated component sample pages (and via iframe)
2. A second set of theme assets which is what we actually load on the docs pages

**Migrate header docs to new docs website**

Keeps most code samples inline rather than in examples as they are partial examples only to illustrate how slots work. Tiny bit of rewriting to avoid referencing sample pages that don't exist (yet?)


design-system-docs/config/webpack.defaults.js
```diff
@@ -17,6 +17,7 @@ const { WebpackManifestPlugin } = require('webpack-manifest-plugin');
 
 const entry = {
   main: path.resolve(rootDir, 'frontend', 'javascript', 'index.js'),
+  theme: path.resolve(rootDir, 'frontend', 'javascript', 'theme.js'),
```
davidrapson, line 20: The way these entrypoints work is super confusing. Would love to migrate away from webpack to something simpler eventually.

marianayovcheva (approved):
Nice!

PR #2322, opened 2022-09-01
Footer: Fix borders when using minimal footer

davidrapson:
When using the minimal version of the footer or just the feedback link (i.e. no navigation links) there are unwanted borders present.

1. Update the styles to place the border on the navigation links, not the copyright information
2. Add a check to the footer component to not render the nav if no links are provided
3. Add a preview for the version of the footer with only a feedback link
4. Update visual regression tests

**Before**

Screenshot 2022-09-01 at 10 16 52

Screenshot 2022-09-01 at 10 16 58

**After**

Screenshot 2022-09-01 at 10 21 23

Screenshot 2022-09-01 at 10 21 18


marianayovcheva (approved):
Nice!

PR #2325, opened 2022-09-02
Publish docs on main build

davidrapson:
Currently our docs build workflow only publishes drafts and on release versions.

Whilst we are still working on the foundations of the new docs website it would be nice to have a continuously deployed version of the latest changes. Added a new workflow that publishes on main builds which we can remove later if we want to stick to release-only builds.

### DIFF
--- a/.github/workflows/docs-workflow-main.yml
+++ b/.github/workflows/docs-workflow-main.yml
@@ -1,0 +1,46 @@
+name: Design System Docs (main)
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish:
+    name: Publish release to Netlify
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+
+      # We link the @citizensadvice/design-system package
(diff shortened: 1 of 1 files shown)


marianayovcheva:
I've had a look - I understand what it's doing and can see that it re-uses logic that's in [docs-workflow.yml](https://github.com/citizensadvice/design-system/blob/main/.github/workflows/docs-workflow.yml#L65) so it's probably good to go but I think it would be best for @davidsauntson to review this one as I'm not so confident with workflows yet.

davidsauntson (approved):
Looks fine - couple of thoughts that you can ignore 😄 
- `./netlify.toml` is empty so I think we can delete it now (I should have done this in https://github.com/citizensadvice/design-system/pull/1778 🙈)
- do we want to publish on every push? This would include all the dependabot updates etc.  I don't think this poses a problem in terms of quotas or anything

davidrapson:
> do we want to publish on every push? This would include all the dependabot updates etc. I don't think this poses a problem in terms of quotas or anything

I'll check this, i reckon we need to ignore dependabot because it won't have permissions?

davidrapson:
Just confirming this ran fine when merging this and then correctly skipped when merging a dependabot PR

image

PR #2326, opened 2022-09-02
Migrate text input and textarea component docs

davidrapson:
Migrate text input and textarea component docs to new website.

Renamed the "Input" docs to "Text input" to reflect that's the actual component name used here.

### DIFF
--- a/design-system-docs/src/_component_examples/_text_input/fixed_widths.erb
+++ b/design-system-docs/src/_component_examples/_text_input/fixed_widths.erb
@@ -1,0 +1,31 @@
+---
+title: fixed widths
+---
+
+<%= render(CitizensAdviceComponents::TextInput.new(
+  width: :two_chars,
+  name: "example-input-fixed-width-two_chars",
+  label: "Example input", type: :text,
+  options: { hint: "This input is two_chars wide" }
+)) %>
+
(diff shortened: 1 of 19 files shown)


design-system-docs/src/_component_docs/input.md
```diff
@@ -0,0 +1,21 @@
+---
```
marianayovcheva, line 1: I think this file can be deleted as it's replaced by Text input

design-system-docs/src/_component_examples/_text_input/fixed_widths.erb
```diff
@@ -0,0 +1,31 @@
...
+---
+
+<%= render(CitizensAdviceComponents::TextInput.new(
+  width: :two_chars,
+  name: "example-input-fixed-width-two_chars",
+  label: "Example input", type: :text,
+  options: { hint: "This input is two_chars wide" }
+)) %>
+
+<%= render(CitizensAdviceComponents::TextInput.new(
```
marianayovcheva, line 12: Nitpick but this input and the one below (on line 19) are the same width. I think this one should be `four_chars`

marianayovcheva (approved):
Looks good! I've noted a couple of small things that need changing before merge but otherwise good to go

PR #2328, opened 2022-09-02
Migrate disclosure docs

davidrapson:
Migrate disclosure docs ➕ ➖


marianayovcheva:
A side thing that is not related to the code. Personally, I wonder if there is a way to visually distinguish or separate the example disclosure (Filter) from the code disclosures (HTML code, ERB code) as they are the same component: 
image
Probably a question for James.

marianayovcheva (approved):
➕

davidrapson:
> A side thing that is not related to the code. Personally, I wonder if there is a way to visually distinguish or separate the example disclosure (Filter) from the code disclosures (HTML code, ERB code) as they are the same component:


Yeah we should ask James. I originally added borders around the controls but we removed those in https://github.com/citizensadvice/design-system/pull/2153 to match the original designs, but we should revisit that as it _does_ look weird here.

davidrapson:
This is what I had originally for context:

image

PR #2329, opened 2022-09-02
Docs tidy up

davidrapson:
Small bits of tidy up after migrating most of the component docs:

- Greedy nav: Use theme colour
- Preload fonts in docs website
- Disable telephone format detection
- Fix accessibility warning with notice banner outside main landmark


marianayovcheva (approved):
🧹

PR #2335, opened 2022-09-05
Upgrade size-limit dependencies

davidrapson:
Noticed that size-limit has stopped working for our JS files (always reporting 0 B).

Uses this as an excuse to upgrade size-limit to the latest version. The latest version uses esbuild now for size-limit/preset small-lib which doesn't make sense for our css files. All we really care about is a file-size check so switches the setup to use the basic file plugin and updates the glob for our JS files to include everything, not just top-level files.

```
> size-limit


  lib/lib.css
  Size limit: 16 kB
  Size:       12.56 kB gzipped

  lib/error-summary.js, lib/header.js, lib/targeted-content.js, lib/disclosure/disclosure.js, lib/greedy-nav/Config.js, lib/greedy-nav/GreedyNav.js, lib/greedy-nav/index.js, lib/on-this-page/on-this-page.js, lib/utils/show-hide.js
  Size limit: 16 kB
  Size:       12.23 kB gzipped
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@babel/plugin-transform-runtime": "^7.14.5",
     "@babel/preset-env": "^7.16.11",
     "@babel/preset-typescript": "^7.16.7",
-    "@size-limit/preset-small-lib": "^6.0.4",
+    "@size-limit/file": "^8.0.1",
     "@storybook/addon-docs": "^6.3.7",
     "@storybook/addon-postcss": "^2.0.0",
     "@storybook/addon-viewport": "^6.3.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -90,12 +90,11 @@
   "size-limit": [
     {
       "path": "lib/lib.css",
-      "limit": "16 kB",
-      "webpack": false
```
marianayovcheva, line 94: Out of interest (and because I couldn't wrap my head around the [size-limit readme](https://github.com/ai/size-limit#size-limit-) explanation) - why did we have `webpack: false` for the css to begin with and why can it be removed now?

marianayovcheva (approved):

package.json
davidrapson: The previous preset default would seemingly load all your JS into a blank webpack project as a size test. That didn't work for CSS so they told you to disable it. We only ever cared about the file-size check so I'm including a simpler preset now which just checks gzipped size.

PR #2347, opened 2022-09-05
Convert search field to ERB

davidrapson:
Converts search field to ERB and adds an `autocomplete` attribute as recommended by erblint.


marianayovcheva (approved):
🔍

PR #2350, opened 2022-09-05
More docs clean up

davidrapson:
- Don't preload icon font fix 404
- Add favicon
- Style logo through css not inline fill
- Basic home layout, don't show breadcrumb
- Push breadcrumbs into layout
- Make default code samples focusable


design-system-docs/frontend/styles/_layout.scss
```diff
@@ -1,9 +1,21 @@
...
-// Listings
+// Layout
 // ============================================================================
 
+.home-intro {
+  padding: $cads-spacing-7 0;
+  margin-bottom: $cads-spacing-7;
+  border-bottom: $cads-border-width-small $cads-language__border-colour solid;
+
+  .home-intro__title {
```
davidsauntson, line 10: Does this need to be nested?

design-system-docs/src/_layouts/default.erb
```diff
@@ -24,6 +23,7 @@
         { url: "/components", title: "Components" }
       ]
     ) %>
+    <%= render Shared::AppBreadcrumbs.new(resource) %>
```
davidsauntson, line 26: Might be worth putting the alpha notice banner here too, so it's always everywhere?

davidsauntson (approved):
🙌🏻 OMG it's actually happeningggg 🥳 

Two minor questions, nothing blocking

PR #2356, opened 2022-09-06
Remove moment as a dependency

davidrapson:
We currently use moment to format a single date in our changelog script. We can reasonably use `Intl.DateTimeFormat` here instead. Updates the script to:

- Use `Intl.DateTimeFormat` in place of moment
- Simplify the formatting of heading so there's less special case formatting and so that heading levels aren't skipped awkwardly.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -5,7 +5,6 @@ const path = require('path');
 const { prompt } = require('inquirer');
 const { execSync, spawnSync } = require('child_process');
 const semver = require('semver');
-const moment = require('moment');
 const { checkRepoStatus, ok, error, showError } = require('./releaseHelpers');
 
 const log = console.log; // eslint-disable-line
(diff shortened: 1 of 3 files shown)


scripts/release.js
```diff
@@ -127,9 +132,7 @@ prompt([
         const changelog = fs.readFileSync(changelogPath, 'utf8');
 
         // Add the new entry
-        const changelogEntry = `## <sub>v${newVersion}</sub>\n\n#### ${moment().format(
-          '_MMM. D, YYYY_'
-        )}`;
+        const changelogEntry = `## v${newVersion}\n\n### ${formattedDate()}`;
```
davidsauntson, line 135: I'm a bit lost, contextually, so I might be being stupid.  This is markdown tho, right? How come there are 2 `#` before and 3 `#` after?  Not that it is really that important!

davidsauntson (approved):
<whitney houston gif>

scripts/release.js
davidsauntson: OMG pls ignore 🙈

davidrapson: Not stupid at all, I was _also_ lost reading this code. Guessing you worked out that this is a h2 and h3. It was super confusing before going 2 to 4.

PR #2357, opened 2022-09-07
Feat/new docs checkbox group

marianayovcheva:
Added checkbox group and single checkbox to the new docs site.
Converted the single checkbox to erb


engine/app/components/citizens_advice_components/checkbox_single.erb
```diff
@@ -0,0 +1,17 @@
...
+    <% end %>
+    <div class="cads-form-group__item">
+      <%= tag.input(class:"cads-form-group__input", **checkbox.attributes(name, error: error?)) %>
+      <label class="cads-form-group__label"
+             for="<%= label_id %>">
+        <%= checkbox.label %>
+      </label>
+    </div>
+  </div>
+</div>
```
marianayovcheva, line 17: I must admit, I also tried to convert `form_group.html.haml` to erb as part of this PR, however I was having issues with the classes not appearing as a string What I had was `<fieldset class="<%= fieldset_classes %>">` which resulted in:
Screenshot 2022-09-07 at 16 40 22
I assume there must me an easy way to make those classes in erb so any suggestions are welcome and I'll pick them up in a new PR

davidrapson: Cool to leave that for another PR, there's probably a couple ways:

1. Call join on the array: `<fieldset class="<%= fieldset_classes.join(" ") %>">`
2. Use a tag helper, which accepts an array: `tag.fieldset(classes: fieldset_classes)`
3. Rails 6.1 introduces a class_names helper which maybe in future we could pull in to make working with classes easier (https://github.com/rails/rails/pull/37918). We still need to support Rails 6.0 though until public-website is upgraded.

davidrapson (approved):
🥳 WOOOO!!!!

PR #2372, opened 2022-09-12
SCSS/Stylelint cleanup

davidrapson:
- Consistent visually hidden styles
- Add a stylelint rule for cads- prefixed classes
- Clean up stylelint config
- Add warnings for variable and placeholder names
- Add implicit rules for BEM stylelint plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "docs:build": "build-storybook -o docs",
     "docs:serve": "http-server -p 6006 docs",
     "lint:format": "prettier --check .",
-    "lint:css": "stylelint \"./scss/**/*.scss\" \"./styleguide/**/*.scss\"",
+    "lint:css": "stylelint \"./scss/**/*.scss\"",
     "lint:es": "eslint . --ext .js,.ts",
     "lint": "npm-run-all lint:*",
     "test": "jest",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "docs:build": "build-storybook -o docs",
     "docs:serve": "http-server -p 6006 docs",
     "lint:format": "prettier --check .",
-    "lint:css": "stylelint \"./scss/**/*.scss\" \"./styleguide/**/*.scss\"",
```
davidrapson, line 31: Dropping this as we're about to remove storybook so we don't really care about linting these styles, and most core rules don't apply here anyway.

scss/7-utilities/_visibility.scss
```diff
@@ -32,50 +37,102 @@
   }
 }
 
-@each $name, $breakpoint in breakpoint-settings.$cads-breakpoints {
```
marianayovcheva, line 35: I don't have strong feelings about this but out of curiosity - why did you decide to change the loop into separate classes?

marianayovcheva (approved):

scss/7-utilities/_visibility.scss
davidrapson: The rationale is in this commit: https://github.com/citizensadvice/design-system/pull/2372/commits/d84ca41b5521ea39ff9ecc985e18e41f296c0764

(Short version, it's to play nice with the implicit BEM linter rules so it can pick these classes up. One of the issues with generated CSS)

PR #2373, opened 2022-09-12
Rails engine: Only load parts of Rails we need

davidrapson:
The engine currently requires the whole of rails. We only really need actionpack and railties here. Limiting to the required dependencies simplifies things when running outside of rails as well as giving us a warning if we start relying on larger parts of rails.


davidsauntson (approved):
🛤️

PR #2374, opened 2022-09-13
Remove storybook

davidrapson:
## Background

🗑️  Now that all component documentation has been migrated over the to the new docs website and website is live we are ready to remove storybook completely.

## Changes

**Remove storybook**

The big one. Removes all storybook related source code and npm dependencies.

**Write examples to a top-level directory**

These were originally added as a way of integrating with storybook but they have proven useful as a simple snapshot test and as reference for changes over time. Moves the examples to a top-level directory so we can continue to support them for now.

**Replace docs build job in CI workflow**

This original build job was used to run a test build of the node package and storybook docs. With the removal of storybook we no longer need this. The node package itself is build several times as part of other jobs.

Replace this with a new job which builds the new documentation website as a smoke test. This is in addition to any builds run when adding the "docs preview" label to a pull request.

## Dependencies

- [ ] Remove `build` job from required checks and add `docs-build`. Required before this PR can be merged.


davidsauntson (approved):
💥

PR #2375, opened 2022-09-14
Wrap form examples in a form context

davidrapson:
Our form component examples weren't wrapped in a `<form>` element. A form context affects how screen readers behave. Rather than adding a form manually to every component I've added an option to the front-matter to optionally add the form in, also set novalidate as we do not rely on html validation for our forms so we don't want this behaviour.

Whilst I'm here makes a small change to the standalone component example styles to add a bit of padding to the body aside from global components (this is the same approach we use in the demo app)


davidrapson:
Confirmed with @rapaccinim that this fixes the unexpected screen reader behaviour on form examples.

PR #2380, opened 2022-09-14
Docs design refinement

davidrapson:
- Move example grid styles into standalone styles
- Fix spacing and sizing example tables
- Update component examples to match latest designs
- Disable iframeResizer logging
- Make nav link colours customisable
- Update docs header colours to match designs

### DIFF
--- a/scss/1-settings/_colour-language.scss
+++ b/scss/1-settings/_colour-language.scss
@@ -29,9 +29,13 @@ $cads-language__warning-colour: palette.$cads-palette__red !default;
 $cads-language__success-colour: palette.$cads-palette__green !default;
 
 // Navigation colours
+$cads-language__nav-colour: palette.$cads-palette__white !default;
 $cads-language__nav-background-colour: $cads-language__brand-primary !default;
 $cads-language__nav-hover-background-colour: palette.$cads-palette__heritage-blue-dark !default;
 $cads-language__nav-hover-border-colour: palette.$cads-palette__heritage-blue-dark !default;
+$cads-language__nav-focus-colour: $cads-language__link-colour !default;
+$cads-language__nav-focus-background-colour: palette.$cads-palette__white !default;
+$cads-language__nav-focus-border-colour: $cads-language__focus-colour !default;
 
 // Header colours
 $cads-header__background-colour: palette.$cads-palette__white !default;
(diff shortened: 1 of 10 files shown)


davidrapson:
Just to say: James proposed these changes and is happy with them

scss/1-settings/_colour-language.scss
```diff
@@ -29,9 +29,13 @@ $cads-language__warning-colour: palette.$cads-palette__red !default;
 $cads-language__success-colour: palette.$cads-palette__green !default;
 
 // Navigation colours
+$cads-language__nav-colour: palette.$cads-palette__white !default;
 $cads-language__nav-background-colour: $cads-language__brand-primary !default;
 $cads-language__nav-hover-background-colour: palette.$cads-palette__heritage-blue-dark !default;
 $cads-language__nav-hover-border-colour: palette.$cads-palette__heritage-blue-dark !default;
+$cads-language__nav-focus-colour: $cads-language__link-colour !default;
```
marianayovcheva, line 36: I'm slightly nervous about the focus-colour being the link-colour. I'm not sure that we will remember to update the colour for advisernet and intranet. I don't think the vrts in the public-website side will catch it.
The rest of the new colours look good to me because they are quire consistent throughout products.

davidrapson: oh yeah, maybe this should use the palette colour instead

davidrapson: Moved this to use the palette colour, but double checked and this doesn't have much affect on intranet and advisernet as we don't override these values. Makes sense to use the palette colour directly still though 👍

marianayovcheva: Won't using a palette colour still display the nav colour on focus as blue in intranet and advisernet?

davidrapson: Yep, but that's what we want. We don't override the colour for those products.

marianayovcheva (approved):
🙌

scss/1-settings/_colour-language.scss
davidrapson: For our future selves: We got to the bottom of this. The focus colour was previously using the _brand_ language colour which _does_ get overridden per-product typically. Have updated it to use that colour for backwards compatibility.

PR #2411, opened 2022-10-05
Proposing different button spacing on form pages

marianayovcheva:
## Why? 
As we are introducing more pages which are either 'one question per page' form (i.e decision trees) or full form on a page, there is often the issue of the spacing between the last button and the footer being bigger because the button has margin bottom. E.g: 
image
I decided to introduce a fix to the the design system as there are other tools which can also use it (e.g smart metering tool or future tools) but if it's too specific, I'm happy to just move it to the public-website directory.

## What I tried
I initially tried removing the margin bottom from all buttons and adding margin top on the form below (with a special class). You can see it in this commit: https://github.com/citizensadvice/design-system/commit/2dd7586ea367c984feca3671735bd8d91ad24f47. After auditing where buttons are used in the public-website I decided that this approach is too drastic and would make the button style overriding more confusing. 

## What worked
I have then introduced a new class for `.cads-page-form`. Any different name suggestions are welcome! As we are introducing more forms into the website, we can gradually add more useful elements to it. The idea is to have the main form with questions and submit button as `.cads-page-from` and the following form containing only a button as `.cads-page-form cads-page-form--single-button`.


scss/6-components/forms/page-form.scss
```diff
@@ -0,0 +1,10 @@
+/** @define page-form; weak */
+.cads-page-form {
+  .cads-button {
+    margin-bottom: $cads-spacing-1;
+  }
+
+  &--single-button {
+    margin-top: $cads-spacing-4;
+  }
+}
```
davidrapson, line 10: I really like the principle behind this, but rather than making this a page-level class I'd be tempted to treat this as a more localised button group. Perhaps `.cads-form-actions` to signify this is a group of buttons as part of a form? Does that match with what you are thinking of?

marianayovcheva: I agree that a localised class would be better but not sure how to implement it here. I think the bit I was struggling with was that often if we have two buttons that are on the same page and they look like they should be part of the same group (e.g next/previous), they are functionally part of different forms. For example in decision trees the 'Next' button is part of the main question form, while the 'Previous' button is part of another form. Similar thing happens on the smart meter tool. I'm not sure how to address the spacing if I'm using a localised class in those cases. The only thing I can think of is trying out `first-of-type`.

davidrapson: Ah I see that's interesting. I think first of type selectors would be fine here, as long as we can broadly wrap the `cads-form-actions` around the area where the buttons are?

marianayovcheva: I've now tried with last of type instead. It didn't work in this case because of how the buttons are nested - i.e if I have 
something like:
scss:
```
.cads-form-actions {
  button:last-of-type {
    margin-bottom: $cads-spacing-1;
  }
}
```
html:
```
.cads-form-actions
  <form 1>
    <first button>
  <form 2>
    <second button>
```
it would think both buttons are last of type

davidrapson: In which case I'd either go with what you had originally (treating it as a page-level class) or go even _more_ targeted and add a little utility class we can add to the specific buttons that we need. But I suspect the second option is less useful. So perhaps ignore me here!

davidrapson (approved):
Approving as a like the general thinking and could go either way on the specifics of how we name this so go with your gut here.

PR #2415, opened 2022-10-06
Use Ruby 3 by default

davidrapson:
In order to spot compatibility issues sooner and to bring things up to date more generally let's set the default ruby version to 3.0.4.

In the process, fixes a syntax issue in the design system docs.

### DIFF
--- a/design-system-docs/src/_component_examples/_checkbox/default.erb
+++ b/design-system-docs/src/_component_examples/_checkbox/default.erb
@@ -3,5 +3,5 @@ title: default
 ---
 
 <%=  render CitizensAdviceComponents::CheckboxSingle.new(name: "a-single-checkbox") do |c|
-  c.checkbox({ label: "I agree to the great terms and conditions", value: "agreed-to-tcs" })
+  c.checkbox(label: "I agree to the great terms and conditions", value: "agreed-to-tcs")
 end %>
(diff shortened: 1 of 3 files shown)


design-system-docs/src/_component_examples/_checkbox/default.erb
```diff
@@ -3,5 +3,5 @@ title: default
 ---
 
 <%=  render CitizensAdviceComponents::CheckboxSingle.new(name: "a-single-checkbox") do |c|
-  c.checkbox({ label: "I agree to the great terms and conditions", value: "agreed-to-tcs" })
+  c.checkbox(label: "I agree to the great terms and conditions", value: "agreed-to-tcs")
```
davidrapson, line 6: See https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

Ruby 3 separates how positional and keyword arguments are handled. Passing in arguments to the single checkbox as a hash relied on a quirk of ruby 2 where hashes were converted to positional arguments.

Now we either need to pass them explicitly or use the ** double splat operator. It's simplest to use the named parameters directly here.

marianayovcheva (approved):
LGTM! (with the caveat that this is the first major Ruby upgrade I've seen)

davidrapson:
> with the caveat that this is the first major Ruby upgrade I've seen

We already test the design system components against Ruby 3 so thankfully this is only changing the defaults.

PR #2416, opened 2022-10-06
Fix label being assigned as checkbox attribute in Ruby 3

davidrapson:
When running against ruby 3 and passing extra options to a checkbox via `c.checkbox(label: "Label", value: "value", extra: "example")` the previous `**additional_attributes` keyword args would include the named label which we don't want.

This is down to the `super` call in the initialise method. Calls to super in subclasses of `ViewComponent::Base` aren't needed by default and can change the behaviour of `**kwargs` so we exclude component classes from this rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 **New**
 
+- Fix label being assigned as checkbox attribute in Ruby 3+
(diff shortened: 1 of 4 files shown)


mrdaniellewis (approved):
Looks good to me

PR #2417, opened 2022-10-06
Build new docs site into github pages directory

davidrapson:
Whilst we determine if we are in a position to move fully over to Netlify we can unblock the launch of the new docs site by publishing directly into the existing github pages folder for the time being.

Once we decide to move over fully we can redirect the old pages, but this let's us replace things now using the tools we currently have.


davidrapson:
We don't need this anymore

Issue #1962, opened 2022-04-22
Explore simpler build tools for the demo app


davidrapson:
We currently use webpacker for the demo app. We know that webpacker can be a pain to work with and is far more complex than we need for the basic demo app.

We should consider migrating to something simpler like cssbundling-rails and jsbundling-rails. This would also let us explore build options for future Rails 7 projects.

PR #2421, opened 2022-10-10
Upgrade demo app to Rails 7

davidrapson:
Upgrades the design system demo app to use Rails 7 and to build assets with cssbundling-rails and jsbundling-rails instead of webpacker.

- Update Rails gems and generated config
- Removes spring (no longer bundled by default with Rails 7)
- Adds `citizens-advice-style` gem for consistency, but also to help format generated upgrade files
- Uses yarn for the demo app. This is assumed by css/jsbundling-rails as Rails ships with yarn by default.
- Loads design system assets relatively for the same reason as https://github.com/citizensadvice/design-system/pull/2206
- Use cssbundling-rails for CSS with sass
- Use jsbundling-rails for JS with esbuild
- Remove all webpacker config

Closes https://github.com/citizensadvice/design-system/issues/1962


davidrapson:
@marianayovcheva Just popped this back into draft status, would be great to run through this with you to get your take on the approach first. No rush to do that though.

demo/config/initializers/assets.rb
```diff
@@ -0,0 +1,17 @@
...
+# Be sure to restart your server when you modify this file.
+
+# Version of your assets, change this if you want to expire all your assets.
+Rails.application.config.assets.version = "1.0"
+
+# Add additional assets to the asset load path.
+# Rails.application.config.assets.paths << Emoji.images_path
+
+# Add design system assets to the load path (for fonts)
+Rails.application.config.assets.paths << Rails.root.join("../assets")
```
davidsauntson, line 12: <3

davidsauntson (approved):
This is great! So much simpler without webpack. 👏🏻 

~All good running locally, worth sticking a `docs preview` label on it?~ No, because it's the demo site 🤦🏻

PR #2423, opened 2022-10-10
Design system docs launch prep

davidrapson:
A few last bits of prep before we launch the new docs website:

- Don't build storybook docs on release — We won't need this anymore and new docs builds are handled via CI
- Add minimal docs roadmap page, just enough to link to from the notice banner.
- Fix some wording and typos based on the documentation review


marianayovcheva (approved):
Exciting! Makes sense to me. As it's the last bit before the launch, probably worth for @davidsauntson to have a look as well

davidsauntson (approved):
🙌🏻

PR #2438, opened 2022-10-17
Feat/small content tweaks

marianayovcheva:
Description coming soon


davidsauntson (approved):
Not sure if this is ready yet but looks ok - will be worth putting the `docs preview` label on though to check

PR #2439, opened 2022-10-20
feat: replaced the require attribute with aria-required

marianayovcheva:
Content designers noticed that when a user hovers on a textarea in Chrome, the default 'Please fill in this field' box appears and it doesn't match our custom error message for that fireld. 
image
This is coming from fields using the `required` attribute. For the public website forms we use the `novalidate` attribute and custom error messages so as far as I'm aware, required is useful for screen readers atm (please correct me if that's not the only use case here). 
I've checked other services, e.g:
- NHS login https://access.login.nhs.uk/enter-email
- GOV register to vote https://www.gov.uk/register-to-vote

Those services don't use `required` for mandatory fields. NHS uses [aria-required](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required) which I think would be useful to add in our case. This is to replicate the current screen reader behaviour.

Another question is whether we need `aria-required` at all as we sometimes have mixed forms - i.e text inputs and radio buttons being part of the same form. Radio groups don't have 'required' pronounced even when they are mandatory. If we remove `aria-required` completely my only concern is whether the assumption that the users will know which fields are mandatory because they don't have `(optional)` is enough.


davidsauntson (commented):
Interesting!  You're correct about why the `required` attribute is there - for screen readers since we don't use HTML5 native validation.

I'm happy with this since it seems to work with Chrome and voiceover and for the NHS also.  I couldn't see an example on gov.uk though.

Do you think it's work checking this with Joanna?  I'd be interested to know what problems the prompt is causing, apart from it having different wording.

davidrapson:
This makes sense conceptually, but possibly isn't the _cause_ of that message showing. As we provide our own error handling we should be adding the `novalidate` attribute to forms. We should do both though.

> Do you think it's work checking this with Joanna? I'd be interested to know what problems the prompt is causing, apart from it having different wording.

We follow a similar rationale to https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation, plus the fact that we can translate our error messages too. Would be good to check the pattern though.

davidsauntson:
> This makes sense conceptually, but possibly isn't the _cause_ of that message showing. As we provide our own error handling we should be adding the `novalidate` attribute to forms. We should do both though.
> 
> > Do you think it's work checking this with Joanna? I'd be interested to know what problems the prompt is causing, apart from it having different wording.
> 
> We follow a similar rationale to https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation, plus the fact that we can translate our error messages too. Would be good to check the pattern though.

We have `novalidate` and the `required` attribute already - this isn't really to do with validating a user's input tho, as the prompt shows on hover over a textarea with the `required` attr - before a user has even submitted the form. Looks like a thing Chrome has just decided to implement.

For me whether or not this is worth doing hinges on two things:
- how big of a problem is it for content designers? It's not really a validation issue - our validation messages will still appear in the summary and by the question.
- whether replacing `required` with `aria-required` has any other consequences we don't know about - this is what I think we should check with Joanna at AbilityNet.

davidrapson:
> We have novalidate and the required attribute already

Ooops! You are totally right I misread! Thought we only had them on the react forms in public website. Carry on 👍

marianayovcheva:
Update: Asked about replacing the `required` attribute with aria-required in today's AbilityNet session. As we are not using the default required validations (i.e preventing the form from submitting and displaying a tooltip), there aren't further accessibility implications and `aria-required` will be fine.

davidrapson (approved):

PR #2449, opened 2022-10-24
Icon font to SVG (navigation and breadcrumbs)

davidrapson:
Makes some changes to the breadcrumbs and navigation components to remove the dependency on the icon font.

- Allows icon font styles to be disabled using a config flag
- Switch breadcrumb and navigation chevrons to us CSS shapes rather than the icon font.


davidsauntson (commented):
The left point chevron in the long wrapper (small) example looks a bit squiffy to me - the bottom line is shorter than the top line

squiffy

davidrapson:
Going to revisit the approach to this.

PR #2474, opened 2022-11-09
Pin rails-i18n version

davidrapson:
Pin rails-i18n to 7.0.5 as 7.0.6 removes welsh language locales due to lack of support. We only use these locales for basic date formats so we could inline the parts we need and remove this dependency in future.

Fixes the broken build.


davidrapson:
@davidsauntson @marianayovcheva Mind giving this a quick review, fixes the currently broken build.

davidsauntson (approved):
🍍

PR #2476, opened 2022-11-09
Feat/accessibility page

davidsauntson:
Adds the accessibility statement page to the design system docs.

### DIFF
--- a/design-system-docs/src/_pages/accessibility-statement.md
+++ b/design-system-docs/src/_pages/accessibility-statement.md
@@ -1,0 +1,67 @@
+---
+title: Accessibility statement
+layout: basic
+---
+
+This accessibility statement refers only to the Design System elements and components.
+
+## Accessibility and Design System
+
+A core value of Citizens Advice is to serve people whoever they are, regardless of their accessibility needs. An accessible design system is the first step towards creating inclusive customer experiences for all users. For that reason we embarked on a journey to improve our accessibility record across our services and products.
+
+The first step on this journey was to create a design system that will help us offer a consistent and accessible experience to all, including clients, advisers, volunteers and all staff members.
+
+## What standards do we target?
+
+The Citizens Advice Design System has been designed, built and tested to meet the WCAG 2.1 AA standards.
+
+## Where is the design system used?
+
+We have decided to introduce the design system to the Citizens Advice public website and AdviserNet gradually, making sure there are no disruptions to our services. So far the ‘Immigration’ and ‘Family’ sections use the design system. It is also used by our Intranet, smart meter checking tool, and referrals products.
+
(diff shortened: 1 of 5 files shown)


design-system-docs/src/_pages/accessibility-statement.md
```diff
@@ -0,0 +1,67 @@
...
+
+The first step on this journey was to create a design system that will help us offer a consistent and accessible experience to all, including clients, advisers, volunteers and all staff members.
+
+## What standards do we target?
+
+The Citizens Advice Design System has been designed, built and tested to meet the WCAG 2.1 AA standards.
+
+## Where is the design system used?
+
+We have decided to introduce the design system to the Citizens Advice public website and AdviserNet gradually, making sure there are no disruptions to our services. So far the ‘Immigration’ and ‘Family’ sections use the design system. It is also used by our Intranet, smart meter checking tool, and referrals products.
```
davidsauntson, line 20: We might want to change this so we don't have to keep coming back and updating the parts of the public advice site that are migrated to content-platform?
```suggestion
We have decided to introduce the design system to the Citizens Advice public website and AdviserNet gradually, making sure there are no disruptions to our services. So far, some sections use the design system. It is also used by our Intranet, smart meter checking tool, and referrals products.
```

design-system-docs/src/_pages/accessibility-statement.md
```diff
@@ -0,0 +1,67 @@
...
+- The Referrals product was tested on 15th March 2022
+- The Online Scams Tool was tested on 30th June 2022
+
+All tests were carried out by Digital Accessibility Centre (DAC). Tests were carried out in the following conditions:
+
+- Screen sizes: Desktop, Tablet and Mobile
+- Operating systems: Windows, Android and iOS.
+- Browsers: IE11, Firefox, Chrome, Safari, Android browser,
+- Assistive technologies:
+  - JAWS 18 and 2019
```
marianayovcheva, line 52: Do we have nested list styles? Not sure if the nested list needs a bit of top spacing but will let Kostas decide

marianayovcheva (approved):
Nice! ✨

design-system-docs/src/_pages/accessibility-statement.md
davidrapson: I'd be inclined to write these as a single-line list. We don't typically encourage nesting lists.

davidrapson (approved):
👍 This looks great

One suggestion, but non-blocking: I'd be tempted to keep the navigation as just "Accessibility" and title the page "Accessibility and Design System", removing the top-part. Thinking about it a bit more I'm not sure I'd consider this an accessibility _statement_ per-se. It's more a primer on accessibility within the design system.

We should probably reserve the statement for something about the website itself. Maybe I'm over thinking this?

PR #2479, opened 2022-11-10
Bundle Welsh translations for date formats

### LINKED ISSUE #2478: Find a new approach for Welsh language translations for dates
The rails-i18n gem we were using to provide Welsh language translations for core Rails behaviour like localised dates recently pulled support for Welsh language in a patch version.

Whilst this is unfortunate we only used a small portion of the translation file to provide date format translations. We could consider inlining the translations for the functions we use as we already bundle our own locale file. I think this is one instance of `I18n.l` for a date and `number_to_human_size`.

davidrapson:
The rails-i18n gem historically provided these but the cy locale was removed in 7.0.6 due to lack of support. See https://github.com/svenfuchs/rails-i18n/pull/1007

As we only need a small set of translations for date formatted provided by design system components we bundle the necessary translations ourselves.

Fixes #2478

### DIFF
--- a/engine/config/locales/cy.yml
+++ b/engine/config/locales/cy.yml
@@ -44,3 +44,64 @@ cy:
       close_label: Cau
       descriptive_label_hide: cau’r adran hon
       descriptive_label_show: dangos yr adran hon
+
+  # The rails-i18n gem historically provided these but
+  # the cy locale was removed in 7.0.6 due to lack of support.
+  # See https://github.com/svenfuchs/rails-i18n/pull/1007
+  #
+  # As we only need a small set of translations for date
+  # formatted provided by design system components we bundle
+  # the necessary translations ourselves.
+  date:
+    abbr_day_names:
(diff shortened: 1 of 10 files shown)


engine/config/locales/cy.yml
```diff
@@ -44,3 +44,64 @@ cy:
...
+      - Dydd Mercher
+      - Dydd Iau
+      - Dydd Gwener
+      - Dydd Sadwrn
+    formats:
+      default: '%d-%m-%Y'
+      long: '%d %B %Y'
+      short: '%d %b'
+    month_names:
+      -
```
davidsauntson, line 91: ?

davidsauntson (approved):

engine/config/locales/cy.yml
davidrapson: If I understand the original code correctly it's because of non-zero-indexing (i.e. January is month 1)

PR #2507, opened 2022-11-25
Patch ViewComponent error in docs

davidrapson:
Includes a patch for ViewComponent `TypeError: no implicit conversion of nil into String` errors when used in the docs site (outside of Rails).

Avoids us needing to pin the version. Although not sure if it's better to just keep the version pinned for a while vs. patching the core design-system package 🤔

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+**Bugfixes**
+
+- Includes a fix for ViewComponent `TypeError: no implicit conversion of nil into String` errors when used outside of Rails
+
 **New**
 
 - Input: Replace the `required` attribute on inputs with `aria-required`
--- a/design-system-docs/Gemfile
+++ b/design-system-docs/Gemfile
@@ -25,11 +25,6 @@ gem "puma", "~> 5.6"
 
 gem "citizens_advice_components", path: "../engine"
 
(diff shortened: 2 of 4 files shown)


marianayovcheva:
I don't have answer to the question:
> Although not sure if it's better to just keep the version pinned for a while vs. patching the core design-system package

So leaving to @davidsauntson to have a look/approve

davidsauntson:
> Includes a patch for ViewComponent `TypeError: no implicit conversion of nil into String` errors when used in the docs site (outside of Rails).
> 
> Avoids us needing to pin the version. Although not sure if it's better to just keep the version pinned for a while vs. patching the core design-system package 🤔

Looks like there is a fix on the way, so might just be worth pinning the version for a while longer?

davidrapson:
> Looks like there is a fix on the way, so might just be worth pinning the version for a while longer?

Yeah, looks like it doesn't it. Will move this to draft and just sit on it for a bit. It's only really affecting the docs site so pinning doesn't do much harm.

davidrapson:
Going to close this as the issue is being tracked here https://github.com/ViewComponent/view_component/pull/1571 and on reflection pinning in the docs site is better than a workaround for the time being.

PR #2508, opened 2022-11-25
Upgrade to node v18

davidrapson:
The latest LTS version of Node is now v18. Update the nvmrc file and upgrade the package-lock version (format changed in recent npm versions).

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/fermium
+18
--- a/demo/backstop-test-url.js
+++ b/demo/backstop-test-url.js
@@ -5,8 +5,7 @@
  * in the backstop:ci npm script.
  * See https://github.com/citizensadvice/design-system/pull/2210
  */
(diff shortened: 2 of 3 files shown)


davidrapson:
image

😢  It's a good thing I'm currently at 61k negative lines.

.nvmrc
```diff
@@ -1 +1 @@
-lts/fermium
+18
```
marianayovcheva, line 1: Curious, can we use `lts/hydrogen`?

marianayovcheva (approved):

.nvmrc
davidrapson: We can but I always forget what those codenames mean and realised we can just put the number in here. Do you have a preference the other way?

marianayovcheva: Not really, flagged mostly because the old one and the public website use the codenames. I assume they won't differ in the required version?

PR #2515, opened 2022-11-28
Upgrade dev dependencies

davidrapson:
Upgrade prettier to 2.8.0 and remove the duplicate version from the demo app. We run prettier as a top-level command so we don't need both versions.

Also upgrade eslint, stylelint, size-limit, and testing-library dependencies.


davidrapson:
Merging this as I'm just batching up some dependabot PRs.

PR #2528, opened 2022-12-05
Support paragraph content in tables

davidrapson:
We occasionally have need to support paragraph content within table cells. This is particularly common in content that comes from our CMS and most useful where a table cell may have a couple of lines of text which is common in some of our more complex content tables.

Updates the table styles to apply the common typographic scale to the default text sizing (previously this was fixed to the default size rather than respecting the scale) and apply paragraph-specific styles to paragraphs themselves. This ensures that text sizes are balanced whether the content is in a paragraph or not.

Includes a visual regression test for different paragraph content.


scss/4-elements/_tables.scss
```diff
@@ -44,6 +49,15 @@
     display: none;
   }
 
+  p {
+    font-size: $cads-font-size;
+    margin: 0;
+  }
+
+  p + p {
+    margin-top: 0.75em;
```
davidsauntson, line 58: Should this be a standard spacing? Normal `p` spacing is 1.5rem (`$cads-blah-5`)

davidsauntson: probs should look at adding the standard line height too
maybe just mixin `%cads-paragraph` ?

davidsauntson (approved):
📎

PR #2589, opened 2023-01-03
On this page: Fix css columns overflow bug

davidrapson:
## Background

The on this page component has three main variants:

- A basic layout (fewer than four items) which is a simple list of links
- A multi-column layout (four or more items) which uses css columns
- An expandable layout when `show_nested_links` is set

## Issue

In newer versions of Chrome when viewing the expandable version of the component at a non-default zoom level it is possible for the links to overflow when expanding a list of child links.

Screenshot 2023-01-03 at 12 03 36

https://user-images.githubusercontent.com/123386/210355634-3cedef85-5516-4693-8497-c39cd6ef91d0.mov

After some investigation, the cause of the issue was down to the way we were using css column styles for this component variant. Thanks to @suzannehamilton for the initial reproduction and screen recording.

The way we had the CSS set up for the expandable variant to unset the column layout by using `columns: 1`. This along with having `break-inside: avoid` set to avoid columns being split when the wrapped caused the underlying issue. CSS column calculation is notoriously buggy; thankfully in this case we can avoid it entirely.

## Changes

Because the multi-column layout is really an exception, rather than the rule, we can adjust our styles to scope and multi-column styles to only the variant that needs them i.e. rather than setting `columns: 1` on our expandable variant simply use default list styles.

This avoids any layout calculation bugs with css columns, especially as we don't need it.


scss/6-components/_on-this-page.scss
```diff
@@ -8,29 +8,15 @@
...
     color: $cads-language__secondary-text-colour;
   }
 
-  &__list {
+  .cads-on-this-page__list {
     @extend %cads-list-no-bullet;
-
-    padding: 0;
-    margin: 0;
-    list-style-type: none;
```
davidrapson, line 23: Removing this as `cads-list-no-bullet` already contains these styles.

scss/6-components/_on-this-page.scss
```diff
@@ -8,29 +8,15 @@
...
-  }
-
-  &__list-item {
-    position: relative;
-
-    @include cads-media-breakpoint-up(md) {
-      -webkit-column-break-inside: avoid;
-      page-break-inside: avoid;
-      break-inside: avoid;
-    }
```
davidrapson, line 33: This was the main issue, combined with `columns: 1` was causing a miscalculation depending on the zoom level. Ironically these styles exist to workaround a _different_ css columns bug, making sure items don't wrap across two columns. Moves these styles to be scoped to the `--columns` variant.

scss/6-components/_on-this-page.scss
```diff
@@ -63,14 +55,17 @@
...
     padding-left: cads-rem(18px);
 
     @include cads-media-breakpoint-up(md) {
       padding-left: $cads-spacing-5;
     }
   }
 
+  .cads-on-this-page__list-item {
+    position: relative;
+  }
```
davidrapson, line 67: We only need the relative positioning for the `--expandable` variant.

scss/6-components/_on-this-page.scss
```diff
@@ -63,14 +55,17 @@
 ///
 .cads-on-this-page--expandable {
   .cads-on-this-page__list {
-    columns: 1;
```
davidrapson, line 66: Removing this avoids any column bugs for this component variant. We don't need to use column styles at all.

marianayovcheva (approved):
Nice! 👏 
A side question - should we include an example of the component with no columns to the docs? Or mention that if there are less than four links, they will appear in one column?

davidrapson:
> A side question - should we include an example of the component with no columns to the docs? Or mention that if there are less than four links, they will appear in one column?

Good shout, have added that in to this PR.

PR #2611, opened 2023-01-17
feat: logo slot and hide_logo attribute added to footer

marianayovcheva:
While working on recent footer updates, I needed to change the logo in the footer and later remove the logo completely. I thought that it would be good to add more flexibility to the design system regarding the footer logo. This PR:
- adds the ability to remove the logo using the `hide_logo` attribute
- adds `logo` slot to allow customising the logo where needed. If there is no logo slot added, it defaults to the current logo link (otherwise this would be a breaking change)

The way the logo slot works is inspired by the `HeaderComponent` logo slot


davidsauntson (commented):
This is great!

I have a couple of thoughts that it would be good to get your's and @davidrapson's views on.

- we could rip the band aid off and have a breaking change where the presence of the logo slot or not determines if it is rendered, rather than a dedicated boolean param
- we could infer the presence of the logo from the `homepage_url` and not provide a fallback for it here https://github.com/citizensadvice/design-system/blob/afe2cd994cf00f897b7bc9a0853184adc7fa4857/engine/app/components/citizens_advice_components/footer.rb#L14 - this wouldn't be a breaking change and make a little bit of sense to me, since the url is only ever rendered in the logo anyway

davidrapson:
I definitely prefer this version:

> we could rip the band aid off and have a breaking change where the presence of the logo slot or not determines if it is rendered, rather than a dedicated boolean param

Especially as it's a closer match for the header component then. I think we can be relatively bullish on breaking changes to component interfaces given if we need to be. Supporting `hide_logo` feels most awkward to me, would be happy with the `homepage_url` option too.

General approach looks ace though, always happy to avoid needing custom css background images for this use case.

marianayovcheva:
Thanks both! I've now added the footer logo slot as a breaking change. I've also ended up removing the `homepage_url` attribute because it was only used by the previous logo. I've mentioned this feature in: https://github.com/citizensadvice/design-system/issues/2626

demo/app/components/app_footer_component.rb
```diff
@@ -3,6 +3,7 @@
 class AppFooterComponent < ViewComponent::Base
   def call
     render CitizensAdviceComponents::Footer.new do |c|
+      c.logo(url: "/")
```
marianayovcheva, line 6: I think we've discussed it before but it feels like the logo in the design system docs should lead to the main CA site (or be a design system specific logo)

davidrapson (approved):
👍 Realised this was still waiting on review but it looks fab to me barring the older conflicts from the viewcomponent 3 upgrade.

cnorthwood:
this is now 2 years old... is this work still of value to be dusted off and merged, or is it not worth it now?

marianayovcheva:
I think it will be a painful merge and remembering what it does, so probably better to close the PR and start fresh (I'll add an issue as a reminder). There is still value in it, but it's more of a nice to have.

PR #2655, opened 2023-02-13
Feat/site indexing

marianayovcheva:
We've added the design system site to the Google Search Console. This is a follow up PR that:
- removes the file that we needed to initially verify the site
- removes the `noindex, nofollow` meta tag, flagged by the Search Console. Removed it from the design system default layout pages but intentionally left it on the component_example pages.

### DIFF
--- a/design-system-docs/src/_layouts/default.erb
+++ b/design-system-docs/src/_layouts/default.erb
@@ -4,7 +4,6 @@
     <meta charset="utf-8" />
     <%= seo %>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <meta name="robots" content="noindex, nofollow">
     <meta name="format-detection" content="telephone=no">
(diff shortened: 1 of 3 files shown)


design-system-docs/src/_layouts/default.erb
```diff
@@ -4,7 +4,6 @@
     <meta charset="utf-8" />
     <%= seo %>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <meta name="robots" content="noindex, nofollow">
```
davidsauntson, line 7: Ah ok - is this why google was flagging the site as suspicious? 🤔 

Taking these out means Google will index the documentation site.

marianayovcheva: I thought so initially (along with the fact that we have password fields examples). But as the issue hasn't happened for a while I think we can leave those as they are. That's also because we've been discussing having a different domain so probably worth allowing indexing once that's in place. I'll close the PR for now, but may re-open in the future

PR #2703, opened 2023-03-24
Upgrade citizens advice style

davidrapson:
Batched up a couple of dependabot updates to upgrade to the latest version of https://github.com/citizensadvice/citizens-advice-style-ruby

### DIFF
--- a/demo/Gemfile
+++ b/demo/Gemfile
@@ -33,9 +33,7 @@ gem "haml-rails"
 group :development, :test do
   gem "citizens-advice-style",
       github: "citizensadvice/citizens-advice-style-ruby",
-      tag: "v9.0.0"
-
-  gem "rubocop-rails", require: false
+      tag: "v10.0.1"
 
   gem "cypress-rails"
 end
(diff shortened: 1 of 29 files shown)


demo/Gemfile
```diff
@@ -33,9 +33,7 @@ gem "haml-rails"
 group :development, :test do
   gem "citizens-advice-style",
       github: "citizensadvice/citizens-advice-style-ruby",
-      tag: "v9.0.0"
-
-  gem "rubocop-rails", require: false
```
marianayovcheva, line 38: Is this because rubocop rails is now required?

marianayovcheva (approved):
LGTM! Only things I'm a bit unsure about is that ruboco-rails and rubocop-rspec are now required but I guess this is because of what the changelog says about required rubocop, rspec etc. 
Should we also update the gem in the `design-system-docs` repo as well?

demo/Gemfile
davidrapson: Yeah citizens-advice-style now bundles rubocop-rails and rubocop-rspec

PR #2743, opened 2023-04-20
Feat/custom legal summary

marianayovcheva:
Added the ability to customise the legal summary - for example if the product using the footer has a different address or requires different legal information. If no custom legal summary was provided, defaults to the standard legal summary.

There is another open PR about footer changes (https://github.com/citizensadvice/design-system/pull/2611), but this one takes priority as the changes to the legal summary are non breaking and cannot currently be achieved unless the html of the component is copied over.

### DIFF
--- a/engine/app/components/citizens_advice_components/footer.rb
+++ b/engine/app/components/citizens_advice_components/footer.rb
@@ -2,17 +2,18 @@
 
 module CitizensAdviceComponents
   class Footer < Base
-    attr_reader :homepage_url, :feedback_url, :legal_summary
+    attr_reader :homepage_url, :feedback_url
 
     renders_one :feedback_link, "FooterFeedbackLink"
 
+    renders_one :legal_summary_slot, "FooterLegalSummarySlot"
+
     renders_many :columns, "FooterColumn"
 
-    def initialize(homepage_url: nil, feedback_url: nil, legal_summary: nil)
+    def initialize(homepage_url: nil, feedback_url: nil)
       super
       @homepage_url = homepage_url || "/"
       @feedback_url = feedback_url.to_s
-      @legal_summary = legal_summary || t("citizens_advice_components.footer.legal_summary")
 
       feedback_url_deprecation
     end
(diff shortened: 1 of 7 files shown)


davidrapson:
Just a quick conceptual question before approving:

Technically because this text is in the locale file you can technically provide your own copy of that text using Rails translations in your app:

```yml
---
en:
  citizens_advice_components:
    footer:
      legal_summary: My custom text
```

Is it worth seeing if that feels good enough on it's own? Is that too confusing a way of customising the built in text?

If that's too weird, how do you feel about a slot for this vs an argument?

engine/app/components/citizens_advice_components/footer.rb
```diff
@@ -50,6 +52,16 @@ def feedback_url_deprecation
       )
     end
 
+    def legal_summary
+      legal_summary_slot.presence || legal_summary_fallback.presence
+    end
+
+    def legal_summary_fallback
```
marianayovcheva, line 59: I've added this to avoid making a breaking change by making the legal summary slot required. I've not added a depreciation warning because in most of the cases, we want to use the default legal summary and only add a slot when a custom text is needed

marianayovcheva:
Update: based on @davidrapson's suggestion I'm now using a slot (with a fallback) instead of a argument for the legal summary

engine/app/components/citizens_advice_components/footer.rb
```diff
@@ -6,6 +6,8 @@ class Footer < Base
 
     renders_one :feedback_link, "FooterFeedbackLink"
 
+    renders_one :legal_summary_slot, "FooterLegalSummarySlot"
```
davidrapson, line 9: As we're working with a string here we probably don't need the inner class here we can probably do this with a single inline method:

```rb
renders_one : legal_summary, lambda do |text|
  text.presence || t("citizens_advice_components.footer.legal_summary")
end
```



Also mind removing the `_slot` suffix? I'm not sure we use that convention anywhere else.

marianayovcheva: Good shout, thanks! I've now updated

PR #2758, opened 2023-05-02
Add ViewComponent 3 support

davidrapson:
ViewComponent 3 has been released. Raising this as a draft to extract out what needs to be updated. We should be able to make all required changes ahead of time as we want this to be backwards compatible.


davidsauntson:
Looking `with_good`

Are you going to separate this out into other PRs?

davidrapson:
> Are you going to separate this out into other PRs?

Yeah that was https://github.com/citizensadvice/design-system/pull/2759 and the remainder I think is mostly in https://github.com/citizensadvice/design-system/pull/2767 now so this can probably be closed. Will double check if there's anything missing.

PR #2767, opened 2023-05-12
Support and test against ViewComponent 3

davidrapson:
This pull request widens the gemspec version range and makes it possible to use view component 3 with no deprecation errors. Includes a single actions job to test against the new version so we can make sure changes are compatible with both versions for a while.


demo/Gemfile
```diff
@@ -25,6 +25,9 @@ gem "high_voltage", "~> 3.1"
 # Bundle the design system engine
 gem "citizens_advice_components", path: "../engine"
 
+## Explicitly pin to ViewComponent 2 for the demo app
+gem "view_component", "~> 2.82.0"
```
davidrapson, line 29: This is required as whilst the engine now supports view component 3 not all of the internal usages are updated. You'd probably need to do this for existing apps that have their own components.

engine/app/components/citizens_advice_components/footer.html.erb
```diff
@@ -5,7 +5,7 @@
         <div class="cads-grid-col">
           <p class="cads-footer__feedback">
             <%= link_to(
-              feedback.title,
+              feedback.title || t("citizens_advice_components.footer.website_feedback"),
```
davidrapson, line 8: This fixes a deprecation warning with `ViewComponent::Translate` in view component 3

.github/workflows/ci-workflow.yml
```diff
@@ -26,13 +26,19 @@ jobs:
...
+        view_component_version: ['2.82.0']
         exclude:
           - rails_version: '6.0.6'
             ruby_version: '3.0'
           - rails_version: '6.0.6'
             ruby_version: '3.1'
+        include:
+          - ruby_version: '3.1'
+            rails_version: '7.0.4'
+            view_component_version: '3.0'
```
davidrapson, line 38: Adds a single build for view compnent 3

marianayovcheva (approved):
I've checked the footer change and it looks good to me! 

>we probably want to update the docs to use the with_ syntax before we merge this to avoid any confusion if you were starting a fresh project with view component 3.

agree because I was a bit confused about it looking between https://github.com/citizensadvice/design-system/pull/2758 and https://github.com/citizensadvice/design-system/pull/2759

demo/Gemfile
davidrapson: Ah I just meant the code in https://github.com/citizensadvice/design-system/tree/main/demo

davidrapson:
@marianayovcheva I've updated this so that everything including the demo app and design system docs are updated to support the latest view component.

marianayovcheva (approved):
⭐ ⭐ ⭐

PR #2816, opened 2023-06-16
chore: updating cypress dependencies

marianayovcheva:
Bundling cypress-related upgrades

Ruby:
- cypress-rails: 0.5.5 -> 0.6.0

JS:
- @testing-library/cypress: 8.0.3 -> 9.0.0
- cypress: 10.11.0 -> 12.14.0 - this is a major upgrade, however we are not using any of the features that cause breaking changes: https://docs.cypress.io/guides/references/migration-guide


davidrapson (approved):
👍

PR #2874, opened 2023-07-21
Remove misleading arguments in header examples

davidrapson:
The examples and demo code for using a custom `account_link` were misleading as they continued to pass arguments to the method as well as providing a block.

When providing a block this arguments are ignored. Updated the example and demo code to reflect this.

Spotted whilst pairing with @pksterling

### DIFF
--- a/design-system-docs/src/_component_docs/header.md
+++ b/design-system-docs/src/_component_docs/header.md
@@ -118,10 +118,8 @@ Or by passing a custom block to render your own HTML:
 ```erb
 <%%= render CitizensAdviceComponents::Header.new do |c| %>
   <%% c.account_link do %>
-    <%% c.account_link(title: "Sign in", url: "/sign-in") do %>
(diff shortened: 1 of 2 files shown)


design-system-docs/src/_component_docs/header.md
```diff
@@ -118,10 +118,8 @@ Or by passing a custom block to render your own HTML:
 ```erb
 <%%= render CitizensAdviceComponents::Header.new do |c| %>
   <%% c.account_link do %>
-    <%% c.account_link(title: "Sign in", url: "/sign-in") do %>
```
davidrapson, line 121: `c.account_link` was also duplicated here.

davidrapson:
@pksterling If you are interested in reviewing this next week you can see the change in this preview app https://64bab3b3562722058a14cc6c--citizens-advice-design-system.netlify.app/components/header/#account-link-slot and/or I can show you how to run the design system source locally if you are interested at some point.

PR #2920, opened 2023-08-15
Feat/tables no responsive headers

marianayovcheva:
On mobile our current table set up adds the table headers to each table cell. This is useful for long tables as the user would scroll past the table headers. However in some instances (especially for short tables) we may want to switch off this behaviour. I've added a `responsive_headers` attribute which by default is set to true and keeps the existing table behaviour (making this a non-breaking change).

Initially I had some concerns around the accessibility of the tables if the stacking behaviour is switched off. However, I have checked examples in different design systems ([gov.uk](https://design-system.service.gov.uk/components/table/), [NHS](https://service-manual.nhs.uk/design-system/components/table), [Co-op](https://www.coop.co.uk/experience-library/elements/tables.html)) and their tables keep the desktop layout. I've also had a look at some accessibility documentations (https://webaim.org/techniques/tables/data and https://www.w3.org/WAI/tutorials/tables/one-header/) around table and there was no mention about the table stacking, they mentioned other practices (such as adding scope) which we're doing anyways.


scss/4-elements/_tables.scss
```diff
@@ -50,58 +46,63 @@
...
-    tr td:last-child {
-      border-bottom: solid $cads-border-width-small
-        $cads-language__input-border-colour;
-    }
-
-    td::before {
-      display: inline;
-      font-weight: bold;
-      text-align: left;
+  &--responsive-headers {
```
davidrapson, line 49: 💅 Style: I prefer using the flatter full selector names i.e. `.cads--responsive-headers` vs these shorthands, but it get's compiled the same so it's more for search-ability than anything else.

davidrapson (changes requested):
👍 This is all really clear to me and makes sense to have this as an option. Could we add a couple test scenarios for this, so a view component spec and am extra visual regression test scenario? Not sure it's worth an extra cypress test as there's not any _behaviour_ here really.

davidrapson:
> I am about 90% certain that we used to have a table variant that didn't shrink and overflowed on the x axis. I might be wrong tho 👴🏻 Either way it doesn't live here any more - just thought I'd mention it tho. I think maybe it wasn't very accessible. @davidrapson can you remember?

All I can see is that the `cads-table-container` has an overflow scroll on it but not 100% sure when that triggers. Overflowing on two axis will normally get flagged up in an accessibility audit so we might not actually want that.

PR #2924, opened 2023-08-16
Upgrade demo app dependencies

davidrapson:


demo/spec/components/previews/section_links_preview/with_additional_content.html.haml
```diff
@@ -1,3 +1,4 @@
 = render CitizensAdviceComponents::SectionLinks.new(title: "Related Content", section_title: "Applying to the EU settlement scheme", section_title_url: "/immigration#h-applying-to-the-eu-settlement-scheme") do |c|
   - c.with_section_links(section_links)
-  .example-links-block= link_to "Adviser resources - EU Settlement Scheme", "/advisernet/immigration/staying-in-the-uk/settled-status/adviser-resources-eu-settlement-scheme"
+  - c.with_custom_content do
```
davidrapson, line 3: This was causing a deprecation warning as it was still using the old format for custom content.

.github/workflows/ci-workflow.yml
```diff
@@ -106,6 +106,10 @@ jobs:
         with:
           working-directory: demo
 
+      # Run rubocop against demo app code
+      - run: bundle exec rubocop --format progress --format github
+        working-directory: demo
```
davidrapson, line 111: We weren't running this against the demo app code so there were a few lint warnings that were missed. Adds this and then autofixes the warnings.

davidsauntson (approved):

PR #2943, opened 2023-08-30
notice banner bugfix: keep label vertically centred on banner

cnorthwood:
Currently, when the content of a banner spreads over multiple lines on desktop, then the label stays aligned at the top which makes it look unbalanced. This keeps it vertically centered.

Before:

Screenshot 2023-08-30 at 16 16 29

After:

Screenshot 2023-08-30 at 16 16 43


davidrapson (approved):
👍 Looks great!

PR #2944, opened 2023-08-30
chore: bump backstop dependency

cnorthwood:
BackstopJS 6.2 adds native support for ARM-based Macs, so there's no need to have a separate config for M1 Macs any more, and it also resolves some issues I had running it locally on my Mac


davidrapson (approved):
Amazing! Last time I tried this there were a few sticky visual regression failures but this looks like it passed first time.

Just ran this locally and can confirm it still works on my Intel Mac too so given that it's also passing in CI and assuming you are happy it runs on ARM-based Macs too we can just get this in 🙌 . Merge away!

PR #2945, opened 2023-08-31
Adjust print visual regression test references

davidrapson:
After upgrading to Backstop 6.2 the print page reference samples have become flaky, sometimes recording extra whitespace at the bottom of the page.

Try re-recording them but using the `#content` container as the selector rather than the document as the narrowest page selector we can use whilst still capturing the full area.

Also removes a redundant `excludeFromA11yTest` flag from when we used pa11y rather than cypress-axe.


cnorthwood:
ah I think I've cracked this one seperately in https://github.com/citizensadvice/design-system/pull/2943/files#diff-26d0eb982d5e35b1411954eac7f573fbb03da38e28a7edb90c9e7a9af000093fL48, I found a comment on the backstopjs issue tracker (which I foolishly closed) that recommended using `html` rather than `document` as a selector and that feels more stable now

davidrapson:
Amazing, feel free to close this if the fix in your PR works fine 👍

cnorthwood:
I think they achieve the same thing and fundamentally take the same approach, so I guess it depends if we want this as a nice separate PR or mine which now muddles 2 different things, I think this one is probably clearer!

davidrapson:
Yeah let's maybe keep this separate then, let me re-run this a couple of times first.

cnorthwood (approved):

PR #2948, opened 2023-08-31
Update project defaults and CI ranges

davidrapson:
- Use Ruby 3.2 as the default project version
- Expand the set up guide for prerequisites
- Loosen dependency version in CI to ensure we're always testing against the latest patch versions of Rails and ViewComponent
- Consistently use ubuntu-latest as the CI runner


.github/workflows/ci-workflow.yml
```diff
@@ -25,17 +25,17 @@ jobs:
       fail-fast: false
       matrix:
         ruby_version: ['2.7', '3.0', '3.1', '3.2']
-        rails_version: ['6.0.6', '6.1.7', '7.0.6']
-        view_component_version: ['2.82.0']
+        rails_version: ['~> 6.0', '~> 6.1', '~> 7.0']
+        view_component_version: ['~> 2.82']
```
davidrapson, line 29: This is the notable change, before this was testing against specific patch versions which change regularly. Softening the ranges here ensures we're testing against the latest version that matches the constraint, e.g. the latest `6.1.x.x` patch of rails or the latest `3.x` version of viewcomponent.

cnorthwood (approved):

.github/workflows/ci-workflow.yml
cnorthwood: I think this is good. The tradeoff is that we might start to get build failures unrelated to a code change if an incompatibility is introduced upstream which is frustrating - but we do want to know about it.

davidrapson: Yeah that could definitely happen. We used to run against Rails `main` for a while but that was too variable. Hopefully minor versions should be safe and as you say we want to know if it breaks.

PR #2949, opened 2023-08-31
Remove generated example files

davidrapson:
These previously existed as a bridge between the view component previews and storybook. They are no longer connected to anything but we kept them around as we anticipated that they might act as a useful reference.

In practice these have not been updated since we removed storybook and the documentation website and demo app provide a better implementation reference than relying on these static example files, which have begun to drift having not been updated in several months.


cnorthwood (approved):
This looks good to me - I don't have the full context if you want to wait for a second review but reducing the size of the codebase to be maintained sounds very sensible, especially if these have fallen out of use

marianayovcheva (approved):
To be completely honest, I don't remember using these files (or at least not in the past year) so very happy for them to be removed. 🧹

PR #2951, opened 2023-08-31
Add dependabot groups for common dev-dependencies

davidrapson:
Now that dependabot groups are in general availability we can try configuring them for a few things.

https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/

Start with rules for the testing tools that have a few related packages (eslint, stylelint, jest)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,12 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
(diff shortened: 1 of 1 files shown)


marianayovcheva (approved):
LGTM! 
(also selfishly want to see what these types of updates will look like as I'd want to use a similar approach for the public website)

PR #2963, opened 2023-09-05
Release/v5.5.0

davidrapson:
Prepares a v5.5.0 release. Rolls up the release notes from all previous alpha releases. In answer to the question "Do we need pre-releases like we used to?" is definitely no as the first 5.5.0 alpha was made nearly a year ago 🙈


davidsauntson (approved):
wow - a year ⚰️

PR #2985, opened 2023-09-15
New guides: Using with Rails

davidrapson:
## Background

In the Design System working group one of the activities we've been looking at is clearer technical documentation for using our components with Rails.

We identified a good place to start as being a guide which stepped through installing and configuring the design system with Rails. Some of this information was in the existing overview, but only up to a point.

## Changes

This pull request does the following:

- Adds a new guides collection
- Adds a draft Using with Rails guide, this is the main substance of this PR
- Moves the existing getting started guide into this new section, renaming it overview and removing any detail that is better served by the new Rails guide
- Restructures some of the docs layouts now that we have a larger number of collections to remove duplication
- Scopes the code copy button to a dedicated class as it is overwhelming on a tutorial guide with lots of smaller code snippets, better to limit it to the component code samples (or at least opt-in)

image


davidrapson:
Marking this as ready to review as it can be code and copy reviewed now but/and won't merge this until after the next working group at the earliest.

marianayovcheva:
I've just had a read and the step by step guidance looks great! I admit I've not tried it out locally but overall makes sense. A couple of comments/suggestions:
- Is it worth changing the "Using with Rails" title to something that reflects more that the page will go into how to set things up from scratch?
- A suggestion for the "Overview" section would be to keep it similar to the existing "Getting started" page - with code snippets, etc. Trying to think about cases where people would have an existing app running or for example use React, etc and only want information on how to install the npm library.
- I feel like there is value in adding the "On this page component" so that developers can skim through what content is available and quickly navigate to the section they're interested in - especially for the "Using with Rails" page.
- Would this be a good place to add information that the gem and the npm library need to be updated together?

marianayovcheva:
@davidrapson Thanks! I had another look and could see the bit about using the same version. Not sure how to make it more prominent though, maybe one for a content designer.
Looks good to me!

design-system-docs/src/_guides/overview.md
```diff
@@ -0,0 +1,88 @@
...
+@import '@citizensadvice/design-system/scss/lib.scss';
+```
+
+The design system requires the use of [Dart Sass](https://sass-lang.com/dart-sass).
+
+The distributed package includes compiled CSS in `lib/lib.css` if you are not using SCSS.
+
+#### Web fonts
+
+The design system uses distributes two sets of fonts. Open Sans, our brand font, and a custom icon font. These are included with the npm package.
```
pksterling, line 57: The design system **_uses distributes_** two sets of fonts.

This doesn't read right to me

design-system-docs/src/_guides/using-with-rails.md
```diff
@@ -0,0 +1,315 @@
...
+  <div class="cads-grid-container">
+    <div class="cads-grid-row">
+      <%%# Using the `md` variant here ensures the colum is responsive %>
+      <main class="cads-grid-col-md-8">
+        <h1 class="cads-page-title">Hello Design System!</h1>
+      </main>
+    </div>
+  </div>
+</div>
+```
```
pksterling, line 151: This snippet is part of the page layout component - could having it here give the impression there isn't an equivalent component? 
On the other hand, this works well being a first addition to the view while showcasing the classes

pksterling (commented):
I haven't yet followed the fresh app guide, however reading through, it seems clear and easy to follow

design-system-docs/src/_guides/using-with-rails.md
davidrapson: The `PageLayout` component doesn't actually exist as part of the design system, it's more a convention we've settled on within applications themselves. Might make a good contribution if you fancy it though!

PR #3009, opened 2023-09-27
chore: upgrade glob to 10.3.10

marianayovcheva:
The dependabot upgrade of glob [failed](https://github.com/citizensadvice/design-system/pull/3001) so I tried to manually upgrade locally and investigate - after manual upgrade there were no more issues


davidrapson (approved):
👍

PR #3010, opened 2023-09-29
Feat/date input

marianayovcheva:
For a while we have had a date input field within the public-website, even though it wasn't used and often people weren't aware it exists. We are moving the date input field into the design system so that it can be reused across products.


design-system-docs/src/_component_docs/date-input.md
```diff
@@ -0,0 +1,45 @@
...
+
+### With validation error
+
+<%= render(Shared::ComponentExample.new(:date_input, :with_validation_error)) %>
+
+### With values
+
+<%= render(Shared::ComponentExample.new(:date_input, :with_values)) %>
+
+## Using with Rails
```
marianayovcheva, line 37: My personal opinion is that the "Using with Rails" section is probably a bit redundant because it often matches information that users can get from the examples "ERB code". This is just an observation and we should have a separate discussion and potentially PR about it.

davidrapson (approved):
👍 Looking great. There's a couple of comments but hitting ✅ approve as they are all small and not blocking.

If you were up for it, it would be amazing to get an example added to the demo app sample form in https://github.com/citizensadvice/design-system/blob/main/demo/app/models/example_form.rb. That might need a little custom code writing to parse the date parts and surface the validation errors but would be good to have a test case of that. Maybe one for a different PR.

marianayovcheva:
Thanks @davidrapson! I've worked though the comments. I've also added a date component in the example form (good shout!). As you suggested we probably need to explore the validation a bit further but I'll leave this for a different PR.

PR #3019, opened 2023-10-05
Feat/select

marianayovcheva:
Added a select component to the design system (thanks David S for the initial implementation)


davidsauntson (approved):
🙌🏻 
Should there be an 'optional' vrt scenario too?

PR #3030, opened 2023-10-09
Fix Rails 7.0.x CI version constraint

davidrapson:
Current CI version constraint has started matching Rails 7.1 which is now out. Make sure the range covers 7.0.x explicitly. We'll need to add Rails 7.1 to the build matrix and fix any issues but this gets the build passing.


davidrapson:
Going to merge this straight in to get the builds ✅ again

PR #3032, opened 2023-10-12
Add Rails 7.1 support and drop Ruby 2.7 support

davidrapson:
## Background

Rails 7.1 has been released https://rubyonrails.org/2023/10/5/Rails-7-1-0-has-been-released

Using this as an opportunity to revise our matrix of supported versions as well as ensuring we support Rails 7.1

## Preparatory changes

The main change is to ensure that our specs pass against Rail 7.1. Updates the default `RAILS_VERSION` to 7.1 along with upgrading the demo application. The only practical change needed was to workaround an issue with `travel_to` in the footer specs, I think due to the fact we are selective in the Rails components we load. Simplest approach is to simply test the current year instead.

## Proposed changes in support matrix

The main functional change I'm proposing here is to streamline the supported version matrix so that we support:

1. Active Ruby versions (https://endoflife.date/ruby)
2. Last 3 major Rails releases
3. Any supported ViewComponent versions

In practice this means:

- Adding Rails 7.1.x to the CI matrix
- Removing Rails 6.0.x from the CI matrix
- **Removing support for Ruby 2.7**. This is now unsupported and this feels like a good enough time to drop support

If folks are broadly on board with this I can also include an ADR for dropping Ruby 2.7 support and/or more explicitly defining our supported version matrix.


pksterling (approved):

akamike (approved):
LGTM!

PR #3071, opened 2023-11-07
Add missing foundation colours and deprecate duplicate colour variable

### LINKED ISSUE #3031: Foundation colours does not contain the red colour used in error messages
## Summary of problem

The red colour used on the error component text is not listed in the foundation colours.

Casebook does not use the design system, but it does try to have a rough alignment in terminology and colours.  I was checking the red you use and noticed it wasn't listed.

## Steps to replicate

Visit [/foundation/colours](https://citizens-advice-design-system.netlify.app/foundations/colours/), and (/components/error-summary/)[https://citizens-advice-design-system.netlify.app/components/error-summary/]

## Expected behaviour

The red colour used in the error summary is listed on the foundations colour page

## Actual behaviour

The red colour is not listed

## Context

- https://citizens-advice-design-system.netlify.app/foundations/colours/
- https://citizens-advice-design-system.netlify.app/components/error-summary/

davidrapson:
Fixes https://github.com/citizensadvice/design-system/issues/3031

Adds missing status colours to the foundations documentation. In the process realised that `$cads-language__warning-colour` is a duplicate of `$cads-language__error-colour`. I've removed internal references to the duplicate variable and document it as deprecated.

We could simply remove the variable and require applications to replace the reference or leave it for a little while and replace it more gradually. There's no way to add a deprecation warning to a plain variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 **New**
 
 - Document error and success colours in foundations
+- Deprecate `$cads-language__warning-colour` as this was a duplicate of `$cads-language__error-colour`
 - Include Rails 7.1 in supported version
 - Remove Rails 6.0 from supported versions
 - Remove support for Ruby 2.7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 **New**
 
+- Document error and success colours in foundations
 - Include Rails 7.1 in supported version
 - Remove Rails 6.0 from supported versions
 - Remove support for Ruby 2.7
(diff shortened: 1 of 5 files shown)


mrdaniellewis:
I've also noticed the callout backgrounds are not specified.  It looks like they are a 10% mix with white, but seeing [this note on Tailwind](https://tailwindcss.com/docs/customizing-colors#generating-colors), it might be better to specify them explicitly.

mrdaniellewis:
Also the success background is implemented differently and is a 10% opacity of the foreground rather than a 10% mix of white?

PR #3072, opened 2023-11-08
Callout: Replace `title` with `attributes`

pksterling:
The `title` attribute in the callout component was used as an `aria-label` for its container.

This is being changed because:
- The callout container is a non-interactive element and `aria-label` is intended for use on interactive elements only (as described in the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label))
- Allowing only one attribute to be passed for a specific option in the container element is quite restrictive

The following changes have been made:
- A deprecation warning has now been added to the use of `title` in the callout component
- `attributes` is now an optional attribute that passes a hash of options to the callout component element

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Include Rails 7.1 in supported version
 - Remove Rails 6.0 from supported versions
 - Remove support for Ruby 2.7
+- Callout: Deprecate `title` attribute; replace with more general  `attributes`
 
 ## v5.6.0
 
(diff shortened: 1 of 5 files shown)


pksterling:
It was also discovered while making these changes, that `ActiveSupport::Deprecation.warn` is being deprecated but is used in a handful of places in the codebase

davidrapson (approved):
🚀 Wonderful pull request description and commit messages, let's merge this!

PR #3115, opened 2023-12-19
Feat: allowing passing an id to the inputs

marianayovcheva:
We are looking into using and customising the Rails form helpers. The by default rails naming convention for `name` of an input field (in the example a text input) would be `name="user[first_name]"` - when we pass the name in this way, this makes it possible for us to then access `params[:user]` and receive `{"first _name"=>"the name that user has filled in"}`. The `id` is then transformed to remove the brackets and replace them with underscores `user_first_name`.  However, in design system we have the id generated based on the name i.e `user[first_name]-input`, which is not a valid syntax as it includes the braces.

The original place we noticed the issue was in https://github.com/citizensadvice/rails-form-builder/pull/2#discussion_r1425502352. Name was used as id, so it required overwriting using additional_attributes: https://github.com/citizensadvice/rails-form-builder/pull/2/files#diff-3a9aa515bf320baab3df81082799974c3443280b9df22de837b4ad45d3231f95R32

This PR adds an optional id attribute which allows us to customise the id of the fields.
This is a non-breaking change, as if an id is not added, it defaults to generating the id based on the name.

I also added a small text change to the checkboxes.
I've not included the date field input and select component for now, but will likely adjust those in the future.


marianayovcheva:
I'm keeping this PR a draft for now as I first want to check a couple of things with @leeky and adjust if needed.

engine/app/components/citizens_advice_components/input.rb
```diff
@@ -70,20 +71,26 @@ def hint?
...
+    end
+
     def label_id
-      "#{name}-label"
+      "#{general_id}-label"
     end
 
     def input_id
-      "#{name}-input"
+      "#{general_id}-input"
```
marianayovcheva, line 85: @leeky wanted to check what would make more sense in the case of form builder. I know that the rails way wouldn't have the `-input` at the end of the id, should we remove it? 🤔

leeky: Rails doesn't really care about the HTML IDs, it's the `name` that it cares about.

demo/spec/components/previews/checkbox_group_preview.rb
```diff
@@ -87,4 +96,11 @@ def additional_attribute_checkboxes
       { label: "Option 2", value: "2", "data-testid": "second-input" }
     ]
   end
+
+  def custom_id_checkboxes
+    [
+      { label: "Option 1", value: "1", id: "test-id-first" },
```
leeky, line 102: Isn't the `id` defined as a parameter on the component itself rather in the options array?

i.e.

```ruby
CitizensAdviceComponents::CheckboxGroup.new(
  name: "example_form[contact_preferences][]",
  id: "example_form_contact_preferences"
)
```

demo/spec/components/previews/checkbox_single_preview.rb
```diff
@@ -12,4 +12,10 @@ def error
       c.with_checkbox(label: "I agree to the great terms and conditions", value: "agreed-to-tcs")
     end
   end
+
+  def custom_id
+    render CitizensAdviceComponents::CheckboxSingle.new(name: "a-single-checkbox[test]") do |c|
+      c.with_checkbox(label: "I agree to the great terms and conditions", value: "agreed-to-tcs", id: "test-id")
```
leeky, line 18: Same here. The `id` is on the component rather than the options list.

engine/spec/components/citizens_advice_components/checkbox_group_spec.rb
```diff
@@ -33,6 +33,25 @@
...
+        c.with_inputs(sample_inputs_with_id)
+      end
+    end
+
+    it "adds the correct name to the input" do
+      expect(page).to have_field "Option 1", name: "checkboxes[test][]"
+    end
+
+    it "adds the correct id to the input" do
+      expect(page).to have_selector "#test-id-first-0"
```
leeky, line 51: I'm not sure if understand why the `-0` is getting added here.

I figure that either:
*  the `id` should be provided at the component level, and the `index` is appended to keep them unique
*  the `id` is provided at the `options` level, so it's up the developer to provide unique ids for each one..?

demo/spec/components/previews/checkbox_group_preview.rb
marianayovcheva: Yeah, good shout, thanks! I'll adjust that

engine/spec/components/citizens_advice_components/checkbox_group_spec.rb
marianayovcheva: Yeah, after I adjust the `id` to be a parameter on the component, it will be the first case: 
> the id should be provided at the component level, and the index is appended to keep them unique

We've had some issues with id generation in the past and this is when the indexed ids were introduced: https://github.com/citizensadvice/design-system/issues/1729

davidsauntson (approved):
This makes sense to me ✨ 

Pls could you add a link to the example PR that you showed me when we chatted?  I know it won't be visible to people outside of Citizens Advice but it will be useful for when we look at why we made this change in the future. 🙏🏻

PR #3166, opened 2024-01-22
Remove need for yarn

davidrapson:
Newer versions of cssbundling-rails and jsbundling-rails remove the assumption that yarn is always present and support a much wider range of package mangers including core npm.

Rails 7+ also has less of a dependency on yarn internally (and technically no longer depends on node.js at all).

Combined with the confusing quality that this project uses plain npm for the package and yarn for the sub-projects this is a good point to see if we can consolidate.

Remove references to yarn and update the docs and demo apps to run using plain npm instead.


davidrapson:
@pksterling Were you able to run the demo app and docs sites locally ok?

pksterling:
@davidrapson Yeah, no problems 👍

PR #3186, opened 2024-02-08
Feat/date fields

marianayovcheva:
Similarly to https://github.com/citizensadvice/design-system/pull/3115 we need a way of changing the names/ids of the date input fields. This will help us when we create more Rails forms where the name has a different format: e.g `date_of_birth[day]`. To make this possible, I've added slots (date_fields) for the dates. I then thought that the behaviour of the fields is close to checkboxes (i.e part of a form group) and have tried to have DateInput inherit from FormGroup and DateField inherit from Input. There were a few problems with this approach (would require too much modification) so I decided to go with the initial date field slots. 

This is a breaking change, but wouldn't necessarily require a major release. The date fields were released fairly recently but to my knowledge, there is not product which currently uses them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+**BREAKING CHANGE**
+- Date input component: add date_field slot instead of auto generating date input fields 
+
 **New**
 
 - Include Rails 7.1 in supported version
--- a/demo/app/views/example_form/new.html.erb
+++ b/demo/app/views/example_form/new.html.erb
@@ -47,7 +47,11 @@
         }
       ) %>
 
-      <%= render CitizensAdviceComponents::DateInput.new(name: :date_of_purchase, label: "When did you purchase the goods or services?") %>
+      <%= render CitizensAdviceComponents::DateInput.new(
+        name: :date_of_purchase,
+        label: "When did you purchase the goods or services?") do |c| %>
+        <% c.with_date_fields([{ name: "purchase[day]", id: "day-id", timespan: :day}, { name: "purchase[month]", id: "month-id", timespan: :month }, { name: "purchase[year]", id: "year-id", timespan: :year, }]) %>
+      <% end %>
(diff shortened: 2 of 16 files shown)


leeky:
Given that this repo is public, there's a (albeit tiny) chance that someone is using this component externally, so I'm wondering if we should avoid introducing breaking changes?

Could `with_date_fields` default to provide behaviour that's backward-compatible with any existing usage..?

Also, I'm wondering if `with_date_fields` can take three parameters, one each for the `day`, `month` and `year` parameters? Presumably all three need to exist once - i.e. if I wanted to select a month and year only, would this work correctly if I omitted the `day` option?

marianayovcheva:
Thanks @leeky!
> Given that this repo is public, there's a (albeit tiny) chance that someone is using this component externally, so I'm wondering if we should avoid introducing breaking changes?
Could it default to providing behaviour that's backward-compatible with any existing usage..?

Yeah, I think I started off with this intention and then things got tangled and I felt like I was adding a lot of complexity for it.  I can give it a go again, or we can release it as 6.0.0 to make it obvious for users that they need to follow the new syntax.

> Also, I'm wondering if with_date_fields could be changed to take three separate parameters, one each for the day, month and year? Presumably all three need to exist once - i.e. if I wanted to select a month and year only, would this work correctly if I omitted the day option?

I originally was "inspired" by gov.uk design system passing `items` to the date input: https://design-system.service.gov.uk/components/date-input/ but there is nothing stopping us from changing the current set up. I have a few ideas (one of them is non-breaking change) and want to understand more about how we'll pass the arguments so maybe best to have a chat to go though those?

marianayovcheva:
Closing for now, so that the PR doesn't go stale. Leeky and I have been discussing an alternative approach - the plan is for him to implement the changes required in the form builder gem first, and then we can apply the learnings to the design system component.

PR #3187, opened 2024-02-08
Remove unused `type` option from select component

leeky:
This was only needed because `CitizensAdviceComponents::Select` inherits from `CitizensAdviceComponents::Input` which does require a `type`.

Instead, we pass `type: nil` by default into the initialiser.


davidrapson (approved):
💯 Thanks for this! Reckon we should release a v5.7 when this is merged?

PR #3195, opened 2024-02-13
Remove external link icon

davidrapson:
We have determined that the external link icon is often confusing and ambiguous for users.

This is for many of the reasons outlined here:
https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/

We will completely remove the external link icon from the Design System which allows us to:

- Avoid confusion and ambiguity where the icon is not widely recognised
- Reduce technical complexity within the codebase

An expanded version of this description has been included as an ADR


davidsauntson (commented):
Oh interesting!

Can I check who has agreed / been consulted on this?  Have product and content design been part of the decision making process?

davidrapson:
> Can I check who has agreed / been consulted on this? Have product and content design been part of the decision making process?

Yep 🙂, this was driven by product design. I'll post a link to the thread over in Slack.

davidsauntson (approved):
_Iconic_ 🥁

davidrapson:
> but I wonder if it should be visual as well since we don't have the icon anymore.

Ah so I guess this is a big part of why we are making the change, because the icon never signified opens in a new tab.

The `external` flag is removed so will need to removed from e.g. the footer link and it's actually the `new_tab` option which controls the new tab behaviour which is staying for now. Although I think there's a separate decision to be made about whether we always open links in the same tab in future.

To limit this to just the external link behaviour I think the aria-label for new tabs is just fine for now.

davidrapson:
Paired with @marianayovcheva on testing the required updates for this in product code and we're happy with the result. Merging this now.

PR #3204, opened 2024-02-20
Feat/license

davidsauntson:
Adds a licence and a notice file, in line with the internal policy on public repos.


davidrapson (approved):
📜 Thanks so much for sorting this

davidsauntson:
> 📜 Thanks so much for sorting this

No worries 😊 I think there is probably also something to do with the gem here, but I'm not as sure since we aren't really publishing it anywhere. Anyway - one for a different PR

PR #3219, opened 2024-02-28
feat: updating targeted content icon

marianayovcheva:
After A/B testing, it was decided to change the targeted content icon from plus/minus to chevron which rotates on click. The icon gets added with javascript so it was only necessary to update the ts and scss files.
It is possible to see what this will look like in the preview page: https://65df0885414eff00811fd871--citizens-advice-design-system.netlify.app/components/targeted-content/


davidrapson (approved):
👍 Happy with this, nice side-effect that the implementation is a little simpler now. Could we update the changelog too?

marianayovcheva:
Thanks @davidrapson! Added to changelog now, I always forgot those 🤦‍♀️. I'll see if I can add a reminder to the default PR description

PR #3220, opened 2024-02-28
update footer component fallback legal text to reflect registered office move

cnorthwood:
This has been implemented to check the current date and return the new registered office from moving day onwards.

### DIFF
--- a/engine/config/locales/cy.yml
+++ b/engine/config/locales/cy.yml
@@ -10,6 +10,7 @@ cy:
     footer:
       copyright: Hawlfraint ©%{year} Cyngor ar Bopeth. Cedwir pob hawl.
       legal_summary: 'Cyngor ar Bopeth yn enw gweithredol ar Gymdeithas Genedlaethol Cyngor ar Bopeth. Rhif elusen gofrestredig 279057. Rhif TAW 726 0202 76. Cwmni cyfyngedig drwy warant. Rhif cofrestredig 0143694 Lloegr. Swyddfa gofrestredig: Cyngor ar Bopeth, 3ydd Llawr Gogledd, 200 Aldersgate, Llundain, EC1A 4HD'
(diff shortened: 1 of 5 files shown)


engine/config/locales/cy.yml
```diff
@@ -10,6 +10,7 @@ cy:
     footer:
       copyright: Hawlfraint ©%{year} Cyngor ar Bopeth. Cedwir pob hawl.
       legal_summary: 'Cyngor ar Bopeth yn enw gweithredol ar Gymdeithas Genedlaethol Cyngor ar Bopeth. Rhif elusen gofrestredig 279057. Rhif TAW 726 0202 76. Cwmni cyfyngedig drwy warant. Rhif cofrestredig 0143694 Lloegr. Swyddfa gofrestredig: Cyngor ar Bopeth, 3ydd Llawr Gogledd, 200 Aldersgate, Llundain, EC1A 4HD'
+      legal_summary_new_address: 'Cyngor ar Bopeth yn enw gweithredol ar Gymdeithas Genedlaethol Cyngor ar Bopeth. Rhif elusen gofrestredig 279057. Rhif TAW 726 0202 76. Cwmni cyfyngedig drwy warant. Rhif cofrestredig 0143694 Lloegr. Swyddfa gofrestredig: Cyngor ar Bopeth, 3ydd Llawr, 1 Easton Street, Llundain, WC1X 0DW'
```
cnorthwood, line 13: I have made a sensible guess as to to the Welsh variant. Handily we're staying on the 3rd floor.

davidrapson (approved):
👍 Love this approach and happy with the judgement call on the welsh translation (we might want to get a sense check on it but happy for us to do that after the fact).

I think @marianayovcheva was going to do a release tomorrow so up to you both if you are happy for that to go into that release or wait.

cnorthwood:
rebased to fix Changelog clash - think it'd be good to get this into the next release so we can roll it out unless you've got any objections @marianayovcheva ?

davidrapson:
🚀

marianayovcheva:
Just had a thought that we probably need to set reminders and let people in the design-system channel know that on 15.03 if their product has VRTs those will likely fail (the VRTs don't always block the build) and will need updating.

PR #3225, opened 2024-03-01
fix: commonise legal summary

marianayovcheva:
Adding `legal_summary_default` method which is reused in `legal_summary_feedback` (if no `legal_summary` slot is added) and if a `legal_summary` slot exists but is empty.

### DIFF
--- a/engine/app/components/citizens_advice_components/footer.rb
+++ b/engine/app/components/citizens_advice_components/footer.rb
@@ -7,7 +7,7 @@ module CitizensAdviceComponents
     renders_one :feedback_link, "FooterFeedbackLink"
 
     renders_one :legal_summary, lambda { |text|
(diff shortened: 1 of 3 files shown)


engine/app/components/citizens_advice_components/footer.rb
```diff
@@ -7,7 +7,7 @@ class Footer < Base
     renders_one :feedback_link, "FooterFeedbackLink"
 
     renders_one :legal_summary, lambda { |text|
-      text.presence || t("citizens_advice_components.footer.legal_summary")
```
marianayovcheva, line 10: This PR is mainly trying to resolve instances where a legal summary is set but empty. At the moment they wouldn't get the updated office move text.

cnorthwood (approved):

cnorthwood:
the logic here is surprisingly complex from what it looked like at first glance 🤔

PR #3255, opened 2024-03-25
Remove conditional footer logic

davidrapson:
The office move date has now passed we can remove the conditional logic and only ship the new office address.

### DIFF
--- a/engine/app/components/citizens_advice_components/footer.rb
+++ b/engine/app/components/citizens_advice_components/footer.rb
@@ -54,22 +54,11 @@ module CitizensAdviceComponents
     end
 
     def legal_summary_text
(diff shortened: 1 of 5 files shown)


engine/app/components/citizens_advice_components/footer.rb
```diff
@@ -54,22 +54,11 @@ def feedback_url_deprecation
     end
 
     def legal_summary_text
-      legal_summary.presence || legal_summary_fallback.presence
-    end
-
-    def legal_summary_fallback
-      return if legal_summary
-
-      legal_summary_default
```
davidrapson, line 63: This logic was redundant so I've inlined the method.

cnorthwood (approved):

PR #3257, opened 2024-03-25
Remove deprecations

davidrapson:
This removes the following deprecations from component code:

- deprecated feedback_url from footer
- deprecated content slot from section links
- deprecated title attribute from callout

We deprecated all of this a while a go and to my knowledge our application code has since been updated so these should be safe enough to remove now.


cnorthwood (approved):
guessing this should also go in a semver-major release alongside #3276 ?

PR #3260, opened 2024-03-25
Simplify typescript set up and package exports

davidrapson:
Currently we have a couple of papercuts with our packaged javascript modules

1. The method of importing and initialising them is inconsistent
2. The typescript configuration needed is relatively complex for a small number of modules

This PR attempts to rationalise that so that we have a consistent pattern for importing and initialising all modules and as a side-benefit a simpler typescript configuration.

- Defines a top-level tsconfig.json that works for all modules and removes any per-module configuration needed
- Adds consistent index.ts files to every module with a flexible set of default and named exports
- Adds a wrapper method to greedy nav to allow it to be initialised in the same way as all other modules

The goal is to support the existing import paths whilst enabling a consistent preferred method using named exports.

I've included a `lib/index.ts` file and a `main` directive in the `package.json` so you can import from the top-level package entrypoint which is the ideal route:

```js
import { initModuleA, initModuleB } from "@citizensadvice/design-system"
```

Alternatively you can import directly via

```js
import { initMyModule } from "@citizensadvice/design-system/lib/my-module"
```

For backwards compatibility all existing import paths are supported too.


tsconfig.json
```diff
@@ -1,26 +1,15 @@
...
+    "outDir": "./lib",
+    "module": "CommonJS",
+    "target": "es5",
+    "strict": true,
+    "declaration": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true
+  },
```
davidrapson, line 12: This is a reduction of the main options that were in tsconfig-base and the sub-module configs.

cnorthwood (approved):

PR #3261, opened 2024-03-25
Update on this page and disclosure components to use SVG icons

davidrapson:
## Background

We currently ship two different sets of icons within the design system. A new SVG icon set and the previous icon font. We've long wanted to remove the icon font completely but to do that we need to ensure that no internal components are using icon fonts.

This pull request makes steps to remove them from two related components: On this page and disclosure.

### Inline toggle code

This is the biggest surface change. The on this page and disclosure components previously shared code via a utility method. In retrospect this makes it quite hard to change each component independently and in practice the disclosure component had subtly different behaviour which had to be patched in separately.

Practically these are both an implementation of https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/, much like targeted content is so it's reasonable to duplicate the pattern here to keep things flexible.

To account for this an additional cypress test has been added to cover the on this page toggle behaviour. This is preferable to the previous test anyway as it checks against a real component example rather than the lower level fixture used by the utility code.

### Use SVG icons in place of icon fonts

For both components the same pattern has been used to make use of the `PlusMinus` icon component. Like with previous versions of this move the SVG icons are slightly thinner than the icon versions so there is a small visual change.


engine/spec/components/citizens_advice_components/disclosure_spec.rb
```diff
@@ -72,7 +72,7 @@
       ).with_content("Example content")
     end
 
-    it { is_expected.to have_css "[data-toggle-target-id='my-id']" }
+    it { is_expected.to have_css "[aria-controls='my-id']" }
```
davidrapson, line 75: This has been switched out as the component no longer needs `data-toggle-target-id` to work. Testing against aria attributes is preferable too.

scss/6-components/_disclosure.scss
```diff
@@ -1,33 +1,44 @@
-/** @define disclosure **/
+// ============================================================================
+// Disclosure
+// ============================================================================
+
+/** @define disclosure; weak **/
 .cads-disclosure {
-  &__details {
+  .cads-disclosure__details {
```
marianayovcheva, line 7: Does this change make the selectors more specific? If we don't need the change in specificity, then can we get the classes outside of the `.cads-disclosure` wrapper?

demo/visual-regression/backstop_data/bitmaps_reference/backstop_default_Components_Disclosure_fallback_0_document_0_small.png
marianayovcheva, line 1: It's hard to tell the difference between the images, but were they flagged because of small spacing changes?

marianayovcheva:
Yay for moving away from the icon font! 🎉  Looks really good!
A wild opinion - I think we don't necessarily need to get rid of the show-hide util. When I was comparing the two files (disclosure.ts and on-this-page.ts) I noticed that the main difference was the disclosure has the `setSummaryText` function. We could update show-hide with the tidied up common functionality and add the differences in the files. That being said, I don't know if there will be more differences between the two in the future, and the separate approach is also good.
Leaving open for David and Chris to also have a look.

demo/visual-regression/backstop_data/bitmaps_reference/backstop_default_Components_Disclosure_fallback_0_document_0_small.png
davidrapson: Yeah it's slightly different mostly owing to the fact that the SVG icons sit in a square bounding box vs a text bounding box and the proportional sizing being a little different. It's probably worth me taking another pass to see if I can get them looking even closer, but owing to the inherent differences as long as we are happy it looks good I think we're OK. Don't want it to be wildly different though!

scss/6-components/_disclosure.scss
davidrapson: Ah yeah it does, happy to move it out. I changed this short hand selectors are a bit of a nightmare for refactoring without expanding things out in your head, but we probably don't want the extra specificity.

davidrapson:
> A wild opinion - I think we don't necessarily need to get rid of the show-hide util

My (loosely held) but strong preference is to keep them separate, especially as we only have two examples of _slightly_ similar code. Not against extracting it again in future though. I'm very much optimising for "make the components easy to change" vs "share the most code" here.

demo/visual-regression/backstop_data/bitmaps_reference/backstop_default_Components_Disclosure_fallback_0_document_0_small.png
davidrapson: After checking and unless I'm losing the plot I think the text is slightly _further away_ than the icon now which I'm happy with given the component itself looks proportional relative to the new icon.

davidsauntson (commented):
Nice to be closer to ditching the icon font 🙌🏻 

We use the abstracted show/hide quite a lot in content platform.  I think will cause a lot of code duplication in that codebase, but we can abstract it out there instead if we need to.

I'm in favour of not abstracting stuff unless it's useful - I think this is useful but it's not a strongly held opinion.  Was changing to an icon font difficult because of the shared show/hide code, or has it made something else harder to change somewhere?

davidrapson:
Moving this back to draft as it it needs updating and I'd like to resolve the outstanding question on the show/hide helper 

(My stance is that we should ditch it in the long term as it still relies on the icon font and is too generic/internal for consumers to be relying on but we should reinstate the helper in this PR so it doesn't need to block this going in)

davidrapson:
Update this to bring it up to date with `main`.

Refreshed my memory a bit on where we left things on the show-hide utility and checked a couple more things:

The existing utility ultimately works based on the old icon font which a) we want to remove and b) works very differently to inline SVG icons. I’m not entirely sure what a generic utility for this would look like for those 🤔

I’d wondered if public-website was relying on this but it turns out it has it’s own copy of show-hide and doesn’t load the one here anyway which makes things simpler (maybe that came first?) so I think we’re safe to remove it as nothing is using and it would need a significant re-think to be useful with the SVG icon setup.

We should see if a suitable alternative comes out of removing icon font uses from public website if it turns out there is a revised version of this utility that works.

Does that sound good to folks?

davidsauntson (approved):
Neither of the energy apps use the show-hide either 👍🏻

PR #3262, opened 2024-03-26
Add option to disable icon font styles

davidrapson:
Adds a sass config option to disable icon font styles. Intended to be used to help us flush out any remaining uses of icon fonts both internally and within applications.


davidrapson:
Closing in favour of https://github.com/citizensadvice/design-system/pull/3296

PR #3265, opened 2024-03-27
Refactor greedy navigation

davidrapson:
## Summary

Introduces a new version of the greedy navigation component with greatly simplified code and improved behaviour.

## Motivation for change

The greedy navigation component is one of our oldest components and was adapted from an external component. It's long been a pain point of the application as it is hard for the team to modify leading to patch work amends and redundant behaviour being maintained.

The main motivation for change is that the component uses the icon font for the toggle button and we'd like to get rid of this in favour of SVG icons for all components.

This requires change. The component is extremely hard to change so we should make the change easy by restructuring the component to only contain the behaviour we need and to match the rest of our components.

The old component also has a completely different API to the rest of our components so this is a good opportunity to make that consistent.

## Improvements

Restructuring the component enables a number of improvements:

1. Switches the icon to an SVG version (the original motivation for change)
2. Improves keyboard and mouse support (tabbing and clicking in and out of the menu now behaves more consistently)
3. Minimises flash of unstyled content (fix ported from the public website)
3. Bundles welsh translations for UI text. Previously this had to be provided as part of the component initialisation (translations taken taken from the public website).
4. Fixes mistranslation of aria-labels (in production when translated they are part welsh part english)
5. Removed significant amounts of redundant behaviour that was brought over from the module we modified.

## Keyboard support

The old component had partial keyboard support but it was patchy. You could trigger the menu via keyboard but in a non-default way by tabbing in (but only in one direction). The component used a mouse up event instead of a typical click so you couldn't control the button with the space bar like a regular button.

The new behaviour is:

1. The more button acts as a regular button and can be controlled with the space bar
2. When the navigation is open you can tab out of it in either direction and the navigation should close (existing behaviour but should work a bit more predictably in both directions)
3. Clicking anywhere outside the navigation should close the navigation (existing behaviour)
3. Pressing the escape key should close the navigation (existing behaviour)

## Configuration free

The new component is configuration free. The old component provided a number of configuration options many of which were partially implemented or redundant.

The exception being label text which is now translated as part of the Rails template.

The new component is initialised like all our other components.

```js
import { initNavigation } from '@citizensadvice/design-system/lib';
initNavigation();
```

## Testing strategy

Previously we had a mix of tests split across jest and cypress. The jest tests ran assertions against internal methods rather than using testing library helpers. Using jest for component tests is limited as they have to work against static fixtures and don't run in a real browser. Our cypress tests run in a real browser and against real examples but previously only covered some behaviour but not the actual greedy nav logic.

I've taken the approach of keeping a basic jest test around to make sure the component isn't completely broken and then pushed the majority of the test scenarios into the cypress tests. These now cover the logic around moving items into the greedy navigation, keyboard interaction (within the limit of cypress' abilities), and translation behaviour.

## Backwards compatibility

To make it easier to migrate to the new component I've kept the old entrypoint around which wraps the new implementation. This allows us to ship the new version and then remove the old entrypoint in a subsequent version.

---

The exploratory work for this was done in https://github.com/citizensadvice/design-system/pull/3264 in case you want to see the process. Depends on https://github.com/citizensadvice/design-system/pull/3260


demo/backstop-common.js
```diff
@@ -112,7 +112,7 @@ module.exports = function backstopCommon(baseUrl) {
         viewports: [{ label: 'desktop', width: 800, height: 200 }],
       },
       {
-        label: 'Components/Header',
+        label: 'Components/Header (search open)',
```
davidrapson, line 115: This previously had the same name as another scenario, surprised it didn't clash before!

demo/app/assets/stylesheets/previews.scss
```diff
@@ -7,7 +7,8 @@
 }
 
 .component-preview--breadcrumbs,
-.component-preview--navigation {
+.component-preview--navigation,
+.component-preview--header {
```
davidrapson, line 11: Updated these rules so the header component examples also have their margin removed. This is why the header visual regression snapshots have updated.

scss/6-components/_navigation.scss
```diff
@@ -1,7 +1,52 @@
 @use 'sass:map';
-@import 'navigation_greedy-nav';
 
-/** @define navigation */
+// ============================================================================
```
davidrapson, line 3: I've moved all the navigation styles including single-use mixins into this file as they were previously split accross three.

src/ts/navigation/navigation.test.ts
```diff
@@ -0,0 +1,81 @@
+/**
```
davidrapson, line 1: I've kept this around as a basic initialisation test but honestly given the cypress tests cover everything anyway and are much better suited I'm inclined to delete this completely? How do folks feel about that? It's a pretty fake test as it stands.

src/ts/navigation/navigation.ts
```diff
@@ -0,0 +1,320 @@
...
+
+function computeContentWidth(containerEl: HTMLElement) {
+  const styles = window.getComputedStyle(containerEl);
+  const padding =
+    parseFloat(styles.paddingLeft) + parseFloat(styles.paddingRight);
+
+  return containerEl.clientWidth - padding;
+}
+
+function computeChildrenOffsetWidth(containerEl: HTMLElement) {
```
davidrapson, line 186: This used to have a 10 pixel offset but the rationale in the comment wasn't clear what benefit it gave. We can add it back if others disagree.

demo/cypress/e2e/navigation.cy.js
```diff
@@ -1,55 +1,147 @@
...
-  it('can close the dropdown menu by clicking outside it', () => {
-    openNavigation();
-    assertNavigationOpen();
-    cy.get('body').click();
-    assertNavigationClosed();
+  // Note: Tab interaction is a bit fraught to test in Cypress
+  // We use https://github.com/dmtrKovalenko/cypress-real-events
+  // to approximate this but full keyboard testing should still be
+  // done manually when changing this behaviour.
+  context('when interacting with a keyboard', () => {
```
davidsauntson, line 95: Should we also test tabbing into it?  If it's too fraught at least some sort of note to test that behaviour manually too?

design-system-docs/frontend/styles/_base.scss
```diff
@@ -7,17 +7,6 @@ body {
   height: 100%;
 }
 
-// https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/
```
davidsauntson, line 10: was this causing problems with the navigation?

src/ts/navigation/navigation.test.ts
davidsauntson: I think getting rid of it is fine

src/ts/navigation/navigation.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    if (child.nodeType !== 3 && !Number.isNaN(child.offsetWidth)) {
+      sum += child.offsetWidth;
+    }
+  }
+
+  return sum;
+}
+
+function doesItFit(containerEl: HTMLElement, breaks: number[]) {
```
davidsauntson, line 201: What is breaks here?  I think it has something to do with the cumulative widths of each nav link but it would be good to make it explicit (in a comment would be fine I think).

In fact some documentation for this function would be good - it's hard to infer the behaviour from the code fully.

Also it seems to do more than answer the question 'does it fit' - should it be broken up or renamed?

scss/6-components/_navigation.scss
```diff
@@ -73,43 +282,58 @@
...
+    margin-left: $cads-spacing-4;
   }
 }
 
-@media print {
-  .cads-navigation {
-    display: none;
+////
+// Prevent overflowing navigation when loaded
+// or when javascript is disable
```
davidsauntson, line 327: `disabled`

davidsauntson (commented):
This looks good to me.  Some minor comments, it would be good to document `doesItFit()` a bit more (and maybe call it something more appropriate (it seems to hide / show elements too - but I might have misunderstood!).

design-system-docs/frontend/styles/_base.scss
davidrapson: Yes as a side-effect of porting over the overflow fix from public-website rather than the refactor itself.

demo/cypress/e2e/navigation.cy.js
davidrapson: I'll need to double check again, I think the answer is still that cypress doesn't handle that well so yes maybe a document that is manual testing steps for greedy nav would be good.

src/ts/navigation/navigation.ts
davidrapson: Good shout.

Short answer: Agreed, let's break this down more. 

Longer answer: For context the the process I followed was basically to copy over the existing file and names and gradually refactor the implementation, rather than a ground-up rewrite. I'd bet on this being part that retains the most naming of the original.

davidsauntson: OK - I'm happy for it to stay like this then and maybe this can be a smaller refactor later down the line.  It would be good to have another look at it at some point tho.

davidsauntson (approved):
Happy for `doesItFit` to be sorted out in another refactor later 👍🏻

davidrapson:
I've brought this back up to date with the intention of merging in the new year.

PR #3296, opened 2024-04-15
Improve font loading options

davidrapson:
## Background

Our SCSS styles include `@font-face` declarations which are hard-coded with relative paths to the assets distributed as part of the node module.

Our main target environment is Rails. Historically we've targeted webpack environments via [shakapacker](https://github.com/shakacode/shakapacker) and used to rely on `resolve-url-loader` to resolve these font paths.

This is one way of loading assets in Rails but not the only way. For newer applications we favour using [cssbundling-rails](https://github.com/rails/cssbundling-rails) which ultimately delegates to the Rails asset pipeline using sprockets or propshaft.

Ultimately we want to provide a way of getting the font files into the right place so that applications can correctly resolve paths.

## Using with Rails

We can use the following config to tell Rails to add the font-files to the load path.

```
Rails.application.config.assets.paths << Rails.root.join("node_modules/@citizensadvice/design-system/assets/fonts")
```

With this configuration running `bin/rails assets:clobber assets:precompile` results in something like the following:

```
├── assets
│   ├── application-8e168b94c2bc67230117de79ddaca9564a41658ef756151b9a28d03c89b61c8c.js
│   ├── application-8e168b94c2bc67230117de79ddaca9564a41658ef756151b9a28d03c89b61c8c.js.gz
│   ├── application-db8938456959929d9030c2eae5218e56479fad40acad5379ad7d6edf1582fc84.css
│   ├── application-db8938456959929d9030c2eae5218e56479fad40acad5379ad7d6edf1582fc84.css.gz
│   ├── application.js-5a2e1d51537bf860487a41faa12cf05f60b275991d5657fbfd3edd68e7c8251e.map
│   ├── application.js-5a2e1d51537bf860487a41faa12cf05f60b275991d5657fbfd3edd68e7c8251e.map.gz
│   ├── cads-9ff379853966eee0177bcf4cba11c485c3fa736c75272d51b7794df8796ebf53.eot
│   ├── cads-9ff379853966eee0177bcf4cba11c485c3fa736c75272d51b7794df8796ebf53.eot.gz
│   ├── cads-b805314b26ac59bc2d4020c7fc067bffe328726c04864597ed5b0fdc80fe26cf.ttf
│   ├── cads-b805314b26ac59bc2d4020c7fc067bffe328726c04864597ed5b0fdc80fe26cf.ttf.gz
│   ├── cads-c2b3f17b3d29b4d02cc77ebe23677db5ba63d2244272f5652b95065e8cd3727e.woff
│   ├── cads-c84380ecca9e8171e86947af64e7ae132c747401d2759413af09f36dfd374fdc.svg
│   ├── cads-c84380ecca9e8171e86947af64e7ae132c747401d2759413af09f36dfd374fdc.svg.gz
│   ├── manifest-04024382391bb910584145d8113cf35ef376b55d125bb4516cebeb14ce788597.js
│   ├── manifest-04024382391bb910584145d8113cf35ef376b55d125bb4516cebeb14ce788597.js.gz
│   ├── open-sans-v26-latin-700-c12de74c75388b99217911061da4f485071392d528338b1f7052007ac0eb396b.woff2
│   ├── open-sans-v26-latin-700-c38bf4cc3dbdb0cfcdf27774e9b9ccbc58bace1597fa450a15b9e04aabc7f24f.woff
│   ├── open-sans-v26-latin-regular-21463e822c091d3350603d57f83c01f777c95280f399ed2ec3a427ce959d9327.woff
│   ├── open-sans-v26-latin-regular-2d8f2b8387a095f3e77a79e59b6026ce3ec5a3f02a62ffafd652496d0e331006.woff2
│   ├── previews-02c789845af6a3e980702f4d711002e25cb67c37c0f50d71d0d66d2c2fcdd533.css
│   └── previews-02c789845af6a3e980702f4d711002e25cb67c37c0f50d71d0d66d2c2fcdd533.css.gz
```

This is what we want, fonts making their way into the asset pipeline, but it's not enough for Rails to correctly replace the asset url as the hard-coded path points elsewhere.

We can provide a configuration option (as `$cads-font-path`) to allow customising this to the needs of the application. In the example above the following works:

```scss
$cads-font-path: ".";

@import "@citizensadvice/design-system/scss/lib";
```

I've also tested this using propshaft by following https://github.com/rails/propshaft/blob/main/UPGRADING.md#3-migrate-from-sprockets-to-propshaft which works too as it's a largely drop-in replacement.

## Changes

The upshot is that we provide a new `$cads-font-path` configuration option. The rest of the changes are in documenting and testing that this works in a standard Rails context.

I've also bundled a change from https://github.com/citizensadvice/design-system/pull/3262 to allow disabling icon fonts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Remove a number of deprecations from the view components.
 - Remove deprecated feedback_url from footer
 - Remove deprecated content slot from section links
 
+### Improve control over font loading
+
+Provides two new configuration option when using the SCSS styles:
+
+- `$cads-font-path` to provide a custom path to control font loading in your application
+- `$cads-enable-icon-font` to enable or disable the legacy icon font (defaults to `true`)
+
 ### Consistent javascript module imports
 
 Allow all modules to be imported using a consistent interface, either via:
--- a/demo/app/views/layouts/application.html.erb
+++ b/demo/app/views/layouts/application.html.erb
@@ -4,9 +4,9 @@
     <meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Demo</title>
-    <link href="/assets/fonts/open-sans-v26-latin-700.woff2" rel="preload" as="font" type="font/woff2" crossorigin="true" />
-    <link href="/assets/fonts/open-sans-v26-latin-regular.woff2" rel="preload" as="font" type="font/woff2" crossorigin="true" />
-    <link href="/assets/fonts/cads.woff" rel="preload" as="font" type="font/woff" crossorigin="true" />
+    <link href="<%= asset_path("open-sans-v26-latin-700.woff2") %>" rel="preload" as="font" type="font/woff2" crossorigin="true" />
+    <link href="<%= asset_path("open-sans-v26-latin-regular.woff2") %>" rel="preload" as="font" type="font/woff2" crossorigin="true" />
+    <link href="<%= asset_path("cads.woff") %>" rel="preload" as="font" type="font/woff" crossorigin="true" />
     <%= csp_meta_tag %>
     <%= stylesheet_link_tag "application", media: "all" %>
     <%= yield :custom_head %>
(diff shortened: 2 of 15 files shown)


davidsauntson (approved):
🙌🏻 

Is it worth making the default value of the asset path what it is at the moment?

davidrapson:
> Is it worth making the default value of the asset path what it is at the moment?

Good question, maybe yeah 🤔 will try that and see if anything breaks

PR #3303, opened 2024-04-18
Fix docs build

davidrapson:
The change in https://github.com/citizensadvice/design-system/pull/3296 accidentally broke the docs build as it needs the old relative path to assets and our CI step didn't do a full production build which would have failed this check. The default build command only builds the static content, the deploy command also runs a clean and a frontend build.


davidrapson:
Merging this straight in to fix the build

PR #3335, opened 2024-05-17
Fix JS documentation for `ErrorSummary` component

leeky:
Update the code snippet for including the JS code to support the ErrorSummary component.

I tried the original snippet, and it didn't work for me. The corrected version is [currently in use here](https://github.com/citizensadvice/public-website/blob/bdf5b4331382fb165d707f4e2e97c5c5f5eb64e0/app/javascript/packs/public-advice.js#L4), among other places.


davidrapson (approved):
Oh! So the syntax you've changed is actually a recent change https://github.com/citizensadvice/design-system/blob/main/CHANGELOG.md#consistent-javascript-module-imports

But that's not out yet so you're right this example won't work, I should have held off on the docs update 🤔. This syntax will continue to work anyway so this is fine 👍

PR #3342, opened 2024-05-22
Fix/import not use for grid

davidsauntson:
This makes overriding the grid settings possible in products that use the design system.  

At the moment, we are `@use` to import files at higher levels of the scss code, rather than `@import`ing them.  `@use` requires you to explicitly override variables when you import a module, rather than respecting what the global state of variables in the `@use`d file are.  This changes ensures that the global values of the grid are respected at all layers of the ITCSS triangle, allowing products to override the variables to make their own grids.

The wider issue might be affecting other areas of the code, but I'd like to keep this change small and limited to what I know isn't working at the moment.

See https://acab.slack.com/archives/C060YQ75MGD/p1716370930966639?thread_ts=1716291046.823689&cid=C060YQ75MGD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@ We historically bundled a compiled version of the CSS as lib.css in the distribu
 
 The new office address is now the default in the footer component and the old address has been removed.
 
+**Fix**
+- Make grid configurable by products
+
 ## v5.8.1
 
 ### 8 March 2024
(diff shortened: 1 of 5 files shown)


davidrapson (approved):
👍 LGTM

For when we merge this: We might want to think about whether we do a `5.9.0` release with this in it or push forward with `6.0.0`, we can pull some of the more semi-breaking changes out if we need to.

davidsauntson:
> 👍 LGTM
> 
> For when we merge this: We might want to think about whether we do a `5.9.0` release with this in it or push forward with `6.0.0`, we can pull some of the more semi-breaking changes out if we need to.

This isn't urgent so I'm happy for it to go into 6.0.0.  Perhaps when #3265 and #3261 are in I can do a 6.0.0-alpha release?

PR #3361, opened 2024-06-05
Apply latest rails patch

davidrapson:
Batches up the latest rails patch across the engine, docs, and demo app code to resolve a dependabot warning.


marianayovcheva (approved):
LGTM!

PR #3362, opened 2024-06-05
Patch bridgetown compatibility with view_component 3.9.0+

### LINKED ISSUE #3145: Bridgetown fails to build against view_component 3.9
The bridgetown project fails to build when updated to view_component 3.9. We should investigate the cause and potentially create a reduced test case to confirm if the issue can be replicated using a fresh bridgetown install.

Related issue: https://github.com/bridgetownrb/bridgetown-view-component/issues/8

davidrapson:
ViewComponent 3.9+ introduced a change which depends on there being an active Rails request or controller in order to conditionally escape HTML. This causes an issue when run in bridgetown as there's no request and we're always rendering static html.

Whilst this is monkey patching an internal method this feels like the least worst approach as it removes just the single check that's dependent on the request whilst still keeping the underlying escaping.

(Temporarily) fixes #3145. Related issue: https://github.com/bridgetownrb/bridgetown-view-component/issues/8

### DIFF
--- a/design-system-docs/Gemfile
+++ b/design-system-docs/Gemfile
@@ -13,8 +13,10 @@ gem "puma"
 
 gem "citizens_advice_components", path: "../engine"
 
-# https://github.com/citizensadvice/design-system/issues/3145
-gem "view_component", "3.8.0"
+# citizens_advice_components bundles view component but we
+# also use this to write docs components so explicitly
+# name it as a dependency.
+gem "view_component", "~> 3.9"
 
 group :development, :test do
   gem "citizens-advice-style",
--- a/design-system-docs/plugins/builders/cads_builder.rb
+++ b/design-system-docs/plugins/builders/cads_builder.rb
@@ -13,6 +13,27 @@ I18n.load_path << Dir[
 
 Time.zone = "London"
 
+# ViewComponent 3.9+ introduced a change which depends on there being an active Rails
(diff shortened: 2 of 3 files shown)


design-system-docs/Gemfile
```diff
@@ -13,9 +13,6 @@ gem "puma"
 
 gem "citizens_advice_components", path: "../engine"
 
-# https://github.com/citizensadvice/design-system/issues/3145
-gem "view_component", "3.8.0"
```
leeky, line 17: Should the PR include a note for why you've removed the `view_component` dependency? Is `citizens_advice_components` now using 3.9.0+ internally?

design-system-docs/plugins/builders/cads_builder.rb
```diff
@@ -13,6 +13,27 @@
 
 Time.zone = "London"
 
+# ViewComponent 3.9+ introduced a change which depends on there being an active Rails
```
leeky, line 16: 👍

leeky (approved):
LGTM! :shipit:

design-system-docs/Gemfile
davidrapson: It will use the latest version of view component (we technically support both 2.x and 3.x) but/and yep let me add a note in the commit and PR description.

PR #3394, opened 2024-07-09
Add the summary list component to the design system

cnorthwood:
This is a component that currently exists in public website (you can see it in use on [a local office job ad](https://cdn.loris.content.citizensadvice.org.uk/about-us/job-and-voluntary-opportunities/job-opportunities-with-local-citizens-advice/25L98R87XAQDhyavCCVU5I/)) which are wanting to also use to display the questions and answers on the review page when submitting the complaints form, so wish to promote to the design system to promote code re-use.


design-system-docs/src/_component_docs/summary-list.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Summary list
+---
+
```
marianayovcheva, line 4: This is a simple component but is there a paragraph we can include at the start to say where it can be used? I'm aware not of our components have this (they probably should 😬) but I wonder if the design team can help with a definition of the use cases for the component (e.g to be used to summarise form data, etc). It can be a part of a separate PR though.

marianayovcheva (approved):
LGTM! 🎉 
Added a couple of comments but they are not blocking.
I'm also selfishly keen to get this component in the design system as I might need it for another form 🙂

design-system-docs/src/_component_docs/summary-list.md
cnorthwood: yep, I've added this now as turns out our designer had prepared something :)

PR #3396, opened 2024-07-11
Release/v6.0.0

cnorthwood:
This is being done as semver-major as technically there are breaking changes in this release (removal of deprecations and the bundled CSS)


davidsauntson (approved):
6️⃣

PR #3437, opened 2024-08-13
Add Rails 7.2 to test matrix

davidrapson:
https://rubyonrails.org/2024/8/10/Rails-7-2-0-has-been-released

Updates our test matrix to include the new Rails 7.2 release and to account for the new Rails maintenance policy https://rubyonrails.org/maintenance

Explicitly test a "legacy" configuration using an older Ruby, Rails, and ViewComponent version to represent applications that have not been upgraded yet with the expectation that we'll drop Rails 6.1 along with ViewComponent 2.x when Rails 6.1 goes end-of-life on October 1st 2024.


marianayovcheva (approved):
LGTM!

PR #3495, opened 2024-10-01
Add `page_heading` option to text area

johnny-m-young:
## Changes
Adds the `page_heading` option to `Textarea`, `TextInput` and `DateInput` so that you can have the label as a page heading.

For the time being, the `page_heading` option overrides the `optional` option, meaning the `(optional)` part of the label is omitted.

## Rationale
I came across a few use cases for this when building the consumer forms in the Referrals application. I managed to work around it ([see here](https://github.com/citizensadvice/referrals/blob/7b79512576557e0b401647c061ae18a7b1b4b1f5/app/views/consumer/forms/post/steps/_desired_outcome.html.haml#L3)) but it felt a bit hacky. 

Further, the GOV.UK design system [suggests](https://design-system.service.gov.uk/get-started/labels-legends-headings/) nesting the `label` within `h1` tags rather than the other way around (which was the result of my solution).


johnny-m-young:
Interested on your thoughts on this @davidrapson

contributing/testing.md
```diff
@@ -10,6 +12,24 @@ From the project root run:
 npm run lint
 ```
 
+### Ruby linting
```
davidrapson, line 15: Thanks for adding this! I have https://github.com/citizensadvice/design-system/pull/3496 open right now which makes some larger changes to these guides. I've tried to incorporate your changes into there. Mind taking a look and if you are happy pulling these changes out of this PR?

demo/backstop-common.js
```diff
@@ -362,6 +362,11 @@ module.exports = function backstopCommon(baseUrl) {
         url: componentUrlFor(`textarea/value`),
         viewports: [{ label: 'form-field', width: 800, height: 400 }],
       },
+      {
```
davidrapson, line 365: I'd love to see scenarios for each of the fields that have been updated

demo/spec/components/previews/textarea_preview.rb
```diff
@@ -54,6 +54,18 @@ def error
...
   end
 
+  def page_heading
+    render(
+      CitizensAdviceComponents::Textarea.new(
+        name: "example-textarea-page-heading",
+        label: "Example text area with page heading",
+        options: {
+          page_heading: true
+        }
```
davidrapson, line 64: In each of these examples do you mind if we include the `hint` option too for completeness?

davidrapson (changes requested):
This is looking great! I think we'll want to extend this to cover all fields that inherit from `Input`, as the current change will affect any field that uses it. This would be `TextInput`, `DateInput` and (the one you've already covered) `Textarea`.  If you do only want to scope this to the `Textarea` we'll need to narrow the scope of the changes somewhat. Arguably we should include fields like radio and checkbox collections here too but I think that's probably too much for this PR.

I'd also suggest chatting with @leeky on how to integrate this into https://github.com/citizensadvice/rails-form-builder as I think for referrals you ultimately want these changes there too.

For your questions:

1. Yes, the order of the heading and label wrapping makes a big difference. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#headings (will see if I can dig out a more detailed summary)
2. I think we can probably make the call to omit the optional label when the field uses a page heading. We'd need to give that a bit more thought and some extra styles anyway so probably fine to side-step for now.

Great work on this and thanks so much for the contribution!

contributing/testing.md
johnny-m-young: Sorry just catching up after holidays. #3496 looks good, happy to take these out

engine/spec/components/citizens_advice_components/input_spec.rb
```diff
@@ -72,6 +72,19 @@
...
+    before do
+      render_inline described_class.new(
+        name: "example-input-page-heading",
+        label: "Example input with page heading",
+        type: :text,
+        options: { page_heading: true }
+      )
+    end
+
+    it { is_expected.to have_css "h1[class=cads-page-title]", text: "Example input with page heading" }
```
davidrapson, line 85: This takes a CSS selector so it can be just `h1.cads-page-title`

davidrapson (approved):
Lovely work on this 🥳, we'll still need to add this to  https://github.com/citizensadvice/rails-form-builder once this is out.

Once last request before merging, can you add something to the CHANGELOG?

PR #3496, opened 2024-10-02
Improve local setup

davidrapson:
## Background

I've recently needed to set up a new laptop so this gave me the opportunity to run through the setup instructions from scratch. These are a few suggested improvements based on that.

## Changes

The main changes are around improving the contribution guides, which were fine but required you to bounce around a fair bit to get the information you need for a first time setup. Restructure the guides to hopefully flow better.

Beyond that, the remaining changes are around consolidating the set up scripts. Originally, sub-projects defined setup scripts which would not work if you hadn't first run the top-level one…which then ran the sub-project ones.

This is pretty confusing and something we want to avoid. Consolidate the setup into a single top-level setup script so you can run `./setup` at the root of the project, at the expense of removing set up scripts in the sub-projects.


cnorthwood:
just to be sure I understand - instead of a new user running the top-level setup script to set up this repo, they now have to go into each directory and run the setup script for each component individually? thinking about the barrier for entry for people who might make occasional contributions, going from a single setup script to 3 different ones seem to increase that

davidrapson:
Yeah that's a good shout. I basically want to go in one direction or another, the current per-project scripts couldn't really be ran standalone. Might have a fresh stab at a revised top-level script that takes that into account 🤔

davidrapson:
@cnorthwood Added a new commit which takes the opposite approach and consolidates setup into a single top-level setup script so you can run `./setup` at the root of the project for comparison. I think I prefer this option on balance. It's pretty rare that you would need to set up just one component of project.

cnorthwood (approved):
I agree, this definitely looks neater!

davidrapson:
@cnorthwood Great 🙂 

@johnny-m-young Noticed you made an update to the setup docs in https://github.com/citizensadvice/design-system/pull/3495, mind taking a look at this PR as you'll be able to give a fresh take on it. I've tried to incorporate your linting doc changes into here too.

davidrapson:
Going to merge this one today whilst I have a moment, @johnny-m-young feel free to to a retroactive review if you get a moment.

PR #3504, opened 2024-10-10
feat: allowing lang attribute to header links

marianayovcheva:
This one came after an internal accessibility testing on the public site.
It was raised that:
> When scrolling the site with a screen reader (we did that with Jaws) the Cymraeg link at the top of the public website homepage is not specified as being in welsh language. As a result it is not read correctly.

I have checked with Voiceover screen reader on mac and even though the language is specified, it didn't change the pronunciation and unlike other languages, doesn't read numbers in the language specified. I would assume this is due to the Welsh language support on screenreaders - would be interesting to test with Jaws. I have still decided to raise a PR as I think it's a good practice to be able to pass the `lang` attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Header: allow `lang` attribute in the header links slot
+
 ## v6.0.0
 
 ### 11 July 2024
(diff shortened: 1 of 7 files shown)


leeky (approved):
EYDIM! :shipit: 

(edrych yn dda i mi = looks good to me)

PR #3517, opened 2024-10-18
fix: default aria-label for footer feedback link

marianayovcheva:
Currently, if no feedback link title is specified and the feedback link is set to open in a new tab, the `aria-label` is " (opens in a new tab)".
This PR fixes this by using the default feedback link title as an `aria-label` fallback.

### DIFF
--- a/engine/app/components/citizens_advice_components/footer.html.erb
+++ b/engine/app/components/citizens_advice_components/footer.html.erb
@@ -4,12 +4,22 @@
       <div class="cads-grid-row">
         <div class="cads-grid-col">
           <p class="cads-footer__feedback">
-            <%= link_to(
-              feedback_link.title || t("citizens_advice_components.footer.website_feedback"),
-              feedback_link.url,
-              class: "cads-footer__hyperlink js-cads-footer-feedback",
-              **feedback_link.new_tab_attributes
-            ) %>
+            <% if feedback_link.new_tab? %>
+              <%= link_to(
+                feedback_link.title || t("citizens_advice_components.footer.website_feedback"),
+                feedback_link.url,
+                class: "cads-footer__hyperlink js-cads-footer-feedback",
+                target: "_blank",
+                rel: "noopener",
+                aria: { label: "#{feedback_link.title || t('citizens_advice_components.footer.website_feedback')} (opens in a new tab)" }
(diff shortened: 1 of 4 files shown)


engine/app/components/citizens_advice_components/footer.html.erb
```diff
@@ -4,12 +4,22 @@
       <div class="cads-grid-row">
         <div class="cads-grid-col">
           <p class="cads-footer__feedback">
-            <%= link_to(
-              feedback_link.title || t("citizens_advice_components.footer.website_feedback"),
-              feedback_link.url,
-              class: "cads-footer__hyperlink js-cads-footer-feedback",
-              **feedback_link.new_tab_attributes
-            ) %>
+            <% if feedback_link.new_tab? %>
```
marianayovcheva, line 7: I wanted to fix the issue just by adding a conditional for the aria-label in the `.rb` file. However, I cannot use translations there and was getting error: https://www.rubydoc.info/gems/view_component/3.11.0/ViewComponent/TranslateCalledBeforeRenderError. So instead I decided to expose the new_tab_attributes here.
Suggestions for a more elegant solution are welcome.

davidrapson: Yeah that rings a bell, I think what you've done here _is_ the elegant solution as it's part of the template logic anyway.

engine/app/components/citizens_advice_components/footer.html.erb
```diff
@@ -4,12 +4,22 @@
...
+                target: "_blank",
+                rel: "noopener",
+                aria: { label: "#{feedback_link.title || t('citizens_advice_components.footer.website_feedback')} (opens in a new tab)" }
+              ) %>
+            <% else %>
+              <%= link_to(
+                    feedback_link.title || t("citizens_advice_components.footer.website_feedback"),
+                    feedback_link.url,
+                    class: "cads-footer__hyperlink js-cads-footer-feedback"
+                  ) %>
```
davidrapson, line 21: Indentation looks a little wonky here to me compared to the other branch of the if statement, but erblint doesn't seem to mind so only amend it if you feel like it.

davidrapson (approved):
👍 Good spot

PR #3526, opened 2024-10-29
Spike/automerge dependabot updates

stringiest:
We wanted to trial auto-approval & auto-merge of dependabot PRs to see if we can decrease the maintenance burden for developers. During a discussion in Content Platform Dev Refinement, @davidsauntson suggested that the `design-system` repo might be a possible place to test the automation.

An additional benefit might be that security patches get merged in more quickly.

This PR represents an initial draft at auto-approve and auto-merge workflows for dependabot PRs meeting the following criteria:
1. PR is raised by `dependabot [bot]`
2. PR is for a `patch` or `minor` version update only
3. PR is not for a `minor` version update to `rails`

We discussed the need to be able to restrict the times of day that auto-merge could happen, so that potentially-breaking changes wouldn't be merged in out of hours, and to allow for things like Epi-UAT not being available overnight. 

That is perhaps less of a problem for the `design-system`, I have included a restriction blocking dependabot PRs from being merged at weekends, just to confirm that the action works. For other repos we may also wish to block merging over public holidays. 

It is also possible to configure dependabot.yml to schedule the approximate time & day that it checks for updates, e.g.
```
    schedule:
      interval: weekly
      day: tuesday
      time: 09:00
```

The `dependabot-auto-merge.yml` workflow is bound by the usual branch protections for the repo, which probably means that it won't be able to merge any dependabot PRs that haven't also been approved by a code owner. But I think we could still test the overall process by manually approving the PRs too for a while.

N.B. The [GH action](https://github.com/yykamei/block-merge-based-on-time) that I've used to block out of hours merging _does_ have a bank holidays file, but it uses the holidays from Google Calendar API and so thinks that days like Valentine's Day are a bank holiday. We would likely need to manually add bank holidays if we wanted to block merging on those days in other repos.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,8 +7,6 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
-      day: tuesday
-      time: 09:00
 
   ##
   # Top-level package
@@ -17,8 +15,6 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
-      day: tuesday
-      time: 09:00
     groups:
       # Group everything except major version updates and security vulnerabilities
       regular-updates:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,8 +37,6 @@ updates:
     directory: '/engine'
     schedule:
       interval: weekly
-      day: tuesday
-      time: 09:00
     groups:
       # Group everything except major version updates and security vulnerabilities
       regular-updates:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
+      day: tuesday
+      time: 09:00
 
   ##
   # Top-level package
@@ -15,6 +17,8 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
+      day: tuesday
+      time: 09:00
     groups:
       # Group everything except major version updates and security vulnerabilities
       regular-updates:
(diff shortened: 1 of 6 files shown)


.github/dependabot.yml
```diff
@@ -7,6 +7,8 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
+      day: tuesday
+      time: 09:00
```
davidsauntson, line 11: You mentioned it in the description, but I don't think we need to do this for the design system.  It's good to have this as a point of reference for when we do this on other repos that deploy to production on a merge to main

davidsauntson (approved):
LGTM!  Will be interesting to see how this interacts with the codeowners file, as you mentioned.  There is some interesting bits in the readme of this action too https://github.com/hmarr/auto-approve-action?tab=readme-ov-file#code-owners

.github/workflows/block-merge-based-on-time.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  block:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'citizensadvice/design-system'
+    permissions:
+      contents: read
+      pull-requests: read
+      statuses: write
+    steps:
+      - uses: yykamei/block-merge-based-on-time@main
```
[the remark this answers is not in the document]
stringiest: Good point - changed 👍🏻

.github/dependabot.yml
stringiest: Now removed

PR #3537, opened 2024-11-06
Cp 145 grid update

marianayovcheva:
From the original ticket:
> In the process of aligning the Global design system and the casebook one, we would like to request a change on the global design system. 
The change we request is to change the margin and the gutter spacing from 40px to 32px for screen sizes bigger than 1024 to 1090. This is our large screen viewports.

Huge thanks to @sjoswald and @KPobeeNorris for working on and testing the update 🙌


davidsauntson (approved):
↔️

PR #3544, opened 2024-11-22
Update supported Ruby and Rails versions

davidrapson:
This makes a couple of changes to reflect the release of Rails 8 and Rails 6.1 reaching EOL.

- Drop support for Rails 6.1 both in the gemspec and in the test matrix
- Add support for Rails 8
- Remove old versions of view component from the test matrix, we've long stopped using v2 in production
- For local development, moves the .ruby-version to the top-level directory and updates the project to use a more recent 3.3 patch version


akamike (approved):
Looks good!

PR #3555, opened 2024-12-05
Prepare v6.3.0 release

davidrapson:
Adds support for Rails 8 and removes support for EOL Rails 6.1.

There's an argument removing support for Rails 6.1 is a breaking change but we have no consumers using this version so recommending we do this as a minor version as there are otherwise no other changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v6.3.0
+
+### 5 December 2024
+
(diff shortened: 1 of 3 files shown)


marianayovcheva (approved):
🚀

davidrapson:
Will release this in the morning now 🚃 8️⃣

PR #3564, opened 2024-12-11
Summary list refinement

davidrapson:
Makes a few quality of life improvements to the summary list component now that we're using it in some more products.

**Refine summary list styles**

Restructures the summary list styles to simplify the padding around elements and avoid the need to override as much by using the lower level typographic scale mixin rather than an extend.

Defines a margin for the top-level component which is in keeping with our spacing scale. The inherited value was the default for definition lists which was tight when combined with other elements.

**Account for paragraphs in summary lists**

Adds some styles to accommodate the scenario where a summary list contains content that comes from a CMS and wrapped in a `cads-prose` container where there may be paragraph elements in the summary list.

**Allow skipping rows with blank values**

Adds a new `hide_empty_rows` option to remove rows with blank values


akamike (approved):
Looks good! I do wonder if there is something to inverting the blank options name (from `reject_blanks` to `allow_blanks`) if only to align it with similarly named Rails options which may make it more intuitive. Not essential, just a thought 😄

davidrapson:
Went with `hide_empty_rows` in the end to favour something more display-oriented than data-oriented.

PR #3580, opened 2025-01-17
new feature: allow footer columns to have text in addition to links

cnorthwood:
This is needed to support the local office minisites which have this format.

Example

Screenshot 2025-01-17 at 15-07-16 Demo


davidsauntson (commented):
Looks good!

Pls could you add the `with text` example to the documentation site too - into the `.md` file here https://github.com/citizensadvice/design-system/blob/main/design-system-docs/src/_component_docs/footer.md

Then it will appear in the footer component page.

davidsauntson (approved):
🙌🏻

PR #3581, opened 2025-01-17
Pin concurrent-ruby version for tests

davidrapson:
Workaround for
https://github.com/ruby-concurrency/concurrent-ruby/issues/1077 which is currently causing our Rails 7.0.x CI tests to fail.

This change shouldn't affect the published gem as this only controls what we install locally or in CI and consumers control their own application gem versions.

Can be removed when either a Rails 7.0.x patch is released or we remove support for Rails 7.0.x whichever is sooner (likely the latter as it is near EOL)

### DIFF
--- a/engine/Gemfile
+++ b/engine/Gemfile
@@ -22,6 +22,12 @@ end
 # rubocop:enable Bundler/DuplicatedGem
 
 group :development, :test do
+  # Can be removed when either a Rails 7.0.x patch is released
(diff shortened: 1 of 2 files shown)


davidrapson:
Assuming this works, unblocks https://github.com/citizensadvice/design-system/pull/3580

leeky (approved):
LGTM! :shipit:

PR #3603, opened 2025-02-10
v6.4.1-alpha.0

davidsauntson:


cnorthwood:
out of interest, why an alpha release rather than a regular point one?

davidsauntson:
> out of interest, why an alpha release rather than a regular point one?

I'm in the habit of doing an alpha release to test it in my product first, and here I'd like to check I've made the breakpoints overridable correctly.  Been a while since I did a design system release though so perhaps that isn't standard anymore.

cnorthwood:
ah maybe I've just never done that, tbh I've done this by linking the two locally rather than going through a release so I wonder if that's preferrable, and especially good for reducing cycle time if it doesn't work? https://github.com/citizensadvice/design-system/blob/main/contributing/03-testing.md#testing-with-your-application

davidsauntson:
> ah maybe I've just never done that, tbh I've done this by linking the two locally rather than going through a release so I wonder if that's preferrable, and especially good for reducing cycle time if it doesn't work? https://github.com/citizensadvice/design-system/blob/main/contributing/03-testing.md#testing-with-your-application

I have already done that too, with the design-system I generally like to run things through production before doing a non-alpha release.  As I say, it's habit more than anything and perhaps it's over-cautious.  If you think it's a blocking issue then I can go straight to `6.4.1`.

davidrapson:
Let's go straight to a 6.4.1, I think we can probably keep the alpha/beta releases for major versions 👍

PR #3604, opened 2025-02-10
Set engine version

davidrapson:
We currently fix the engine version to 0.1.0 which prevents us from using things like the deprecator to log version warnings and is at best just a little confusing.

Set the engine version to match the package version and add a spec to require them to match when updating the version.

### DIFF
--- a/contributing/04-release-process.md
+++ b/contributing/04-release-process.md
@@ -12,11 +12,17 @@ To perform a release you need:
 
 ## Prepare the release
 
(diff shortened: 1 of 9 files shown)


davidrapson:
> prevents us from using things like the deprecator to log version warnings

Did a bit more testing and realised it doesn't actually prevent us from using it—we've used it before—it just makes the deprecation horizon confusing. Still worth syncing these versions up though I think!

cnorthwood (approved):

PR #3606, opened 2025-02-10
Proposal: Remove TypeScript build step

davidrapson:
We've historically bundled our JavaScript using TypeScript. This presents a couple of trade-offs when developing the design system:

1. It requires an up-front build during development before the demo app and docs website can see changes
2. It prevents directly installing the design system package via git to test pre-release changes like we do with the Rails engine

In practice we make vanishingly little use of a type system for design system code as it consists of small amounts of DOM-scripting. We have no custom types, no meaningful business logic, and our components are configuration-free (or configuration is handled through data-attributes) so there's no typing benefit to consumers.

Removing TypeScript in this case greatly simplifies developing the design system with little meaningful impact to consumers of the project.

Includes an ADR with a more detailed version of this proposal.


lib/disclosure/index.d.ts
```diff
@@ -0,0 +1,3 @@
+import initDisclosure from './disclosure';
```
xswoodsx, line 1: Haven't seen these `.d.ts` files before. Here's a helpful [reddit comment](https://www.reddit.com/r/typescript/comments/hly19w/help_me_understand_dts_files/) on them for anyone interested

xswoodsx (approved):

lib/disclosure/index.d.ts
davidrapson: Oh good spot, these should actually go as they are the old generated typescript output

davidsauntson (approved):
I think this is fine and I agree with the points in the ADR 👍🏻

PR #3619, opened 2025-02-25
Update supported Ruby and Rails versions

davidrapson:
Both Ruby 3.1.x and Rails 7.0.x are due to reach end-of-life at the end of March. In preparation for that update our supported versions to match.

Any application using newer versions of the design system is running _at least_ Rails 7.2 and Ruby 3.2 anyway.


davidsauntson (approved):
Confirmed that it's ruby 3.1.X that is end of life at the end of March

davidrapson:
> Confirmed that it's ruby 3.1.X that is end of life at the end of March

Ooops I wrote 3.0, didn't I, you are correct!

PR #3630, opened 2025-03-05
Release/v6.4.2

davidsauntson:


davidrapson (approved):
Running the top-level `./setup` and the committing the result is probably the quickest way to get the lockfiles in sync with the latest gem version to get the build green 👍

PR #3632, opened 2025-03-07
Experimental:  Migrate rails form builder to design system engine

davidrapson:
This is an experiment to migrated the https://github.com/citizensadvice/rails-form-builder code into the main design system code base. Given we're now using this code in production at scale and the design system engine both contains the source components and is intended to house any related Rails code it makes sense to try merging these two together.

This is mostly a lift-and-shift of the existing code with the addition of some auto-loading simplifications thanks to embedding this code in an engine (the current library is a plain ruby gem).

Otherwise I've updated the demo app example form to use the form builder with relatively minimal changes.

Consider this a draft PR to get early review, to do this for real we'd want to write and agree an ADR for this as well as writing a companion guide to https://citizens-advice-design-system.netlify.app/guides/using-with-rails/ to cover using the form builder in your application.

### DIFF
--- a/demo/config/locales/en.yml
+++ b/demo/config/locales/en.yml
@@ -1,3 +1,4 @@
 ---
 en:
-  hello: Hello world
+  errors:
+    format: '%{message}'
--- a/demo/Gemfile.lock
+++ b/demo/Gemfile.lock
@@ -14,6 +14,10 @@ PATH
   specs:
     citizens_advice_components (6.4.2)
       actionpack (>= 7.1.0, < 9.0)
+      actionview (>= 7.1.0, < 9.0)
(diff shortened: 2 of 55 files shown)


demo/config/locales/en.yml
```diff
@@ -1,3 +1,4 @@
 ---
 en:
-  hello: Hello world
+  errors:
+    format: '%{message}'
```
davidrapson, line 4: @leeky Noticed that we've needed this here and in the existing apps using the form builder given the way we format error messages. Does that suggest we should be using `message` vs `full_message` in the library? Any objection to changing that here as we're moving this in?

davidrapson:
Closing in favour of https://github.com/citizensadvice/design-system/pull/3645

PR #3645, opened 2025-03-24
Migrate rails form builder to design system engine

davidrapson:
## Background

This is a clean version of https://github.com/citizensadvice/design-system/pull/3632 within an included ADR.

We've historically maintained a separate [rails-form-builder](https://github.com/citizensadvice/rails-form-builder) gem for using the design system components as a standard Rails form builder.

This was done to allow iterating on the rails form builder concept independently of design system releases.

Whilst this current set up was faster to iterate on the internal API changes to components themselves required multiple steps before any new updates could be implemented leading to drift as can be seen through the change to introduce a page_heading option to form fields (#3495) which was never implemented in the rails form builder.

As the interface for the library has now stabilised and we've reached the point where we are using the library in multiple public production systems it's a good time to bring the form builder into the main design system engine.

## Changes

Ports all the code from [rails-form-builder](https://github.com/citizensadvice/rails-form-builder) to the design system engine. Includes some updates to the demo application to use the form builder along with a basic Cypress test.

A companion ADR is also included.

In #3632 I suggested we write a new guide before merging this, but as the current library doesn't have an equivalent of this suggesting we defer this to allow us to merge this relatively soon. Plus it would be good for someone who has used the library in a recent project to try writing the guide.


engine/app/lib/citizens_advice_components/elements/base.rb
```diff
@@ -0,0 +1,123 @@
...
+          field_name_shim(object_name, attribute)
+        else
+          template.field_name(object_name, attribute)
+        end
+      end
+
+      def field_id
+        # We still support Rails 6.1 but rely on the use of field_id
+        # so we provide a shim for this method for older versions of Rails.
+        if Rails::VERSION::MAJOR < 7
```
leeky, line 58: Given the engine's gemspec requires Rails > 7.1, should we remove this? Same with `field_name`.

leeky (approved):
Thank you so much for doing this! Personally, it's nice to see the endless `require`s gone now it's part of the engine itself.

I swapped `content-platform-forms` over to this branch of the design system and removed reference to the form builder and it worked straight away, barring the change of the namespace in `default_form_builder`.

PR #3647, opened 2025-03-25
Release/v7.0.0

davidrapson:
Calling this a major version release as the Ruby and Rails version changes are technically breaking and whilst the new form builder can co-exist with the existing library and we won't archive that until all projects are moved across together they feel like enough of a trigger for a major version bump.


marianayovcheva (approved):
7️⃣

PR #3648, opened 2025-03-25
Support page_heading option when using the form builder

davidrapson:
Fills in the gaps from https://github.com/citizensadvice/design-system/pull/3495 now that the form builder is part of the design system to add full support for the page heading option.

Found structuring the tests for this a little fiddly so thoughts welcome on how to improve this.


engine/spec/features/text_input_spec.rb
```diff
@@ -20,6 +20,12 @@
     end
   end
 
+  it "renders a page heading" do
+    within "#page_heading_text_field" do
+      expect(page).to have_css("h1.cads-page-title", text: "First name")
+    end
+  end
```
davidrapson, line 27: I think I want to isolate these a little bit more and assert more e.g. a field with a page title doesn't support "optional" labels (as it doesn't make sense to have a single, but optional field. Also a little unclear how to debug when the test fails i.e. is it possible to _mount_ the test app (probably yes)

engine/app/lib/citizens_advice_components/elements/date_input.rb
```diff
@@ -21,7 +21,11 @@ def render
       private
 
       def legend_html
-        tag.legend(class: "cads-form-field__label") { safe_join([label, " ", optional_html]) }
+        if options[:page_heading].present?
+          tag.h1(class: "cads-page-title") { label }
```
johnny-m-young, line 25: I think this `h1` needs to be nested within a `legend` tag for it to be the same as the Design System (see [here](https://github.com/citizensadvice/design-system/blob/baa16a5af029e2f8c62bb2dc149e148ee8918ee2/engine/app/components/citizens_advice_components/date_input.html.erb#L5))

engine/spec/test_app/app/views/elements/text_inputs.html.erb
```diff
@@ -10,4 +10,8 @@
   <div id="text_field_with_additional_attributes">
     <%= f.cads_text_field :first_name, additional_attributes: { "data-example": "custom-attribute" } %>
   </div>
+
+  <div id="page_heading_text_field">
+    <%= f.cads_text_area :first_name, page_heading: true %>
```
johnny-m-young, line 15: This should be `f.cads_text_field`

johnny-m-young (commented):
I've just left a couple of small comments for now whilst I have a more in-depth think about the testing 🤔

davidrapson:
Updated this to cover the comments plus @johnny-m-young and I had a chat the other day and are planning to move the feature specs in the `test_app` folder over to the demo app to match how we test components. Essentially we'll end up with a reference implementation of a multi-step form to test the different permutations of the form builder via Cypress.

For now just updating enough of the tests to allow us to use this feature.

@johnny-m-young Mind testing this out with the referrals app? You should be able to replace `tag` with `branch` in the Gemfile temporarily to point to this change.

johnny-m-young:
@davidrapson [I've tested on referrals](https://github.com/citizensadvice/referrals/pull/1251/files) and this works for text areas. I might create some other input fields that we can remove later, just to test them

davidrapson:
@johnny-m-young The test you've done looks enough to me if you're happy to approve this?

johnny-m-young (approved):
Looks good, tested on the Referrals application and this works as expected 👍

PR #3649, opened 2025-03-27
Re-add rails-i18n

davidrapson:
Way back in https://github.com/citizensadvice/design-system/pull/2479 we removed rails-i18n as welsh language support had been removed. This was actually reinstated shortly after and seems to be well supported now so suggesting we re-add it as a dependency.


leeky (approved):
LGTM! :shipit:

PR #3650, opened 2025-03-27
Update deprecator horizon

davidrapson:
Increase version to 8.0 now that we've released 7.0.0. Also remove a Rails version check as the minimum Rails versions supported is 7.1.0.


davidrapson:
Going to merge this directly as it's just a housekeeping change.

PR #3652, opened 2025-03-28
[Experimental] Migrate common page helpers to design system

davidrapson:
An experiment to extract some common page helpers from our applications out into the design system

PR #3653, opened 2025-03-28
Add PageContent component

davidrapson:
## Background

Extracted from https://github.com/citizensadvice/design-system/pull/3652.  In many of our applications we've implemented some variant of a helper component which wraps the grid classes for a common page content template.

This is *such* a common pattern that we've implemented it ourselves in both the demo app and the docs.

## Changes

Adds a new `PageContent` component to the Rails engine. We've typically called this `PageLayout` in applications but going with `PageContent` as it better reflects the intent of this helper and fits with the fact the wrapping class is named `cads-page-content`.

Included usage documentation in both the grid foundations and "Using with Rails" guide. This is a Rails-specific convenience helper so it's not given a dedicated component page.


davidsauntson (approved):
👍🏻

davidrapson:
Just added an entry to the changelog to this before merging

PR #3654, opened 2025-03-28
Add page layout classes

davidrapson:
## Background

Builds on from https://github.com/citizensadvice/design-system/pull/3653. In many of our applications we've implemented some variation of page/site layout classes to handle the typical sticky footer layout for short pages. We even do this in the docs website for this reason.

## Changes

Suggesting we upstream these as a set of layout classes, named to match the other companion `cads-page-` prefixed classes we already have.

To correctly demonstrate this working in an application layout I've updated the demo app to work more like our production apps including a basic breadcrumb concern (arguably we could include this in the engine too, but that's for another PR).

The upshot of this is that the demo app home page get's a full layout too.

Screenshot 2025-03-28 at 15 39 49


davidsauntson (approved):
📄 🙌🏻 

Could you update the changelog pls?

PR #3666, opened 2025-04-07
Demo dependency update

davidrapson:
Batch update demo app dependencies, including updating the application to Rails 8.


davidrapson:
Merging this as it's routine maintenance and doesn't affect published packages.

PR #3669, opened 2025-04-08
Add new typography classes

davidrapson:
The method we currently have for using typography classes requires you to either use a cads-prose container or a custom extend. This changes introduces some companion utility classes for applying paragraph and heading styles.

There's an ideal scenario where someone can use the design system without writing any custom CSS and whilst we don't want to provide endless utility classes working with paragraphs and headings is a common enough scenario that it helps to have matching classes to go with the lower level styles.

I've also restructured the typography guide for clarity, particularly to better cover the expected use of cads-prose as this is a frequent point of confusion as well as adding a section on the cads-page-title.

### DIFF
--- a/design-system-docs/frontend/styles/_foundations.scss
+++ b/design-system-docs/frontend/styles/_foundations.scss
@@ -43,13 +43,6 @@
 
 // Typography
 
-// all foundation body copy is wrapped in a .cads-prose for the docs site
-.cads-prose {
-  .my-small-paragraph {
-    @extend %cads-paragraph-small;
-  }
-
-  .my-plain-list {
-    @extend %cads-list-no-bullet;
-  }
+.my-plain-list {
+  @extend %cads-list-no-bullet;
 }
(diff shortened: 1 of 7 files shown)


design-system-docs/frontend/styles/_foundations.scss
```diff
@@ -44,12 +44,6 @@
 // Typography
 
 // all foundation body copy is wrapped in a .cads-prose for the docs site
```
davidsauntson, line 46: I don't think we need this comment anymore

design-system-docs/src/_component_examples/_foundations/typography_small_paragraph.erb
```diff
@@ -3,5 +3,5 @@ title: Small paragraph
 iframe: false
 ---
 
-<p class="my-small-paragraph">Small paragraph font size is 16px on large screens and 14px on smaller screens.</p>
+<p class="cads-paragraphy-small">Small paragraph font size is 16px on large screens and 14px on smaller screens.</p>
```
davidsauntson, line 6: typo in `paragraphy`

davidsauntson (approved):
🗣️

design-system-docs/src/_component_examples/_foundations/typography_small_paragraph.erb
davidrapson: An exciting one! 😅

PR #3670, opened 2025-04-08
Rails engine font loading

davidrapson:
## Background

Historically we've required either adding a node_modules folder to the asset path or using a custom loader of some kind to handle font-loading within Rails.

We can in fact load assets as part of the Rails engine and they will automatically get included as part of the asset path. Use this behaviour to simplify the font-loading experience when using the Rails engine (which all but one of our applications is)

## Changes

- Bundle fonts as part of the Rails engine
- Include a view helper for pre-loading of fonts
- Include a rails.scss entrypoint for configuring the font path
- Include a rake task for syncing fonts between the package and engine if they change (rarely happens so acts more as documentation)
- Update "Using with Rails" guide to cover revised font loading
- Update the demo app to use propshaft to make sure we're testing with the new default asset pipeline


davidsauntson (approved):
Hurrah!

Can you update the changelog please (I think it'll be worth drawing attention to this change esp since users will need to add the additional `.scss` import).

davidrapson:
> Can you update the changelog please

Good spot! I'm going to get https://github.com/citizensadvice/design-system/pull/3668 in first too and then update this to remove pre-loading the cads font as it won't be needed by default.

PR #3677, opened 2025-04-22
Initial setup of wizard steps example in demo app

johnny-m-young:
## Changes
- Adds the `wizard_steps` gem to the demo app
- Creates a basic wizard steps form using the form builder.

### Start page
image

### Text area step
image


marianayovcheva:
This is great! 🎉  Is the long term plan to link to it from the main docs site?
> Should we namespace this?

I don't have strong opinion on this one - at the moment I don't think we have namespaces for the other pages in demo, but you mentioned that it might be useful for the models so I agree on that.

> Does the naming of the different parts make sense? I wasn't sure about the routes, and the controller name feels like it could be better.

Erm, I'm not the best at naming. On the controller, instead of `ExampleStepsController`, maybe something like `ExampleMultistepFormController` or `ExampleWizardController`? Essentially trying to read it out of context and `Steps` might be associated with different things, not just the form, so trying to reword that.

> What should the review page look like? At the moment I've taken a similar design to what is in the Referrals application.

That's a good question. In the witness referral form we have used this component: https://citizens-advice-design-system.netlify.app/components/summary-list/. But I noticed that here you have a `Change` button (which I love and might want to steal at some point) -  so not sure how this will work with the summary list component. Probably a design question. 

David and Chris have also worked on forms using the form builder so will leave them to also review - I'll also tag @leeky as we've worked on the initial forms together so might be worth for him to also have a look.

cnorthwood (approved):
this looks good to me as an example

PR #3707, opened 2025-06-10
Allow formatting lib directory

davidrapson:
Since we removed the build step from our javascript modules we now include the lib directory directly in our source. This directory was still being excluded from prettier formatting, this commit fixes that.


davidrapson:
Going to merge this in straight away if it passes CI as it's just a formatting fix.

PR #3708, opened 2025-06-10
Add character count component

### LINKED ISSUE #1691: Text Area Character Count Component
# Text Area Character Count Component 

Please provide a summary of the changes you would like to make to the existing component, along with any reasoning and supporting information.

To add a Character Count numbering to the TextArea (https://citizensadvice.github.io/design-system/?path=/story/forms-textarea--basic) to support:

• integration with other platforms (such as Casebook) when data transfer and receiving fields require limits added to map correctly across systems
• encourage users to provide sufficient information (on the principle that we would want to transmit the minimum amount of information)
• provide users understanding how many characters are allowed to be added to certain form fields using the TextArea component
• this would be a variant to the existing Text Area - so not all TextAreas would require a character count

Follow-up questions:
• what is the accessibility requirement for non-js and best practice guidance? 
• Do we follow the GDS model (https://design-system.service.gov.uk/components/character-count/)?

Depending on the above follow-up questions:
• designs will need to be generated to support development

davidrapson:
## Background

We currently have a basic character counter in use on the referrals application. It's general enough and we have an open issue (#1691) that suggests it's worth making this available through the design system itself.

## Changes

Adds the scaffolding for a character count component supported by textareas. Takes care to follow the same accessibility patterns as the GOV.UK equivalent which has recently undergone some major accessibility improvements[^1].

This version also differs from the current version in our products in that it doesn't use `maxlength` as the limit. This is deliberate to avoid data-loss when pasting text[^2].

As a result this version has the following additional states:

1. A no-js fallback state
2. A error state when exceeding the character limit
3. A screen-reader only version of the character count which updates at a slower rate.

## Screenshots

Empty state:

character-count-empty

Partial state:

character-count-partial

Error state:

character-count-error

Fallback (no-js) state:

character-count-fallback

## Caveats

The current implementation doesn't have any tests (neither rspec updates or Cypress tests), no documentation, nor is the text localised.

- [x] Update specs for engine code
- [x] Add Cypress tests for interaction behaviour
- [x] Implement welsh language capability for the character counter
- [ ] Provide translated text
- [x] Update documentation

Using this as a starting point to validate against the current implementation.

[^1]: https://dav-idc.com/making-a-character-count-component-more-accessible/
[^2]: https://adamsilver.io/blog/dont-use-the-maxlength-attribute-to-stop-users-from-exceeding-the-limit/

### DIFF
--- a/engine/app/components/citizens_advice_components/textarea.rb
+++ b/engine/app/components/citizens_advice_components/textarea.rb
@@ -33,7 +33,19 @@ module CitizensAdviceComponents
     end
 
     def base_input_attributes
-      super.merge(rows: @rows, value: nil, "data-character-count": character_count.presence)
+      if character_count.present?
+        super.merge(
+          rows: @rows,
+          value: nil,
+          "data-character-count": character_count.presence,
+          "data-character-count-over-limit": t("citizens_advice_components.character_count.over_limit"),
+          "data-character-count-remaining-zero": t("citizens_advice_components.character_count.remaining_zero"),
+          "data-character-count-remaining-one": t("citizens_advice_components.character_count.remaining_one"),
+          "data-character-count-remaining-other": t("citizens_advice_components.character_count.remaining_other")
+        )
+      else
+        super.merge(rows: @rows, value: nil)
+      end
     end
   end
 end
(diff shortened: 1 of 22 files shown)


engine/app/components/citizens_advice_components/textarea.rb
```diff
@@ -2,28 +2,38 @@
 
 module CitizensAdviceComponents
   class Textarea < Input
```
davidrapson, line 4: Noting that having this subclass from `Input` is getting a bit unwieldy. We should probably separate these out.

PR #3709, opened 2025-06-10
Fix erb_lint deprecations

davidrapson:
Fixes:

> Calling `erblint` is deprecated, please call the renamed executable `erb_lint` instead.
>
> The config file has been renamed to `.erb_lint.yml` and `.erb-lint.yml` is deprecated. Please rename your config file to `.erb_lint.yml`.


davidrapson:
Also going to merge this straight in if it passes

PR #3719, opened 2025-06-17
Migrate from jest to vitest

davidrapson:
In lieu of upgrading to Jest 30. Includes an ADR for the proposed change.

### DIFF
--- a/scss/__tests__/generics.test.js
+++ b/scss/__tests__/generics.test.js
@@ -3,7 +3,11 @@ const path = require('path');
 
 test('default font-path', () => {
   // Compile against top-level entrypoint as we're testing settings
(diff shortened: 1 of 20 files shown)


scss/__tests__/generics.test.js
```diff
@@ -3,7 +3,11 @@ const path = require('path');
 
 test('default font-path', () => {
   // Compile against top-level entrypoint as we're testing settings
-  const output = sass.compile('scss/lib.scss').css.toString();
+  const output = sass
+    .compile('scss/lib.scss', {
+      logger: sass.Logger.silent,
+    })
```
davidrapson, line 9: The `logger` change is just to better manage the test output for these sass checks

PR #3730, opened 2025-06-30
bugfix: when header links are collapsed into the nav menu, always show the nav menu when the header links are hidden

cnorthwood:
this fixes a bug (NAT-1407) on the Intranet, where there are not that many header links so it does not collapse down to the dropdown menu on mobile, but because the header links are hidden and only behind the More dropdown, it becomes impossible to then access those header links. This changes the behaviour so that the More button is always shown if the header links exist but are not visible.


davidrapson (approved):
👍 Good catch. One thought for now, one for later:

1. Could we add something to the Cypress tests to check this behaviour?
2. For later: I wonder if there's some value in merging `header.js` and `navigation.js`. One doesn't _really_ work without the other.

PR #3736, opened 2025-07-07
Remove icon font completely

davidrapson:
Removes all references to the legacy icon font, projects still relying on it for their own components will be required to self-host. Targeting v9.0.0 with this so doesn't need to go in straight away.


davidrapson:
Closing this due to conflicts and as we should probably make sure all consumers have migrated away before ditching this completely (rather than foisting work on teams prematurely). Captured as an issue https://github.com/citizensadvice/design-system/issues/3810

PR #3738, opened 2025-07-15
Add page_heading to radio and checkbox collections

davidrapson:
Support `page_heading` option for radio and checkbox collections

This option was added to text inputs, date inputs, and text areas in https://github.com/citizensadvice/design-system/pull/3495 and then added to the form builder in https://github.com/citizensadvice/design-system/pull/3648.

This change is to add the same option to radio and checkbox collections. However we need to do some houskeeping in the process as we added a version of this way back in https://github.com/citizensadvice/design-system/pull/2071 under the name `legend_heading` with slightly inconsistent behaviour, so provides a deprecation warning to support the old option alongside the new one (albeit with revised behaviour).


engine/app/components/citizens_advice_components/checkbox_single.erb
```diff
@@ -1,17 +1,19 @@
...
-        <%= error_message %>
-      </p>
-    <% end %>
-    <div class="cads-form-group__item">
-      <%= tag.input(class:"cads-form-group__input", **checkbox.attributes(name, id, error: error?)) %>
-      <label class="cads-form-group__label"
-             for="<%= label_id %>">
-        <%= checkbox.label %>
-      </label>
+  <div class="cads-form-field__content">
```
davidrapson, line 3: This change is just lifting up the `cads-form-field__content` element for consistency with other templates.

engine/app/components/citizens_advice_components/form_group.html.erb
```diff
@@ -1,29 +1,31 @@
...
+  <div class="cads-form-field__content">
+    <%= tag.fieldset(class: fieldset_classes) do %>
+      <legend class="cads-form-field__label">
+        <% if page_heading? %>
+          <h1 class="cads-page-title"><%= legend %></h1>
+        <% else %>
+          <%= legend %>
+          <% if optional? %>
+            <span class="cads-form-field__optional"><%= "(#{t('citizens_advice_components.input.optional')})" %></span>
+          <% end %>
```
davidrapson, line 12: For consistency with other fields a field using `page_heading` can't use the optional marker as they are contradictory goals.

engine/app/components/citizens_advice_components/radio_group.rb
```diff
@@ -4,13 +4,9 @@ module CitizensAdviceComponents
   class RadioGroup < FormGroup
     renders_many :inputs, Checkable::Radio
 
-    def initialize(legend:, name:, id: nil, options: nil)
-      super
-      set_options(options)
-    end
```
davidrapson, line 10: This isn't needed

engine/app/lib/citizens_advice_components/elements/collections/check_boxes.rb
```diff
@@ -8,8 +8,10 @@ def render
           tag.div(class: form_field_classes) do
             safe_join([
               error_marker,
-              tag.fieldset(class: "cads-form-field__content cads-form-group cads-form-group--checkbox") do
-                safe_join([legend_html, hint_html, error_message_html, checkboxes_html])
+              tag.div(class: "cads-form-field__content") do
```
davidrapson, line 11: Mirrors the change in the component

scss/6-components/forms/fields.scss
```diff
@@ -29,14 +29,6 @@
   margin-bottom: $cads-spacing-3;
 }
 
-.cads-form-field__label--with-heading {
-  margin-bottom: 0;
-}
-
-.cads-form-field__heading {
-  @extend %cads-h1;
-}
```
davidrapson, line 38: Handled via the `cads-page-title` class directly

engine/app/components/citizens_advice_components/radio_group.rb
marianayovcheva: For my personal understanding - is it because it's inheriting from `FormGroup`?

PR #3740, opened 2025-07-16
Simplify textarea interface

davidrapson:
Like with #3739 the textarea is another component where inheritance causes an issue. In this case the component both _inherits_ from `Input` but also _calls_ `Input` for it's template.

As a textarea ultimately doesn't behave the same way as an input (no type, value handled differently, can have rows etc.) it's simpler for us to inline the parts we need again.

It's possible we might be able to extract a `FormField` component which handles just some shared markup, but cautious about introducing _new_ interfaces here. Particularly as in reality the main way we use form fields now is through the new form builder.

### DIFF
--- a/engine/spec/components/citizens_advice_components/textarea_spec.rb
+++ b/engine/spec/components/citizens_advice_components/textarea_spec.rb
@@ -38,7 +38,6 @@ RSpec.describe CitizensAdviceComponents::Textarea, type: :component do
         name: "example-input[test]",
         id: "test-id",
         label: "Example input",
-        type: :text,
         options: { hint: "This is the hint text",
                    error_message: "Enter your name" }
       )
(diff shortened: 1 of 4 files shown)


engine/spec/components/citizens_advice_components/textarea_spec.rb
```diff
@@ -154,20 +153,6 @@
     it { is_expected.to have_css "textarea[aria-describedby='example-textarea-error example-textarea-hint']" }
   end
 
-  context "when a type is specified" do
```
davidrapson, line 157: Type is just no longer an argument

leeky (approved):
LGTM! :shipit: 

I very much agree that `<textarea>` should be treated as entirely separate things to text inputs, so this makes a lot of sense.

PR #3741, opened 2025-07-16
Simplify select interface

davidrapson:
Another component where inheritance complicates the interface.

Like with a textarea a select has subtle differences from an input (no type, value handled differently).

Includes a public `render_select_options` method to make it easier for consumers to sub-in things like `grouped_options_for_select`. We could probably support that as a first-class feature but would need more updates to the form builder for that to work properly so just introducing a seam to do it for now.

### DIFF
--- a/engine/app/components/citizens_advice_components/select.html.erb
+++ b/engine/app/components/citizens_advice_components/select.html.erb
@@ -1,0 +1,23 @@
+<div class="cads-form-field<% if error? %> cads-form-field--has-error<% end %>">
+  <% if error? %><div class="cads-form-field__error-marker"></div><% end %>
+  <div class="cads-form-field__content">
+    <label class="cads-form-field__label" id="<%= label_id %>" for="<%= input_id %>">
+      <%= label %>
+      <% if optional? %>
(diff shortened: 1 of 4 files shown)


engine/app/components/citizens_advice_components/select.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="cads-form-field<% if error? %> cads-form-field--has-error<% end %>">
+  <% if error? %><div class="cads-form-field__error-marker"></div><% end %>
+  <div class="cads-form-field__content">
+    <label class="cads-form-field__label" id="<%= label_id %>" for="<%= input_id %>">
+      <%= label %>
+      <% if optional? %>
+        <span class="cads-form-field__optional"><%= "(#{t('citizens_advice_components.input.optional')})" %></span>
+      <% end %>
+    </label>
```
davidrapson, line 9: This deliberately doesn't support `page_heading` as we don't have a use-case for selects to support that.  Previously it _did_ support this but by accident as the property was shared by all `Input` sub-classes.

leeky (approved):
LGTM! :shipit:

PR #3760, opened 2025-08-01
Update to ViewComponent 4

davidrapson:
ViewComponent 4 is now out (https://github.com/ViewComponent/view_component/releases/tag/v4.0.0). This change brings the appraisals and demo app up to date with this version. We'd already added support to the engine by testing against the release candidates so this just brings things up to date internally.

Unfortunately—but not unsurprisingly—the bridgetown shim breaks again with v4.0.0. Keeping this pinned to 3.x until we can find another fix (or potentially an alternative approach).


design-system-docs/Gemfile
```diff
@@ -13,9 +13,9 @@ gem "puma"
 
 gem "citizens_advice_components", path: "../engine"
 
-# citizens_advice_components bundles view component but we
-# also use this to write docs components so explicitly
-# name it as a dependency.
+# ViewComponent is bundled by the design system but include
+# it explicitly as we use it for example components in the
+# demo app, and so we can control the version if necessary.
 gem "view_component", "~> 3.23"
```
leeky, line 19: Should this use V4 of `view_component` too?

davidrapson: Sadly not yet:

> Unfortunately—but not unsurprisingly—the bridgetown shim breaks again with v4.0.0. Keeping this pinned to 3.x until we can find another fix (or potentially an alternative approach).

PR #3763, opened 2025-08-04
Rails-based static site generation for docs website

davidrapson:
## Background

Builds on a recent prototype to use https://github.com/benpickles/parklife along with https://github.com/benpickles/decant to build a static site using Rails. Parklife does the static site generation (works with any Rack app) and Decant handles directories of static content and front matter. Attempts to bypass the trade-offs we get integrating Rails-components in a non-Rails environment with the current Bridgetown docs.

The prototype was successful enough (everything ported over) that I thought it worth moving it straight into a PR along with and ADR. Set up instructuctions are included in `website/README.md` as well as the global setup script being updated.

## How to review

The structure of the new `website` directory should be familiar as a standard Rails application. The notable code to review is under `app/models` which defines the `Decant` definitions to find the content. Otherwise the main place to look is in `content` which contains all the docs and example files for the website. This follows the same structure as the previous website just cleaned up a bit as a result of us being able to control the structure a bit more.

Otherwise a pretty straight port of the previous code.


.github/workflows/docs-workflow.yml
```diff
@@ -32,61 +32,23 @@ jobs:
...
-          build-dir: ./design-system-docs/output
+          build-dir: ./website/build
           draft: true
           comment-on-pull-request: true
           config-path: ./netlify.toml
-
-  # This job will:
-  #   * deploy a production build every time there is a release created
-  #   * comment that commit with the deploy URL
-  publishOnRelease:
```
cnorthwood, line 59: I'm unsure why this has been removed, was this dead code (as in, docs-workflow-main achieves the same end?)

cnorthwood (approved):
this is way too big for me to have meaningfully reviewed all of it, but the general direction of travel and intent here is good and seems sounds and I understand what it's doing so I'm giving this a +1, but I think it might be best for someone else to look after this as well given the size of the PR as I've not been able to go over the code comprehensively.

.github/workflows/docs-workflow.yml
davidrapson: Oops, mean to drop a comment on it that this is redundant as it gets overridden every time we merge to main.

marianayovcheva (approved):

davidrapson:
Merging this. @marianayovcheva and I went through the changes on a call. Thanks for spotting the broken links on the home page and the navigation example!

PR #3784, opened 2025-08-20
Remove serve as a development dependency

davidrapson:
Included as a convenience for the new website code but is a bit overkill to actually include as a dependency so remove it.


davidrapson:
Going to merge this straight in as it passes as it avoids a dependenbot warning for a dependency that can be installed on-demand if needed https://github.com/citizensadvice/design-system/security/dependabot/485

PR #3785, opened 2025-08-20
Clean up dependabot rules

davidrapson:
- Remove ignores that have been fixed (e.g. running eslint v9)
- Adjust frequencies to all be weekly for consistency. Should be easier to manage now that the docs website has more consistent rails dependencies.


akamike (approved):
Looks good, sensible interval change 👍

PR #3791, opened 2025-08-27
CP-606: add slot in footer to allow additional logo

cnorthwood:
This is needed because we want to include the Fundraising Regulator logo on the public website.


engine/app/components/citizens_advice_components/footer.html.erb
```diff
@@ -55,6 +55,11 @@
         <p class="cads-footer__meta-text" data-testid="copyright"><%= t("citizens_advice_components.footer.copyright", year: current_year) %></p>
         <p class="cads-footer__meta-text" data-testid="legal-summary"><%= legal_summary_text %></p>
       </div>
+      <% if additional_logo.present? %>
+        <div class="cads-footer__additional-logo">
+          <img src="<%= additional_logo.src %>" alt="<%= additional_logo.alt %>">
```
davidrapson, line 60: I'm a little wary of the `img` tag here. For the similar custom logo behaviour in the header we allow passing a plain slot instead (we typically render an inline SVG but it gives the consumer the choice then).

davidrapson (commented):
The general idea makes sense. How do you feel about making this a plain slot rather than _requiring_ the image tag? Would match better with how we handle this in the header.

davidrapson (approved):
Sounds good then 🙂. Hitting approve so you can merge once you've made that change to the interface. Also glad to see the process for updating the website pages still feels the same.

PR #3793, opened 2025-08-27
Form builder docs

davidrapson:
Now that it's possible to use the form builder in the website code without workarounds we can start updating our form components to reference the form builder.

Starts with the button component, showing how to use the form builder helpers for next and previous buttons.


davidrapson:
Merging this to fix the broken image

website/app/controllers/application_controller.rb
```diff
@@ -3,6 +3,8 @@
 class ApplicationController < ActionController::Base
   include CitizensAdviceComponents::Helpers
 
+  default_form_builder CitizensAdviceComponents::FormBuilder
```
leeky, line 6: Does the documentation need to mention that `cads_button` is only available when the `CitizensAdviceComponents::FormBuilder` is used (I don't believe it's enabled by default in the design system gem)..?

leeky (approved):
LGTM! :shipit:

PR #3796, opened 2025-08-28
Refactor actions workflows

davidrapson:
Split the main and pull request workflows out and compose the test steps using a shared workflow. Allows some simplification of docs builds, mostly just to clarify the role of each workflow.

Removed the nightly cron run as in practice it doesn't help us catch anything we wouldn't otherwise.

**Steps before merging:** 

- [x] Update the required checks to reflect the new workflow name


akamike (approved):
Looks good, I think the conflicts are just due to the prettier changes so should be easily resolved.

PR #3811, opened 2025-09-17
Remove cads-sr-only-focusable class

davidrapson:
Has an old deprecation comment and a search confirms that we've never used this in any internal or product code.


leeky (approved):
LGTM! :shipit:

PR #3813, opened 2025-09-19
Upgrade to Cypress 15

davidrapson:
The last remaining plugin was updated to support Cypress 15 so we can upgrade now.


davidrapson:
The other plugins were updated as part of https://github.com/citizensadvice/design-system/pull/3832 so going to merge this in as it's just a straight dependency upgrade now.

PR #3814, opened 2025-09-19
Fix/badge text styles

marianayovcheva:
After scss update to v1.92.0, there is a breaking change, meaning that the includes are being overridden by the component styles. We are overriding the font styles of the badge component anyway, so we can just extract them inline.

If the component is used in a host app which has scss v1.92.0 or greater, this results in a change in specificity.
Before the update:
image
After the update:
image

Also the build css code shows the difference:
Before the update:
```
.cads-badge{font-size:1rem;line-height:1.625rem;line-height:.875rem;font-weight:700;display:inline-block;margin-bottom:.75rem;color:#4a4e4f}@media(max-width: 767.98px){.cads-badge{font-size:.875rem;line-height:1.5rem}}
```
After the update (note the second `.cads-badge` declaration at the end):
```
cads-badge{font-size:1rem;line-height:1.625rem}@media(max-width: 767.98px){.cads-badge{font-size:.875rem;line-height:1.5rem}}.cads-badge{line-height:.875rem;font-weight:700;display:inline-block;margin-bottom:.75rem;color:#4a4e4f}
```
It's like before the compiled code took all the import + badge code first and followed it with the media query. The code after the update does the import first, followed by import's media query and then the badge code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Inline text styles for the badge component, instead of using `@include cads-typographic-scale-text-small;`. This resolves a breaking change introduced by host apps using scss >= v1.92 
 - Remove `.cads-sr-only-focusable` class. This has never been used in any internal design system or product code.
 
 ## v8.3.0


marianayovcheva:
I've also checked and we don't have other instances of a component which uses `cads-typographic-scale-text-small` but overrides its text styles.

PR #3835, opened 2025-10-06
Proposal: Deprecate form components in favour of form builder

davidrapson:
Proposal to deprecate view component form interfaces in favour of form builder methods.

PR #3836, opened 2025-10-06
Increase test coverage for form builder methods

davidrapson:
Updates how we test form builder methods to allow using capybara matchers and expands tests to cover all form builder methods with existing specs the equivalent view component interfaces.

Supports https://github.com/citizensadvice/design-system/pull/3835 but can be merged independently as this is solely about increasing test coverage.

These tests surfaced a few bugs or missing options. ~Some of these have been fixed with the remainder marked as pending to be fixed in subsequent commits.~ **Edit**: Ended up extracting the fixes and replaced everything with `pending` notices, allows issues to be fixed separately from increasing test coverage.


engine/spec/helpers/citizens_advice_components/form_builder_cads_button_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        render_inline builder.cads_button
+      end
+
+      it "renders a default button" do
+        expect(page).to have_button text: "Save changes"
+      end
+    end
+
+    describe "variant" do
+      let(:builder_method) { builder.cads_button(variant: variant) }
```
cnorthwood, line 23: I'm not entirely sure what "builder_method" means here. I understand what the code does, but the name isn't obvious to me

davidrapson: Good shout. This is meant to mean the specific form builder method we're testing (hence `builder_method`) but I could make this more descriptive 👍 . Maybe something like `form_builder_field`

engine/app/lib/citizens_advice_components/elements/collections/base.rb
```diff
@@ -37,13 +37,23 @@ def hint_html
...
+          tag.p(class: "cads-form-field__error-message", id: error_id) { error_message }
         end
 
         def optional_html
           return unless optional
 
-          tag.span(class: "cads-form-field__optional") { "(optional)" }
+          tag.span(class: "cads-form-field__optional") do
+            "(#{I18n.t('citizens_advice_components.input.optional')})"
+          end
```
davidrapson, line 48: Might be worth me pulling the actual fixes out into separate PRs so this change is just test coverage. Wanted to give an indication of the kinds of issues the tests spotted.

engine/spec/helpers/citizens_advice_components/form_builder_cads_button_spec.rb
davidrapson: Went with `field` for the different field types or more specific terms where it made sense i.e. `button` and `error_summary`. Also did a pass of all the tests so we're consistently using this pattern which cleaned up the various context/describe/before blocks. Thanks for the nudge on the test design here!

engine/spec/helpers/citizens_advice_components/form_builder_cads_collection_check_boxes_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+RSpec.describe CitizensAdviceComponents::FormBuilder do
+  include FormBuilderTestHelpers
+
+  subject { page }
+
+  let(:model) { ExampleForm.valid_example }
+  let(:builder) { form_builder_for(model) }
+  let(:example_options) do
+    [
+      FormOption.new(id: "GBP", name: "GBP"),
```
leeky, line 12: Is it maybe worth having different `id` and `name` values here?

PR #3892, opened 2025-11-19
Experimental: Add character count to textareas

davidrapson:
Fresh version of https://github.com/citizensadvice/design-system/pull/3708

## Background

We currently have a basic character counter in use on the referrals application. It's general enough and we have an open issue (#1691) that suggests it's worth making this available through the design system itself.

## Changes

Adds the scaffolding for a character count component supported by textareas. Takes care to follow the same accessibility patterns as the GOV.UK equivalent which has recently undergone some major accessibility improvements[^1].

This version also differs from the current version in our products in that it doesn't use `maxlength` as the limit. This is deliberate to avoid data-loss when pasting text[^2].

As a result this version has the following additional states:

1. A no-js fallback state
2. A error state when exceeding the character limit
3. A screen-reader only version of the character count which updates at a slower rate.

[^1]: https://dav-idc.com/making-a-character-count-component-more-accessible/
[^2]: https://adamsilver.io/blog/dont-use-the-maxlength-attribute-to-stop-users-from-exceeding-the-limit/

### DIFF
--- a/demo/app/assets/javascript/application.js
+++ b/demo/app/assets/javascript/application.js
@@ -5,6 +5,7 @@ import {
   initOnThisPage,
   initDisclosure,
   initErrorSummary,
+  initCharacterCount
 } from '@citizensadvice/design-system';
 
 initHeader();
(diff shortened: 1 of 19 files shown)


davidrapson:
Closing until we're ready to pick this up again

PR #3913, opened 2025-12-03
Configure dependabot cooldown

davidrapson:
Mirrors configuration on other repositories using a 7 day cooldown for all ecosystems.


KPobeeNorris (approved):
✨

PR #4100, opened 2026-05-07
Add Configuring Dependabot guide

davidrapson:
We've been using multi-ecosystem groups successfully on a few projects to handle grouping design system updates automatically when there's a new release. Document the pattern for reference.

image


marianayovcheva (approved):
Nice! 🤖 🎉

PR #4109, opened 2026-05-14
chore: update design system docs

marianayovcheva:
This takes into account the updates in v9.1.0 named `additional_attributes` was deprecated. Other named parameters like `collection`, `text_method` and `value_method` have also been deprecated.

Having read https://github.com/citizensadvice/design-system/blob/main/contributing/adr/2025-11-28-deprecate-form-components.md I was also contemplating removing the "View component version" section of the docs for the form elements or adding a bit of text to discourage users from going for that option. However, I've decided not to do it as part of this PR so that this decision is kept separate. Also, it might be too early to do it and worth waiting until we officially deprecate the view components version of the form fields (or move them to Legacy as per the suggestion in https://github.com/citizensadvice/design-system/issues/4049). In any case, opinions/ideas welcome!

### DIFF
--- a/website/content/components/checkbox-group.md
+++ b/website/content/components/checkbox-group.md
@@ -44,7 +44,7 @@ The components we provide use wrap checkbox groups in a fieldset. In this contex
 When using with Rails we recommend using the form builder method provided by `CitizensAdviceComponents::FormBuilder`.
 
 ```rb
-cads_collection_checkboxes(attribute, collection:, value_method:, text_method:, options = {})
+cads_collection_checkboxes(attribute, collection, value_method, text_method, options = {}, html_options = {})
 ```
 
 The method works similarly to the default [`collection_checkboxes` helper](https://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-collection_checkboxes).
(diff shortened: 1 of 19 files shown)


website/content/components/checkbox-group.md
```diff
@@ -74,16 +74,14 @@ But this can also work with any collection:
...
-    value_method: :name,
+    Locations.all,
+    :id,
+    :name,
     hint: "Example hint text"
   ) %>
 <%% end %>
 ```
 
-The `:value_method` and `:text_method` parameters are methods to be called on each member of `collection`. The return values are used as the value attribute and contents of each checkbox.
```
marianayovcheva, line 85: I've removed this but happy to add some text if it's still useful to have a line about the id and the name.

davidrapson (approved):
🙏 Thank you for picking this up. I'm keen to drop the view component parts of the docs for forms sooner rather than later but agree doesn't need to be in this PR

website/content/guides/rails-form-builder.md
```diff
@@ -204,14 +204,14 @@ From here you can use the available form helpers to render each field:
...
+  [
     ["Choose an option", ""],
     ["GBP", "GBP"],
     ["EUR", "EUR"],
     ["USD", "USD"]
   ],
-  text_method: :first,
-  value_method: :last,
+  :id,
+  :name,
```
davidrapson: Oh just noticed these should be `:last, :first` as the collection is still an array in the examples. Just in case people copy and paste this as part of running through the guide

marianayovcheva: That's a good catch, thanks! I've now rebased and updated the examples where I've replaced `:first` and `:last`

PR #4133, opened 2026-06-05
Move form components to CitizensAdviceComponents::Legacy namespace

davidrapson:
Follows through with the [ADR to Deprecate form components in favour of form builder](https://github.com/citizensadvice/design-system/blob/main/contributing/adr/2025-11-28-deprecate-form-components.md)  by moving all form view components to a `CitizensAdviceComponents::Legacy` namespace.

This allows applications to continue to use these components for a while longer with explicit opt-in but without large amounts of deprecator logging.


website/content/examples/text_input/full_name.html.erb
```diff
@@ -3,7 +3,7 @@ title: full_name
 form: true
 ---
 
-<%= render CitizensAdviceComponents::TextInput.new(
+<%= render CitizensAdviceComponents::Legacy::TextInput.new(
```
marianayovcheva, line 6: (Not blocking) Annoying question about this and the other two examples in the names pattern - if we discourage users to use the view component, should the examples in the pattern be updated to use the form builder instead?

marianayovcheva (approved):
LGTM! ✨

PR #4135, opened 2026-06-05
Remove support for Rails 7.2

davidrapson:
Rails 7.2 is nearing EOL, as we are prepping for v10 this is a good time to deprecate support for this version. All Rails application using the components should already be running at least Rails 8.x


seymourski-github (approved):
👍

marianayovcheva (approved):

PR #4157, opened 2026-06-24
Grouped collection select refactored

mpapp88:
This PR implements a new grouped collection select helper (`cads_grouped_collection_select`) within the core `FormBuilder` engine. Also this PR is focusing only the engine implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+### 24 June 2026
+
+**New**
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## Unreleased
+
+### 24 June 2026
```
davidrapson: The release script will add this for you so we can remove this part, just starting with the `**New**` line

davidrapson (approved):
Lovely work, small note on the changelog entry but happy for this to go in once we've confirmed it works in one of our existing apps.

PR #4171, opened 2026-07-08
Remove deprecations for v10 release

davidrapson:
- Remove deprecated collection params
- Remove additional_attributes named parameter from field methods
- Remove deprecator warnings from legacy components

### DIFF
--- a/engine/app/components/citizens_advice_components/legacy/form_group.rb
+++ b/engine/app/components/citizens_advice_components/legacy/form_group.rb
@@ -21,8 +21,6 @@ module CitizensAdviceComponents
         @optional = fetch_or_fallback_boolean(options[:optional], fallback: false)
         @legend_heading = fetch_or_fallback_boolean(options[:legend_heading], fallback: false)
         @page_heading = fetch_or_fallback_boolean(options[:page_heading], fallback: false)
-
-        legend_heading_deprecation
       end
 
       def extra_fieldset_classes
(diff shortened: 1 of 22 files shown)


engine/app/lib/citizens_advice_components/elements/traits/field_group.rb
```diff
@@ -74,7 +74,9 @@ def label_html
         end
 
         def options_for_html
-          options.without(:label, :hint, :required, :page_heading, :additional_attributes)
+          # The default behaviour of fields is to pass options on to the HTML element
+          # Extract any custom options that we don't want passing on and add expected defaults.
+          options.without(:label, :hint, :required, :page_heading)
```
marianayovcheva, line 79: Question - would we still be able to pass attributes like `autocomplete` or `inputmode`. We use them in the forms repo.

davidrapson: Yes anything in `options` gets passed through, with the exception of the named properties listed in `without`

marianayovcheva: Oh, I see. I didn't read it properly the first time

engine/app/components/citizens_advice_components/legacy/form_group.rb
```diff
@@ -21,8 +21,6 @@ def set_options(options)
         @optional = fetch_or_fallback_boolean(options[:optional], fallback: false)
         @legend_heading = fetch_or_fallback_boolean(options[:legend_heading], fallback: false)
         @page_heading = fetch_or_fallback_boolean(options[:page_heading], fallback: false)
-
-        legend_heading_deprecation
```
marianayovcheva, line 25: Question to help my understanding - I assume this can be removed because they are now in the legacy namespace anyways?

marianayovcheva (approved):
LGTM!

PR #4205, opened 2026-08-03
Migrate from ESLint to Oxlint

davidrapson:
Reduces the number of dependencies for JavaScript linting from 5 to 1 with drop-in compatibility. Includes an ADR with further context.


davidrapson:
Testing out the stacks preview with this out of curiosity but can split again if that's confusing.

cnorthwood (approved):
I have a feeling I'm a bit out of the loop with the JS ecosystem maybe as I've not come across Oxlint before and haven't really found Eslint a rough edge in using it. I'm generally a bit wary of the shiny thing syndrome in JS dev sometimes so swapping out something mature for new I approach with a bit of trepidation. That said, it looks like it's mostly a drop in replacement, and the PR looks technically sound, and I trust your judgement on reading the wider community if this is the current lowest friction toolset

davidrapson:
100% agree in general, in this case we've found it make a huge difference in reducing churn around eslint-dependency versions and is a drop-in replacement and is based on us running it for a good while in other products.

PR #4208, opened 2026-08-03
Test grouping dependabot updates by type

davidrapson:
Based on recommendations in https://github.blog/security/supply-chain-security/tame-dependabot-group-your-updates-slow-the-cadence-keep-security-fast/ tests using the new `group-by: depedency-name` along with multi-directory groups to see if we can reduce the cascade of updates where there are duplicates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,6 @@
 version: 2
 updates:
-  ##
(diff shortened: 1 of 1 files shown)


davidrapson:
Going to merge this to give it a go

PR #4231, opened 2026-09-01
Consolidate testing tools

### LINKED ISSUE #4220: Explore replacements to BackstopJS
We need to look for a replacement to BackstopJS for maintenance reasons. We currently run a combination of Cypress and BackstopJS together for browser and visual regression testing.

There now major testing libraries that support visual regression testing, including Playwright directly: https://playwright.dev/docs/test-snapshots, and Vitest (also via playwright) https://vitest.dev/guide/browser/visual-regression-testing.html

Whilst Cypress has guides for visual testing too they defer to either additional low level libraries or commercial services https://docs.cypress.io/app/tooling/visual-testing

This task is to explore a replacement. There is work going on within internal teams around this same problem too so we should converge on the same approach.

davidrapson:
## Context

We currently run a mix of different testing tools for component code:

- BackstopJS for visual regression testing
- Cypress for some behavioural tests and end to end testing of forms
- Cypress for basic accessibility smoke testing using aXe
- Vite for minimal component JavaScript testing using Testing Library
- Vite for node-base SCSS library tests

The primary motivation to revisit this test design is that BackstopJS is no longer maintained and depends on an outdated Puppeteer version with known security vulnerabilities.

## Changes

Proposes the following simplifications:

- Consolidate both Cypress and BackstopJS with Playwright, which is actively maintained and provides built-in visual comparison support.
- Include accessibility checks inline within each components Playwright scenarios. The previous Cypress tests had a generic aXe scenario which loaded up each component in turn. This had the benefit of applying when any new component was added but did not allow us to re-check on different states. Forgoes the automatic test in favour of more comprehensive state tests as a trade-off.
- Remove minimal `lib` tests using Vite, Testing Library, and JSDOM, in favour of a consolidated Playwright test suite. There's nothing being tested in here that isn't better tested via a full browser environment
- Rewrite the remaining SCSS library tests using plain `node:test` as we no longer require the features of Vitest. The inbuilt test runner is stable since Node 20 and even has snapshotting built in since 22.

A fuller version of this along with consequences included as an ADR.

## Notes for reviewers

Used the `run-server` command for playwright which I've found to be the best way to run the browsers themselves in Docker for rendering consistency whilst still being able to run `npx prettier` commands directly.

The `just check-all` command should pass locally, but I recommend trying this out using a local dev server to compare against e.g. Cypress UI mode. Details in the testing guide, but for brevity:

1. Start the playwright server with `bin/start-playwright-server`
2. Separately run a local Rails server `bin/dev`
3. You can now run individual playwright commands e.g. `npx playwright --ui playwright/sample-form.spec.js`

**Note:** Status checks will need updating before merging as names have changed


demo/bin/playwright-ci
```diff
@@ -0,0 +1,23 @@
...
+    until nc -z localhost 5000; do
+      echo "Waiting for server on port 5000..."
+      sleep 2
+    done
+  '
+
+  # Clean up after ourselves
+  trap "docker stop playwright-server" EXIT INT TERM
+
+  npx playwright test $@
```
davidrapson, line 22: This is a bit fiddly but essentially trying to wrangle starting the docker component along with waiting for ports to be ready. The `run-server` command relies on websockets wo we use `nc` to wait for the port only rather than something like curl.

For local running it's nicer to run the components separately (noted in the testing guide) but this is enough so that CI and the `just check-all` command can pass without additional set up.

demo/playwright/button.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+test("Button (with right icon)", async ({ page }) => {
+  await page.goto(componentUrl("button/with_right_icon"));
+
+  await expectNoAxeViolations(page);
+
+  await expectButtonStates(page, "with-right-icon");
+});
+
+async function expectButtonStates(page, type) {
```
cnorthwood: is this how it was before? feels weird to end up with 4 different tests in one test (but not sure if it just mega slows it down so maybe it's pragmatic?)

demo/playwright/callout.spec.js
```diff
@@ -0,0 +1,66 @@
...
+
+  await expect(page).toHaveScreenshot(`callout-nested.png`, { fullPage: true });
+});
+
+test("Callout (example)", async ({ page }) => {
+  await page.goto(componentUrl("callout/example"));
+
+  await expectNoAxeViolations(page);
+
+  for (const viewport of defaultViewports) {
```
cnorthwood, line 50: similarly here, does Playwright support things like parameterised tests so any test failures are specific?

cnorthwood (commented):
so I haven't read this all yet, but noticed a pattern in the tests that I don't know if it's worth talking about first?

demo/playwright/callout.spec.js
davidrapson: It doesn't support `test.each` the guidance is just a loop https://playwright.dev/docs/test-parameterize#parameterized-tests, having this _in_ the test vs around the test seemed the same difference. But worth a quick revisit in case I missed something nicer.

demo/playwright/button.spec.js
davidrapson: It's the same _logic_ as the backstop scenarios but yeah different on loading. This feels like the right mode to me as we're testing interacting with a button as a user would. Are you thinking more from a test isolation perspective? Speed seems great overall compared to the previous setup so no major issue with batching either way if we're rather separate these.

cnorthwood: yeah it's more like test isolation - so if a test starts failing it's easier to see exactly what it is from the test name (or a failed assert early on doesn't hide failed asserts later on that you still need to debug but later)

davidrapson: 👍 Tried a refactor here to give a bit more isolation

demo/playwright/callout.spec.js
davidrapson: Refactored this scenario to run these as separate test block. Let me know your thoughts, if we're happy I can match the other scenarios to use that. Did try moving some more of the boilerplate into `test-helpers` but the UI started to get confused when `test` definitions lived in a separate file.

cnorthwood: I  wonder if wrapping the inner tests in a describe loop is slightly more idiomatic? but I am being very nit picky! I'm definitely a believer in test isolation/one assert (at least conceptually if not actually) per test but that might be a lot of work

davidrapson: I'm afraid you have lost me as to what would be different in terms of assertions there

cnorthwood: sorry, I realise I'm posting on this specific thread but my last point was more general in terms of reinforcing this approach vs some of the other tests elsewhere

davidrapson: Ah, we agree then 😅

davidrapson:
This is updated now with `main`, and have adjusted all the scenarios to be isolated where possible. I've deliberately taken this PR to a logical inclusion by including the vitest to node-test shift with the goal of having the leanest set of tools for this testing but it's also _chunky_ as a result so happy to shift some things out as needed.

demo/playwright/asset-hyperlink.spec.js
```diff
@@ -0,0 +1,18 @@
...
+    await page.goto(componentUrl("asset_hyperlink/example"));
+  });
+
+  test("accessibility check", async ({ page }) => {
+    await expectNoAxeViolations(page);
+  });
+
+  test("visual regression check", async ({ page }) => {
+    await expect(page.getByText("Test PDF")).toHaveScreenshot(
+      "asset-hyperlink.png",
```
cnorthwood, line 15: how come we specify the screenshot filename here (as opposed to using the default generated one?)

demo/playwright/button.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe("Button (with right icon)", () => {
+  buildButtonTestsFor("with-right-icon", "button/with_right_icon");
+});
+
+function buildButtonTestsFor(type, urlPath) {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(componentUrl(urlPath));
+    await page.setViewportSize({ width: 400, height: 200 });
+    await expectNoAxeViolations(page);
```
cnorthwood, line 32: having an expect in a beforeEach feels weird!

demo/playwright/button.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  test("active state", async ({ page }) => {
+    const button = page.getByRole("button");
+
+    // Active state
+    await button.hover();
+    await page.mouse.down();
+    await expect(page).toHaveScreenshot(`button-${type}-active.png`);
+    await page.mouse.up();
```
cnorthwood, line 60: do we need to mouse up after the assert or is this now superfluous?

demo/playwright/disclosure.spec.js
```diff
@@ -0,0 +1,58 @@
+import { describe, test, expect } from "@playwright/test";
+import {
+  componentUrl,
+  viewports,
+  expectNoAxeViolations,
+  expectScrolledIntoView,
+} from "./playwright-helpers";
+
+describe("Disclosure", () => {
```
cnorthwood, line 9: there's a few tests in here which test quite a few behaviours in one test rather than isolating the cases

demo/playwright/header.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  for (const viewport of defaultViewports) {
+    test(`visual regression check ${viewport.label}`, async ({ page }) => {
+      await page.setViewportSize(viewport);
+      await expect(page).toHaveScreenshot(`header-${viewport.label}.png`);
+    });
+  }
+
+  test("allows toggling search on small screens", async ({ page }) => {
```
cnorthwood, line 27: also here where it tests a whole journey with multiple assertions rather than the individual states

demo/playwright/navigation.spec.js
```diff
@@ -0,0 +1,151 @@
...
+    await moreButton.click();
+    await expectItemsInMainNavigation(page, ["Home"]);
+    await expectItemsInGreedyNavigation(page, [
+      "AdviserNet",
+      "Cymraeg",
+      "Sign in",
+    ]);
+  });
+
+  test("when interacting with a mouse", async ({ page }) => {
```
cnorthwood, line 87: another case which might be better isolated?

demo/playwright/on-this-page.spec.js
```diff
@@ -0,0 +1,109 @@
...
+        .click();
+
+      await expect(page).toHaveScreenshot(
+        `on-this-page-with-nested-links-${viewport.label}-open.png`,
+        { fullPage: true },
+      );
+    });
+  }
+
+  test("interactivity check", async ({ page }) => {
```
cnorthwood, line 42: Another one - at least the accessibility check could be pulled out but I think maybe the individual states should be too rather than one test for the whole journey

demo/playwright/pagination.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    }) => {
+      await page.setViewportSize(viewport);
+      await page.getByText("Last").click();
+      await expect(page).toHaveScreenshot(
+        `pagination-page-last-${viewport.label}.png`,
+      );
+    });
+  }
+
+  test("interaction check", async ({ page }) => {
```
cnorthwood, line 51: another test for a whole journey

demo/playwright/sample-form-builder.spec.js
```diff
@@ -0,0 +1,105 @@
+import { test, expect } from "@playwright/test";
+
+test("Sample multi page form", async ({ page }) => {
```
cnorthwood, line 3: I wonder if this is also worth breaking down?

cnorthwood (commented):
So I've made quite a few comments but in hindsight I wonder if a lot of these are actually more "component end to end" style tests that in places like public-website we might have described in Cucumber. Perhaps what's tripping me up is that it's mixed in with the unit tests, I wonder if they need being made explicitly named (rather than just 'interaction check') to make it more obvious that these aren't unit tests but more end-to-end ones and it's okay for them to do multiple assertions on a long journey?

demo/playwright/asset-hyperlink.spec.js
davidrapson: Largely clarity in this case the generated would be `Asset-hyperlink-visual-regression-check-1.png`, or from the callout example `Callout-adviser-visual-regression-check-large-1.png` vs `callout-adviser-large.png`.

It also makes the format explicit in playwright, the _default_ is `png` but if you were doing any other format you need to specify a filename anyway, so prefer always making it explicit.

demo/playwright/button.spec.js
davidrapson: Oh yeah!

davidrapson:
> Perhaps what's tripping me up is that it's mixed in with the unit tests, I wonder if they need being made explicitly named (rather than just 'interaction check') to make it more obvious that these aren't unit tests but more end-to-end ones and it's okay for them to do multiple assertions on a long journey?

Yeah I think this makes sense, there's definitely a "some of these are not like the others" thing still lingering. Let's pick up over chat 🙂

demo/playwright/components/navigation.spec.js
```diff
@@ -0,0 +1,151 @@
...
+    await moreButton.click();
+    await expectItemsInMainNavigation(page, ["Home"]);
+    await expectItemsInGreedyNavigation(page, [
+      "AdviserNet",
+      "Cymraeg",
+      "Sign in",
+    ]);
+  });
+
+  test("when interacting with a mouse", async ({ page }) => {
```
cnorthwood: I wonder if this and the keyboard test below also belong in the e2e tests?

davidrapson: Glad this and the header still jump out, they were the two left that felt a bit hybrid

davidrapson:
Pretty happy with where this is at now. The `components` directory feels as lean as it can be now and content _enough_ with the shape of the `e2e` directory; that feels like the one we'd most want to expand on anyway.

cnorthwood (approved):

PR #4232, opened 2026-09-01
Experimental: Use node:test for non-Playwright tests

davidrapson:
Based on the changes in https://github.com/citizensadvice/design-system/pull/4231 which consolidates all browser testing to use Playwright I realised the only remaining javascript tests we have are the sass output tests and so a full vitest setup might be overkill.

It turns out `node:test` is great now, and even has snapshot support. Tests to see if we can switch all the sass tests to use plain `node:test`

Consider this experimental until we've decided if we're happy with the Playwright set up, but/and combined with #4205 would bring the number of npm dependencies down to the bare-bones.


davidrapson:
Combined with https://github.com/citizensadvice/design-system/pull/4231

